PR #2, opened 2026-05-12
security: replace hardcoded test RSA key and add gitleaks config

lml2468:
## Summary

Replace the hardcoded RSA private key in `pkg/wkrsa/rsa_test.go` with a freshly generated test-only key and add a `.gitleaks.toml` configuration.

## Changes

- **`pkg/wkrsa/rsa_test.go`**: Replace the old RSA key with a new 2048-bit test key. Extract into a shared `testPrivateKey` const.
- **`.gitleaks.toml`**: Allowlist test file paths to suppress future false positives.

## Verification

- All `pkg/wkrsa` tests pass with the new key (`go test ./pkg/wkrsa/ -v`).

## Follow-up needed (post-merge)

Git history still contains the old RSA private key from the initial commit. A `git filter-repo` pass on the default branch is recommended before making this repository public.

PR #3, opened 2026-05-13
ci: add Octo issue/PR feed workflows

lml2468:
Add real-time GitHub event notification to Octo IM channels via GitHub Actions.

- `octo-issue-feed.yml`: pushes Issue events (opened/closed/reopened/labeled) to Octo 🛠️ issue-feed and the repo's project channel
- `octo-pr-feed.yml`: pushes PR events (opened/closed/merged/etc.) to Octo 🛠️ pr-feed and the repo's project channel

### DIFF
--- a/.github/workflows/octo-issue-feed.yml
+++ b/.github/workflows/octo-issue-feed.yml
@@ -1,0 +1,20 @@
+name: Octo Issue Feed
+
+on:
+  issues:
+    types: [opened, closed, reopened, labeled]
+
+jobs:
+  notify:
+    uses: Mininglamp-OSS/.github/.github/workflows/octo-issue-feed.yml@main
+    with:
+      repo_name: ${{ github.event.repository.name }}
+      issue_number: ${{ github.event.issue.number }}
+      issue_title: ${{ github.event.issue.title }}
+      issue_url: ${{ github.event.issue.html_url }}
+      issue_author: ${{ github.event.issue.user.login }}
+      issue_labels: ${{ toJson(github.event.issue.labels.*.name) }}
+      event_action: ${{ github.event.action }}
+      project_group_id: 6095b6ded00046009e72882ce4067289
+    secrets:
+      OCTO_BOT_TOKEN: ${{ secrets.OCTO_BOT_TOKEN }}
--- a/.github/workflows/octo-pr-feed.yml
+++ b/.github/workflows/octo-pr-feed.yml
@@ -1,0 +1,23 @@
+name: Octo PR Feed
+
+on:
+  pull_request:
+    types: [opened, closed, reopened, ready_for_review, review_requested]
+
+jobs:
+  notify:
+    uses: Mininglamp-OSS/.github/.github/workflows/octo-pr-feed.yml@main
+    with:
+      repo_name: ${{ github.event.repository.name }}
+      pr_number: ${{ github.event.pull_request.number }}
+      pr_title: ${{ github.event.pull_request.title }}
+      pr_url: ${{ github.event.pull_request.html_url }}
+      pr_author: ${{ github.event.pull_request.user.login }}
+      pr_merged: ${{ github.event.pull_request.merged }}
+      pr_additions: ${{ github.event.pull_request.additions }}
+      pr_deletions: ${{ github.event.pull_request.deletions }}
+      pr_changed_files: ${{ github.event.pull_request.changed_files }}
+      event_action: ${{ github.event.action }}
+      project_group_id: 6095b6ded00046009e72882ce4067289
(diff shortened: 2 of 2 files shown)


yujiawei (commented):
# Code Review — PR #3 (octo-lib)

## Summary

Two thin caller workflows that delegate to the reusable workflows in `Mininglamp-OSS/.github`. The change is small (2 files, +43/-0), purely additive (no existing CI affected), and the input wiring matches the reusable workflow signatures verified against `Mininglamp-OSS/.github/.github/workflows/octo-{issue,pr}-feed.yml@main`.

The reusable workflow uses the safe `env:` indirection pattern (inputs are exposed as env vars and read via `os.environ` in Python), so passing user-controlled fields like `issue_title` / `pr_title` through these caller files does not create a script-injection sink at this layer.

Verdict: low-risk, ship as-is. A few minor hardening notes below — none blocking.

## Findings

### 1. No `permissions:` block (minor, hardening)

Neither workflow declares `permissions:`. The job inherits the repo/org default `GITHUB_TOKEN` scope, which on many setups is `contents: write` or broader. These workflows do not need the token at all (auth to the IM API uses `OCTO_BOT_TOKEN`), so an explicit minimal block is recommended:

```yaml
permissions:
  contents: read
```

Add it at the top level (not on the reusable-workflow `uses:` job, since `permissions:` on a calling job has no effect — it must be at the workflow or top-level scope).

### 2. Reusable workflow pinned to `@main` (minor, supply-chain)

`uses: Mininglamp-OSS/.github/.github/workflows/octo-{issue,pr}-feed.yml@main`

For internal org-owned reusable workflows this is the common pattern, but `@main` means every push to that file ships immediately to all callers, with no review gate at the consumer side. If the central `.github` repo's `main` is ever force-pushed or compromised, every repo using this caller runs the new code on the next event. For non-critical IM notifications this is acceptable, but worth knowing — pinning to a tag (e.g. `@v1`) would let you cut releases of the reusable workflow without auto-rolling consumers.

### 3. `project_group_id` hardcoded per repo (minor, maintainability)

`project_group_id: 6095b6ded00046009e72882ce4067289` is the same string in both files and is repo-specific (octo-lib's project channel). Two minor consequences:

- If the channel ID ever changes, both files need to be updated in lockstep (easy to miss one).
- This caller pair cannot be copy-pasted to another repo without remembering to change the ID. If you anticipate rolling this out to many repos, consider moving the per-repo ID to a repo or org variable (`vars.OCTO_PROJECT_GROUP_ID`) so the YAML body becomes identical across repos.

Not a defect for octo-lib alone.

### 4. `pull_request` event types — `merged` is correctly absent, doc wording is loose

The PR description lists "opened/closed/merged/etc." but the workflow uses `[opened, closed, reopened, ready_for_review, review_requested]`. This is correct — `merged` is not a `pull_request` activity type; merges arrive as `closed` with `pull_request.merged == true`, which is exactly how the reusable workflow's `pr_merged` input is consumed. No code change needed; the description is just imprecise.

### 5. `issues` event types — `labeled` without `unlabeled` (intentional?)

`octo-issue-feed.yml` listens for `labeled` but not `unlabeled`. If the goal is "notify when an issue is triaged," that's fine. If the goal is "keep the channel in sync with label state," `unlabeled` would also be needed. Worth a quick confirmation; either answer is defensible.

### 6. No `concurrency:` group (informational)

Rapid label additions / status flips (e.g. an automation labeling an issue several times in a second) will spawn multiple parallel notification jobs and produce duplicate IM messages. For a notification feed this is usually fine — events are individually meaningful — but if you ever see noisy bursts, a per-issue concurrency group like:

```yaml
concurrency:
  group: octo-issue-feed-${{ github.event.issue.number }}
  cancel-in-progress: false
```

would coalesce the queue without dropping events.

## Non-issues verified

- ✅ `issue_labels: ${{ toJson(github.event.issue.labels.*.name) }}` — JSON-encoded, parsed safely by the reusable workflow's `json.loads(os.environ['ISSUE_LABELS'])` with a `try/except` fallback.
- ✅ `secrets: OCTO_BOT_TOKEN` is forwarded explicitly (not `secrets: inherit`), which is the right level of disclosure for a workflow that only needs that one secret.
- ✅ Both files are under `.github/workflows/` with valid YAML and the input names match the reusable workflow's `workflow_call.inputs` schema (verified against `Mininglamp-OSS/.github@main`).
- ✅ No third-party actions are pinned by tag — there are no `uses: external/action@v...` lines, so no SHA-pinning concern beyond finding #2.

## Recommendation

Ship as-is. The findings above are hardening / hygiene only and can be addressed in a follow-up if/when the pattern is rolled out to more repos.

PR #6, opened 2026-05-14
ci: improve GitHub Actions workflows for octo-lib

lml2468:
### A5: Add CI workflow for Go 1.20 library

Added `.github/workflows/ci.yml` with build, test, and vet jobs (no lint for base library):
- Triggers on push/PR to `main`
- Go version: 1.20
- `build`: `CGO_ENABLED=0 go build -v ./...`
- `test` (after build): `go test -race -count=1 -timeout 5m ./...`
- `vet`: `go vet ./...`

### B: Migrate `octo-pr-feed.yml` to `pull_request_target` + pinned SHA + `permissions: {}`

- Changed trigger from `pull_request` to `pull_request_target` so org secrets are available for fork PRs
- Pinned reusable workflow to SHA `c6c51348fccac99161166dce1b52dec1063aeea2` for supply-chain security
- Added `permissions: {}` to restrict default token permissions

### D: Add `permissions: {}` to `pr-merged-done.yml`

Adds explicit `permissions: {}` block between `on:` and `jobs:` to clarify that `GITHUB_TOKEN` is not used; only `PROJECT_TOKEN` is passed explicitly.

### E: Add `issue-welcome.yml`

Added missing `.github/workflows/issue-welcome.yml` that posts a welcome comment when new issues are opened, using the org-level reusable workflow.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,52 @@
+name: CI
+
+on:
+  pull_request:
+    branches: [main]
+  push:
+    branches: [main]
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+env:
+  GO_VERSION: '1.20'
+
+jobs:
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v5
+        with:
+          go-version: ${{ env.GO_VERSION }}
+      - name: Build
+        run: CGO_ENABLED=0 go build -v ./...
+
+  test:
+    name: Test
+    runs-on: ubuntu-latest
+    needs: build
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v5
+        with:
+          go-version: ${{ env.GO_VERSION }}
+      - name: Test
+        run: go test -race -count=1 -timeout 5m ./...
+
+  vet:
+    name: Vet
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v5
+        with:
+          go-version: ${{ env.GO_VERSION }}
+      - name: Vet
+        run: go vet ./...
--- a/.github/workflows/issue-welcome.yml
+++ b/.github/workflows/issue-welcome.yml
@@ -8,6 +8,4 @@ jobs:
   welcome:
     permissions:
       issues: write
-    uses: Mininglamp-OSS/.github/.github/workflows/issue-welcome.yml@main
-    secrets:
-      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    uses: Mininglamp-OSS/.github/.github/workflows/issue-welcome.yml@3d90d99236e0427a4fbe5b8f25b209ac0b799b3e
--- a/.github/workflows/issue-welcome.yml
+++ b/.github/workflows/issue-welcome.yml
@@ -1,0 +1,13 @@
+name: Issue Welcome
+
+on:
+  issues:
+    types: [opened]
+
+jobs:
+  welcome:
+    permissions:
+      issues: write
+    uses: Mininglamp-OSS/.github/.github/workflows/issue-welcome.yml@main
+    secrets:
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/octo-pr-feed.yml
+++ b/.github/workflows/octo-pr-feed.yml
@@ -1,12 +1,14 @@
 name: Octo PR Feed
 
 on:
-  pull_request:
+  pull_request_target:
     types: [opened, closed, reopened, ready_for_review, review_requested]
+
+permissions: {}
 
 jobs:
   notify:
-    uses: Mininglamp-OSS/.github/.github/workflows/octo-pr-feed.yml@main
+    uses: Mininglamp-OSS/.github/.github/workflows/octo-pr-feed.yml@c6c51348fccac99161166dce1b52dec1063aeea2
     with:
       repo_name: ${{ github.event.repository.name }}
       pr_number: ${{ github.event.pull_request.number }}
--- a/.github/workflows/pr-merged-done.yml
+++ b/.github/workflows/pr-merged-done.yml
@@ -4,6 +4,8 @@ on:
   pull_request_target:
     types: [closed]
 
+permissions: {}  # GITHUB_TOKEN not used; PROJECT_TOKEN is passed explicitly
+
 jobs:
   move-to-done:
     if: github.event.pull_request.merged == true && github.event.pull_request.base.ref == github.event.repository.default_branch
--- a/.github/workflows/pr-merged-done.yml
+++ b/.github/workflows/pr-merged-done.yml
@@ -9,6 +9,6 @@ permissions: {}  # GITHUB_TOKEN not used; PROJECT_TOKEN is passed explicitly
 jobs:
   move-to-done:
     if: github.event.pull_request.merged == true && github.event.pull_request.base.ref == github.event.repository.default_branch
-    uses: Mininglamp-OSS/.github/.github/workflows/pr-merged-done.yml@main
+    uses: Mininglamp-OSS/.github/.github/workflows/pr-merged-done.yml@3d90d99236e0427a4fbe5b8f25b209ac0b799b3e
     secrets:
       PROJECT_TOKEN: ${{ secrets.PROJECT_TOKEN }}


yujiawei (commented):
# Code Review — PR #6 (octo-lib)

Thanks for tightening up the workflows. The general direction (CI for Go 1.20, locked-down permissions, SHA-pinned reusable call) is good. A few items below — one is likely to break the new `Issue Welcome` workflow at runtime, the others are consistency / efficiency.

## Verification summary

| Item | Status | Notes |
|---|---|---|
| `ci.yml` Go version matches `go.mod` (1.20) | ✅ | `go.mod:3` declares `go 1.20`; CI sets `GO_VERSION: '1.20'` (`.github/workflows/ci.yml:16`). |
| `octo-pr-feed.yml` `pull_request_target` switch is safe in this caller | ✅ | The pinned reusable workflow (`@c6c5134`) does not check out PR code and exposes user-controlled fields (`pr_title`, `pr_author`) via `env:` to a Python script (read with `os.environ`), not as inline shell. Combined with `permissions: {}`, the classic `pull_request_target` exfiltration vectors do not apply here. |
| `pr-merged-done.yml` `permissions: {}` addition | ✅ | Purely additive; explicit-secret pass for `PROJECT_TOKEN` was already in place. |
| `issue-welcome.yml` secret pass | ❌ | See P0 below. |
| Supply-chain pinning consistent across workflows | ❌ | See P1 #1 below. |

## Findings

### P0 — `issue-welcome.yml` will fail validation: cannot pass `GITHUB_TOKEN` to a reusable workflow that doesn't declare it

`.github/workflows/issue-welcome.yml:11-13`
```yaml
uses: Mininglamp-OSS/.github/.github/workflows/issue-welcome.yml@main
secrets:
  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

The called reusable workflow (`Mininglamp-OSS/.github/.github/workflows/issue-welcome.yml@main`) declares only:

```yaml
on:
  workflow_call:
```

…with no `secrets:` block. GitHub Actions rejects callers that pass any named secret the reusable workflow has not declared (regardless of whether it's `GITHUB_TOKEN` or a custom secret). Expected error at run time:

> `secret 'GITHUB_TOKEN' is not defined in the called workflow`

`GITHUB_TOKEN` is also auto-injected into every workflow run context, so the reusable workflow already has it as `${{ secrets.GITHUB_TOKEN }}` without any explicit pass.

**Fix (pick one):**
- Drop the `secrets:` block entirely. The reusable workflow already reads `secrets.GITHUB_TOKEN` from its own context.
- Or use `secrets: inherit` (also drops the explicit token line).

Recommended diff:
```diff
   welcome:
     permissions:
       issues: write
-    uses: Mininglamp-OSS/.github/.github/workflows/issue-welcome.yml@main
-    secrets:
-      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    uses: Mininglamp-OSS/.github/.github/workflows/issue-welcome.yml@main
```

This is worth verifying with a one-shot test issue on a fork before relying on it.

### P1

**1. Inconsistent SHA pinning across reusable workflow calls**

The PR pins `octo-pr-feed.yml` to commit `c6c51348fcc…` for "supply-chain security" (`.github/workflows/octo-pr-feed.yml:11`), but the new `issue-welcome.yml` uses `@main` (`.github/workflows/issue-welcome.yml:11`). The supply-chain risk is symmetric — both invoke a reusable workflow from another repo and expose write permissions (`issues: write` for the welcome bot, secrets for the feed). If pinning is the policy, pin both to a SHA. If `@main` is acceptable for first-party `Mininglamp-OSS/.github`, document that in the PR description so the inconsistency is intentional.

While at it, the same applies to actions used in `ci.yml` (`actions/checkout@v4`, `actions/setup-go@v5`). For first-party GitHub Actions this is usually fine, but if the new pinning policy is meant to be uniform, these should be pinned too.

**2. CI cost: `test` job re-runs the whole build, no module cache reuse across jobs**

`.github/workflows/ci.yml:30-49`

Three jobs (`build`, `test`, `vet`) each re-run `actions/checkout` + `actions/setup-go` and re-download all module dependencies (`go.sum` is large — see `go.sum`, ~86 KB). `setup-go@v5` does cache modules per job by `go.sum` hash, so the second and third jobs in a run usually hit cache, but each still pays the checkout + cache-restore overhead.

More importantly: `test` declares `needs: build`, but does not consume any artifact from `build` — it does its own `go test`, which compiles everything again from scratch. So `needs: build` only serializes the jobs without sharing work. For a base library this serialization just slows CI by ~30–60s with no signal benefit, since `go test ./...` already implies compile. Suggested simplification:

```diff
 jobs:
-  build:
-    name: Build
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v4
-      - uses: actions/setup-go@v5
-        with:
-          go-version: ${{ env.GO_VERSION }}
-      - name: Build
-        run: CGO_ENABLED=0 go build -v ./...
-
-  test:
-    name: Test
-    runs-on: ubuntu-latest
-    needs: build
-    steps:
-      - uses: actions/checkout@v4
-      - uses: actions/setup-go@v5
-        with:
-          go-version: ${{ env.GO_VERSION }}
-      - name: Test
-        run: go test -race -count=1 -timeout 5m ./...
+  build-test:
+    name: Build & Test
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v5
+        with:
+          go-version: ${{ env.GO_VERSION }}
+      - name: Build
+        run: CGO_ENABLED=0 go build -v ./...
+      - name: Test
+        run: go test -race -count=1 -timeout 5m ./...

   vet:
     name: Vet
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-go@v5
         with:
           go-version: ${{ env.GO_VERSION }}
       - name: Vet
         run: go vet ./...
```

`vet` can stay as a parallel job (it gives a different signal and can fail fast).

### P2

**3. No `gofmt` / formatting check.** A common minimum for a base library, since reviewers can't catch subtle formatting drift in a year of contributions. One step suffices:
```yaml
      - name: gofmt
        run: |
          out=$(gofmt -l .)
          if [ -n "$out" ]; then echo "$out"; exit 1; fi
```

**4. No `paths-ignore` for doc-only changes.** Optional, but README/CHANGELOG/`.md` updates currently trigger the full Go matrix. Low-impact, easy win.

**5. `octo-pr-feed.yml` `types:` list under `pull_request_target`.** The type list (`opened, closed, reopened, ready_for_review, review_requested`) is unchanged from the prior `pull_request` trigger. Worth confirming the `closed` event under `pull_request_target` is the desired behavior given that `pr-merged-done.yml` already handles the closed/merged transition; otherwise you'll get two notifications on every merged PR (one from the feed, one from the board move). Not a defect — just flagging because it's a behavior change from running in fork context too.

## Extra observations (not blocking)

- `concurrency` group in `ci.yml:13-15` correctly uses `pull_request.number || github.ref` so PR runs and main pushes don't cancel each other. Good.
- `permissions: contents: read` at workflow level in `ci.yml:8-9` is appropriately minimal.
- The PR description is unusually clean and labeled by sub-change (A5/B/D/E). That made review easier — keep that style.

## Verdict

`COMMENTED` — please address P0 (will block the welcome workflow from running) and P1 #1 (consistency with the stated supply-chain rationale). P1 #2 and P2 items are recommendations.

lml2468:
Fixed based on review feedback:
- **P0/P1**: Removed the `secrets: GITHUB_TOKEN` block — the reusable workflow does not declare secrets, causing a runtime validation error
- **P1**: Pinned reusable workflow to commit SHA `3d90d99236e0427a4fbe5b8f25b209ac0b799b3e` instead of `@main` for supply-chain security

yujiawei (dismissed):
# Code Review — PR #6 (octo-lib)

## Summary

Solid CI/security hygiene pass for `octo-lib`. Adds a build/test/vet workflow targeting Go 1.20 (matches `go.mod`), adopts `pull_request_target` + SHA-pinned reusable workflows for org-level secret access, tightens default token permissions, and restores the missing `issue-welcome.yml`. No blocking issues found.

**Verdict: Approved with a few non-blocking suggestions.**

---

## Verification

| Item | Result | Notes |
|---|---|---|
| `ci.yml` Go version matches `go.mod` | ✅ | Both at `1.20` |
| `CGO_ENABLED=0 go build ./...` | ✅ | Verified locally on the PR head — exit 0; `_ "github.com/mattn/go-sqlite3"` is a blank-import driver register and compiles fine without cgo |
| `go test -race -count=1 -timeout 5m ./...` | ✅ | Race detector + cache-bust + bounded timeout — good defaults |
| Pinned SHA for `octo-pr-feed.yml` reusable | ✅ | `c6c51348fccac99161166dce1b52dec1063aeea2` exists at `Mininglamp-OSS/.github` and matches the consumed input/secret schema |
| Pinned SHA for `issue-welcome.yml` reusable | ✅ | `3d90d99236e0427a4fbe5b8f25b209ac0b799b3e` exists at `Mininglamp-OSS/.github` (`workflow_call`, no inputs/secrets required) |
| `pull_request_target` security review | ✅ | PR-controlled fields (`pr_title`, `pr_author`, …) flow through typed reusable inputs into `env:` and are read via `os.environ` in the Python step — not interpolated into a shell `run:` block, so no script-injection vector |
| `permissions: {}` on `octo-pr-feed.yml` / `pr-merged-done.yml` | ✅ | Caller permissions cap reusable permissions; neither reusable needs `GITHUB_TOKEN` scopes (one calls Octo IM via HTTP, the other uses `PROJECT_TOKEN`) |
| `permissions: issues: write` on `issue-welcome.yml` (job level) | ✅ | Required so the called reusable, which posts via `actions/github-script`, can comment on issues |

---

## Findings

### P2 — `pr-merged-done.yml` reusable still references `@main`

`octo-pr-feed.yml` and `issue-welcome.yml` were pinned to a specific commit in this PR, but `pr-merged-done.yml` still uses `@main`:

```yaml
# .github/workflows/pr-merged-done.yml:13
uses: Mininglamp-OSS/.github/.github/workflows/pr-merged-done.yml@main
```

Given `pr-merged-done.yml` runs under `pull_request_target` with `PROJECT_TOKEN`, the supply-chain argument that motivated pinning the other two applies equally here. Suggest pinning to a SHA in the same PR (or as an immediate follow-up) so the policy is consistent across all three reusable callers.

### P2 — `test` job's `needs: build` adds queue latency without sharing artifacts

```yaml
# .github/workflows/ci.yml:32
test:
  ...
  needs: build
```

`build` produces no uploaded artifact, and `go test ./...` independently compiles every package, so making `test` wait on `build` only adds runner-acquisition latency. Either:

- Run `build`/`test`/`vet` in parallel (drop `needs: build`), or
- Drop the `build` job entirely — `go test ./...` already proves the module compiles. Keeping a separate `build` job is only useful if you want to verify `CGO_ENABLED=0` builds, which is a defensible intent; in that case still drop the `needs:` edge.

### P2 — Single Go version

`GO_VERSION: '1.20'` is the only matrix entry. This is an OSS library; consumers may be on newer Go releases. A small `strategy.matrix.go-version: ['1.20', '1.22']` (or the current latest) on the `test` job would catch consumer-side regressions without much cost. Optional — fine to defer.

### Nits (non-blocking)

- `actions/checkout@v4` and `actions/setup-go@v5` are not SHA-pinned. First-party Actions on `vN` major tags are widely accepted, and you've correctly SHA-pinned the third-party / org reusable workflows where the supply-chain risk is higher. Calling out as a stylistic note only.
- `concurrency.group` uses `github.event.pull_request.number || github.ref` — works, but `github.head_ref || github.ref` is the more common idiom and avoids cancelling unrelated `push` runs that happen to share a number space. Cosmetic.
- The repeated `actions/checkout@v4` + `actions/setup-go@v5` block across three jobs could be a composite action or matrix, but for a 52-line workflow it's fine to leave as-is.

---

## What's done well

- `permissions: contents: read` at workflow level + job-scoped writes only where necessary — least-privilege baseline.
- `concurrency` with `cancel-in-progress: true` keyed on PR number — correct cancellation behavior for branch pushes.
- `go test -race -count=1 -timeout 5m` — race detector on, cache disabled, bounded.
- `pull_request_target` switch is the right call for fork PRs needing org-level `OCTO_BOT_TOKEN`, and the data flow through typed reusable inputs into `env:` is safe.
- Reusable workflows pinned by full SHA (where pinned) — supply-chain hygiene.
- Restoring the missing `issue-welcome.yml` brings this repo back into parity with the org template.

---

## Recommendation

**Approve.** Address the P2 SHA-pinning inconsistency on `pr-merged-done.yml` either in this PR or as a fast-follow; the rest are optional.

Jerry-Xin (changes requested):
Relevant CI/automation PR, but one supply-chain issue remains in a touched workflow that passes a privileged secret.

**🔴 Blocking**
- 🟡 Warning: [.github/workflows/pr-merged-done.yml:12](/private/tmp/pr-review-B1T2uC/repo/.github/workflows/pr-merged-done.yml:12) still calls `Mininglamp-OSS/.github/.github/workflows/pr-merged-done.yml@main` while passing `PROJECT_TOKEN` at line 14. `permissions: {}` only restricts `GITHUB_TOKEN`; it does not protect `PROJECT_TOKEN` from changes to the reusable workflow on mutable `main`. Pin this reusable workflow to a reviewed commit SHA, matching the approach used in `octo-pr-feed.yml`.

**💬 Non-blocking**
- 🔵 Suggestion: [.github/workflows/ci.yml:24](/private/tmp/pr-review-B1T2uC/repo/.github/workflows/ci.yml:24) could pin third-party actions by SHA for stronger supply-chain hardening, but this is broader than the current org reusable workflow risk.
- 🔵 Suggestion: [.github/workflows/ci.yml:17](/private/tmp/pr-review-B1T2uC/repo/.github/workflows/ci.yml:17) uses Go `1.20`, which matches `go.mod`; consider a future compatibility job for a newer supported Go version.

**✅ Highlights**
- `octo-pr-feed.yml` now uses `pull_request_target` without checking out untrusted PR code, passes PR fields through environment variables in the reusable workflow, and pins the reusable workflow SHA.
- Local verification passed: `CGO_ENABLED=0 go build -v ./...`, `go test ./...`, `go test -race -count=1 -timeout 5m ./...`, and `go vet ./...`.

lml2468 (commented):
Verdict: COMMENTED; I found no additional issues beyond the existing unresolved change request, so I am not duplicating it as a new review finding.

**Blocking items**
1. None newly raised in this review. The existing blocker on `.github/workflows/pr-merged-done.yml:12` has already been raised and should remain the merge blocker until resolved.

**Non-blocking notes**
- None newly raised. Existing suggestions around SHA-pinning official actions and future Go-version coverage are already covered in prior reviews.

**Highlights**
- CI gate is green: Build, Test, Vet, and code-review all passed.
- I read the changed workflow files from `fix/workflow-improvements-octo-lib` at head SHA `94bbdfbcf077e280a22a297920074ba8662466fc`.
- The pinned reusable workflow targets for issue welcome and PR feed exist at their referenced SHAs.
- `octo-pr-feed.yml` uses `pull_request_target` without checking out untrusted PR code, has `permissions: {}`, and passes PR data as reusable-workflow inputs/environment values rather than interpolating it into shell.

Note: the authenticated GitHub account is the PR author, so any would-be approval must be submitted as a comment review rather than `--approve`.

lml2468:
Fixed: pinned pr-merged-done.yml reusable workflow reference from @main to a specific commit SHA, matching the supply-chain hardening already applied to octo-pr-feed.yml in this PR.

yujiawei (commented):
# Code Review — PR #6 (octo-lib)

## Summary

This PR adds a Go 1.20 CI workflow (`build` / `test` / `vet`), migrates `octo-pr-feed.yml` to `pull_request_target` with a SHA-pinned reusable workflow, adds explicit `permissions: {}` to the PR/issue feed and `pr-merged-done` callers, pins `pr-merged-done.yml` to a commit SHA, and adds a new `issue-welcome.yml` caller. Scope is tight, well-described, and the security posture moves in the right direction.

I verified the changes end-to-end against the org-level reusable workflows; nothing here is a merge blocker. All findings below are improvements or follow-ups.

## 1. Verification Results

| Item | Status | Evidence |
|---|---|---|
| Pinned SHA `c6c51348fccac99161166dce1b52dec1063aeea2` exists in `Mininglamp-OSS/.github` | ✅ | API resolves, contains `octo-pr-feed.yml` reusable with `workflow_call` trigger |
| Pinned SHA `3d90d99236e0427a4fbe5b8f25b209ac0b799b3e` exists in `Mininglamp-OSS/.github` | ✅ | API resolves; both `pr-merged-done.yml` and `issue-welcome.yml` callable at this SHA |
| `pull_request_target` migration is safe (no untrusted-code execution) | ✅ | Caller does NOT `actions/checkout` PR head; PR-controlled values (`pr_title`, `pr_author`, `pr_url`) flow via `inputs:` → `env:` → `os.environ[...]`, never via bash interpolation, so they cannot escape into shell context |
| `permissions: {}` correctly denies default `GITHUB_TOKEN` scopes for `octo-pr-feed.yml` and `pr-merged-done.yml` | ✅ | Neither reusable workflow uses `GITHUB_TOKEN` — `octo-pr-feed` uses `OCTO_BOT_TOKEN`, `pr-merged-done` uses `PROJECT_TOKEN`, both passed explicitly |
| `issue-welcome.yml` caller's job-level `permissions: issues: write` intersects correctly with the reusable's `permissions: issues: write` | ✅ | Effective scope: `issues: write`, which is what `github-script` needs to post the comment |
| Concurrency group keys correctly per-PR and per-branch | ✅ | `${{ github.event.pull_request.number || github.ref }}` falls back to `github.ref` on push events |
| Go 1.20 in CI matches `go.mod` | ✅ | `go.mod` declares `go 1.20` |
| `CGO_ENABLED=0 go build ./...` will not fail despite `pkg/db/sqlite.go` importing `mattn/go-sqlite3` | ✅ | `mattn/go-sqlite3 v1.14.17` ships `static_mock.go` with `// +build !cgo` — it compiles to a runtime-error stub rather than failing to compile |

## 2. Findings

### P1 — `pull_request_target` migration lacks an in-file safety comment

**File:** `.github/workflows/octo-pr-feed.yml`

The migration from `pull_request` to `pull_request_target` is correct (and necessary for fork PRs to access org secrets), but `pull_request_target` is one of the most footgun-prone GitHub Actions triggers. The current PR is safe because:

1. The caller does not `actions/checkout` PR head code.
2. The reusable workflow at the pinned SHA uses `env:` (not `run: echo ${{ inputs.foo }}`) for PR-controlled values, so attacker-controlled titles/URLs cannot break out into bash.
3. The pinned SHA freezes that contract.

But future maintainers bumping the SHA — or adding a `steps:` block before the `uses:` — could reintroduce a critical fork-PR-RCE vulnerability. Recommend a short header comment to make the invariants explicit, e.g.:

```yaml
# SECURITY: this workflow uses `pull_request_target`, which runs in the base
# repo's context with secrets available. Do NOT add `actions/checkout` of
# PR head code, do NOT interpolate `github.event.pull_request.*` directly
# into shell `run:` blocks, and re-audit before bumping the reusable SHA.
```

This is preventive, not a current bug.

### P2 — `CGO_ENABLED=0` build vs CGO-enabled test is inconsistent

**File:** `.github/workflows/ci.yml` (lines 28, 39)

- `build`: `CGO_ENABLED=0 go build -v ./...` — uses the `mattn/go-sqlite3` stub
- `test`: `go test -race -count=1 -timeout 5m ./...` — `-race` requires CGO, so this implicitly runs with `CGO_ENABLED=1` and the real sqlite3 driver

So the build job and the test job validate different binaries. For a library this is mostly a portability check (does it compile in pure-Go contexts?), but the two jobs ought to share intent. Options:
- Drop `CGO_ENABLED=0` from `build` if the goal is just "does it compile".
- Add a separate `build-nocgo` job and keep `build` matching `test`.
- Document the intent in a comment.

Either way, today the two jobs answer different questions without saying so.

### P2 — `needs: build` adds latency without sharing artifacts

**File:** `.github/workflows/ci.yml` (line 33)

```yaml
test:
  needs: build
  steps:
    - uses: actions/checkout@v4
    - uses: actions/setup-go@v5
    - run: go test -race -count=1 -timeout 5m ./...
```

`test` waits for `build` to finish, but it then re-checks-out, re-installs Go, and re-compiles everything from scratch (`go test` does its own build). The `needs:` only adds queue/setup overhead — it doesn't gate on anything `test` would otherwise miss, since `go test` won't run if the package fails to compile.

Recommend either (a) drop `needs: build` so the two run in parallel, (b) merge `build` and `test` into one job, or (c) keep `needs:` but cache the build output (`actions/setup-go@v5` already caches modules; sharing the build cache would need `actions/cache` keyed on `go.sum`).

### P2 — `cancel-in-progress: true` on push to `main`

**File:** `.github/workflows/ci.yml` (lines 12-14)

```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
  cancel-in-progress: true
```

For PR pushes this is correct — cancel the previous run when the contributor force-pushes. But for `push` to `main`, the group key is `CI-refs/heads/main`, so two merges landing back-to-back will cancel each other's CI. For a library where `main` builds are also a record-of-correctness, you usually want to keep them all. Common pattern:

```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```

This is opinion, not a defect.

### P2 — `actions/checkout@v4` and `actions/setup-go@v5` not pinned to SHA

**File:** `.github/workflows/ci.yml`

The PR pins reusable workflows to commit SHAs (good supply-chain practice), but third-party actions are pinned only to major-version tags. For consistency, those would also be SHA-pinned (with Dependabot configured to bump them). Acceptable to leave as-is for a base library, but inconsistent with the stated hardening goal.

### P2 — Go 1.20 has been EOL since Feb 2024

`go.mod` declares `go 1.20`; CI follows. Out of scope for THIS PR, but worth a follow-up issue: 1.20 no longer receives security patches, and several deps in `go.sum` (gin, dbr, machinery) have published versions that require a newer toolchain.

### P2 — `octo-issue-feed.yml` not migrated

The repo also has `.github/workflows/octo-issue-feed.yml`, which is not touched in this PR. If it has the same fork-issue-secret needs, it should receive the same `pull_request_target` + pinned-SHA + `permissions: {}` treatment for consistency. Either include it here or open a follow-up.

## 3. Suggestions

- Add the security comment block to `octo-pr-feed.yml` (P1).
- Reconcile `CGO_ENABLED=0` build with the CGO-enabled test job, or document the intent (P2).
- Drop `needs: build` or share build cache between jobs (P2).
- Consider `cancel-in-progress: ${{ github.event_name == 'pull_request' }}` to preserve `main` build history (P2).
- Open a follow-up to (a) bump Go past EOL, (b) align `octo-issue-feed.yml`, (c) optionally SHA-pin first-party actions.

## 4. Additional Observations

- The PR description's claim "no lint for base library" is consistent with the diff — `golangci-lint` / `staticcheck` are intentionally absent. That's a deliberate scope decision and not flagged.
- `pr-merged-done.yml` correctly uses `PROJECT_TOKEN` (not `GITHUB_TOKEN`) for cross-repo project board mutations; the `permissions: {}` block at the caller correctly reflects that `GITHUB_TOKEN` is unused. Comment on line 7 is accurate.
- Commit history is clean — six small commits, each addressing one concern. The two `fix:` commits at the end (`94bbdfb` and `0c1678e`) reflect the SHA-pinning iteration; squash-merge will collapse them cleanly.

## Verdict

**COMMENTED.** The changes are well-scoped, security-conscious, and verified end-to-end against the org's reusable workflows. The P1 finding is preventive (a comment to lock in `pull_request_target` invariants), and the P2 findings are quality-of-life improvements that can ship in this PR or as follow-ups. No blockers.

Jerry-Xin (approved):
Summary: This PR is in scope for `octo-lib`; it improves CI and repository automation without introducing blocking workflow, security, or compatibility issues.

🔴 Blocking: None.

💬 Non-blocking:
- 🔵 Suggestion: If SHA pinning is intended as a repository-wide supply-chain policy, consider also pinning `actions/checkout@v4` and `actions/setup-go@v5` in `.github/workflows/ci.yml:24`, `:25`, `:36`, `:37`, `:47`, and `:48`. Current tag usage is common and not a blocker.

✅ Highlights:
- `.github/workflows/ci.yml:17` matches the module’s `go 1.20` declaration.
- `.github/workflows/octo-pr-feed.yml:4` uses `pull_request_target` without checking out or executing PR code, and `.github/workflows/octo-pr-feed.yml:7` restricts default token permissions.
- `.github/workflows/pr-merged-done.yml:7` correctly documents that `GITHUB_TOKEN` is not used while passing `PROJECT_TOKEN` explicitly.
- `.github/workflows/issue-welcome.yml:9` scopes the caller job token to `issues: write`, which matches the workflow’s purpose.

Verified locally with Go 1.20.14:
- `CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build ./...`
- `go test -race -count=1 -timeout 5m ./...`
- `go vet ./...`

lml2468 (commented):
Re-review at `0c1678e` (prev: `94bbdfb`).

**Previous blocker — RESOLVED** ✅
`.github/workflows/pr-merged-done.yml:12`: `@main` → `@3d90d99236e0427a4fbe5b8f25b209ac0b799b3e` (SHA pinned). Verified the target commit exists in `Mininglamp-OSS/.github`.

**New issues in incremental diff:** None.

CI: Build ✅ / Test ✅ / Vet ✅

Verdict: **Would APPROVED** — all prior blockers resolved, no new issues. (Submitted as COMMENTED due to author = authenticated account constraint.)

PR #7, opened 2026-05-14
ci: add auto-add-to-project workflow

lml2468:
## Summary

Add `auto-add-to-project.yml` workflow: automatically adds new issues and PRs to [Octo Board (Project #2)](https://github.com/orgs/Mininglamp-OSS/projects/2) when they are opened.

## Why

Previously this was only handled for the `.github` repo itself. This extends coverage to all org repositories.

## Details

- Trigger: `issues: [opened]` + `pull_request_target: [opened]`
- Action: `actions/add-to-project@v1.0.2` (pinned commit hash)
- Requires: `PROJECT_TOKEN` org secret (already configured)
- `permissions: {}` — GITHUB_TOKEN not used

### DIFF
--- a/.github/workflows/auto-add-to-project.yml
+++ b/.github/workflows/auto-add-to-project.yml
@@ -1,0 +1,19 @@
+name: Auto-add to Octo Board
+
+on:
+  issues:
+    types: [opened]
+  pull_request_target:
+    types: [opened]
+
+# GITHUB_TOKEN not used; PROJECT_TOKEN (PAT) is used instead
+permissions: {}
+
+jobs:
+  add-to-project:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/add-to-project@244f685bbc3b7adfa8466e08b698b5577571133e  # v1.0.2
+        with:
+          project-url: https://github.com/orgs/Mininglamp-OSS/projects/2
+          github-token: ${{ secrets.PROJECT_TOKEN }}


yujiawei (approved):
# Code Review — PR #7 (octo-lib)

## Summary

Adds `.github/workflows/auto-add-to-project.yml` (19 lines, single new file) to auto-add newly opened issues and PRs to the org-level **Octo Board (Project #2)**, extending behavior previously scoped to the `.github` repo to all org repositories.

**Verdict: Approve.** The workflow is small, idempotent, and follows GitHub's recommended security patterns for project automation. A few optional polish suggestions below — none are blockers.

## 1. Verification

| Check | Result | Evidence |
|---|---|---|
| Action pinned to commit SHA | ✅ | `actions/add-to-project@244f685bbc3b7adfa8466e08b698b5577571133e` (line 16) |
| Pinned SHA matches the `v1.0.2` comment | ✅ | `git/refs/tags/v1.0.2` resolves to `244f685bbc3b7adfa8466e08b698b5577571133e` (verified against the action's GitHub API) |
| `permissions:` declared explicitly | ✅ | `permissions: {}` (line 10) — empty set, GITHUB_TOKEN granted no scopes |
| Token choice appropriate | ✅ | Uses `secrets.PROJECT_TOKEN` (a PAT) because GITHUB_TOKEN cannot write to **org-level** Projects v2; this is the documented pattern |
| `pull_request_target` usage safe | ✅ | Only `types: [opened]` and no checkout of PR head → no execution of untrusted code from forks; secret exposure is limited to invoking the marketplace action with a project URL |
| Trigger filter narrow | ✅ | `[opened]` only (not `edited`/`reopened`/`synchronize`) — avoids redundant adds and minimizes secret-using runs |
| Workflow idempotent | ✅ | `add-to-project` is a no-op when the item is already on the board |

## 2. Findings

No P0/P1/P2 issues.

## 3. Optional polish (non-blocking)

1. **Concurrency guard.** Cheap insurance against duplicate runs (e.g., a webhook redelivery) and a clearer audit trail in the Actions UI:
   ```yaml
   concurrency:
     group: add-to-project-${{ github.event.issue.number || github.event.pull_request.number }}
     cancel-in-progress: false
   ```
   Skip if you'd rather keep the file at exactly 19 lines.

2. **`timeout-minutes`.** A small bound (e.g. `timeout-minutes: 5`) on the `add-to-project` job avoids the default 360-minute ceiling if GitHub's API ever stalls. Defensive only — the action itself is fast.

3. **Token rotation hygiene.** `PROJECT_TOKEN` is a PAT with `project` scope; consider documenting (in `README` or a comment in the workflow) who owns it, the rotation cadence, and the expected scope. Not a code concern, but it's the only non-obvious operational dependency this workflow introduces.

## 4. Additional observations

- The PR body notes "`PROJECT_TOKEN` org secret (already configured)" — worth confirming the secret is set at **org** level and made available to this repo, otherwise the workflow will fail silently on first trigger. Visible only at runtime, not from the diff.
- `pull_request_target` from forks is the right choice here (it's the only way the fork-opened PR's run can read `secrets.PROJECT_TOKEN`). The narrow `[opened]` filter plus no checkout makes this the safe variant of the pattern.

lml2468:
Closing: wrong approach, will use GitHub org Projects native auto-add instead.

PR #8, opened 2026-05-14
ci: add auto-add-to-project caller workflow

lml2468:
## Summary

Add `auto-add-to-project.yml` caller workflow: automatically adds new issues and PRs to [Octo Board (Project #2)](https://github.com/orgs/Mininglamp-OSS/projects/2) when they are opened.

Calls the org-level reusable workflow at `Mininglamp-OSS/.github/.github/workflows/auto-add-to-project.yml@main`.

## Dependency

Requires [Mininglamp-OSS/.github#12](https://github.com/Mininglamp-OSS/.github/pull/12) to be merged first.

### DIFF
--- a/.github/workflows/auto-add-to-project.yml
+++ b/.github/workflows/auto-add-to-project.yml
@@ -1,0 +1,16 @@
+# Automatically add new issues and PRs to the org-level GitHub Project.
+name: Auto add to project
+
+on:
+  issues:
+    types: [opened]
+  pull_request_target:
+    types: [opened]
+
+permissions: {}
+
+jobs:
+  add-to-project:
+    uses: Mininglamp-OSS/.github/.github/workflows/auto-add-to-project.yml@main
+    secrets:
+      PROJECT_TOKEN: ${{ secrets.PROJECT_TOKEN }}
--- a/.github/workflows/auto-add-to-project.yml
+++ b/.github/workflows/auto-add-to-project.yml
@@ -11,6 +11,6 @@ permissions: {}
 
 jobs:
   add-to-project:
-    uses: Mininglamp-OSS/.github/.github/workflows/auto-add-to-project.yml@main
+    uses: Mininglamp-OSS/.github/.github/workflows/auto-add-to-project.yml@239d0b47cfe1947d5e6de826e795feac4ed90562
     secrets:
       PROJECT_TOKEN: ${{ secrets.PROJECT_TOKEN }}


yujiawei (dismissed):
# Code Review — PR #8 (octo-lib)

## Summary

A 16-line caller workflow that delegates "add issue/PR to org project board" to the org-level reusable workflow at `Mininglamp-OSS/.github`. The change is minimal, well-formed, and follows the same caller pattern used by the other org reusable workflows (`issue-welcome`, `octo-issue-feed`, `octo-pr-feed`). I'm approving with a few non-blocking observations.

**Verdict:** APPROVED — non-blocking notes below.

## 1. Verification

| Check | Result |
| --- | --- |
| YAML is syntactically valid and uses the documented `workflow_call` shape | ✅ |
| `permissions: {}` drops all default `GITHUB_TOKEN` scopes at the caller | ✅ (`.github/workflows/auto-add-to-project.yml:10`) |
| Secret contract matches the callee — caller forwards `PROJECT_TOKEN`, callee declares `secrets.PROJECT_TOKEN: required: true` | ✅ |
| `pull_request_target` does not check out PR-controlled code anywhere in the call chain | ✅ (callee only runs `actions/add-to-project@<sha>` against a project URL) |
| Third-party action in callee is pinned to a commit SHA (`244f685b…` ≈ v1.0.2) | ✅ |
| Triggers `issues: [opened]` and `pull_request_target: [opened]` are appropriate for the "auto-add on creation" goal | ✅ |

## 2. Observations (P2 — non-blocking)

### 2.1 Reusable workflow is referenced at `@main` (mutable ref)

`.github/workflows/auto-add-to-project.yml:13`:

```yaml
uses: Mininglamp-OSS/.github/.github/workflows/auto-add-to-project.yml@main
```

`@main` is a moving target: any future commit on `Mininglamp-OSS/.github`'s default branch will execute against this repo's `PROJECT_TOKEN` (a PAT) on every issue / PR opened here. Because `.github` is org-internal and `PROJECT_TOKEN` is scoped to the project board (not repo write), the blast radius is small — but pinning to a tag (e.g. `@v1`) or full commit SHA is the more defensive default and matches what you already do for `actions/add-to-project` inside the callee.

Since you're rolling this out across 8 sub-repos, picking one style now and keeping all callers consistent is worth the upfront cost — chasing it later means 8 PRs.

If you intentionally want centrally-rolling updates without per-repo PRs, `@main` is fine; just call it out in the workflow header comment so future maintainers don't assume it's an oversight.

### 2.2 Trigger only fires on `opened`

`.github/workflows/auto-add-to-project.yml:5-8`:

```yaml
on:
  issues:
    types: [opened]
  pull_request_target:
    types: [opened]
```

This is the typical pattern, but worth confirming intent:

- Issues that are **transferred in** from another repo do not fire `opened` — they fire `transferred`. They will not be auto-added.
- Issues / PRs **reopened** after being closed will not be re-added (the action is idempotent for items already on the board, but if the item was manually removed from the project, reopen won't put it back).

If the desired behaviour is "if it lives in this repo, it's on the board," consider also listening for `reopened` (and `transferred` for issues). Otherwise this is fine.

### 2.3 Merge ordering

The PR body correctly notes the dependency on `Mininglamp-OSS/.github#12`. Until that lands and `auto-add-to-project.yml` exists on `.github@main`, this workflow's first run will fail with a "workflow not found" error. Just confirm the merge order and that `PROJECT_TOKEN` is configured as a secret (org-level or repo-level) accessible to this repo before the first issue/PR is opened post-merge.

## 3. Suggestions

- **Optional:** add a one-line header comment pointing to the reusable workflow source, e.g. `# See Mininglamp-OSS/.github/.github/workflows/auto-add-to-project.yml`. Future maintainers searching this repo for "where does the project automation actually live?" will thank you.
- **Optional:** if you go with SHA pinning for the reusable workflow, add a short note about how you intend to bump it (manual PR vs. Dependabot's `github-actions` ecosystem, which does support reusable workflow refs).

## 4. Additional findings

None. The change is appropriately minimal — no scope creep, no unrelated edits.

Jerry-Xin (changes requested):
Summary: The PR is relevant to this repository, but the new privileged caller workflow should not merge while it references a mutable reusable workflow branch.

🔴 Blocking

- 🟡 Warning: [`.github/workflows/auto-add-to-project.yml`](/private/tmp/pr-review-kMlPFW/repo/.github/workflows/auto-add-to-project.yml:14) uses `Mininglamp-OSS/.github/.github/workflows/auto-add-to-project.yml@main` while running on `pull_request_target` and passing `PROJECT_TOKEN` at lines 7 and 16. This gives a mutable branch ref access to a project-scoped secret in a privileged event context. Pin the reusable workflow to an audited commit SHA after `Mininglamp-OSS/.github#12` lands, matching the safer pattern already used by other privileged workflows in this repo.

💬 Non-blocking

- None.

✅ Highlights

- `permissions: {}` is appropriate because this workflow appears to rely on the explicit `PROJECT_TOKEN` rather than broad `GITHUB_TOKEN` permissions.
- The trigger scope is narrow: only newly opened issues and PRs are handled.

lml2468 (commented):
## Verdict: CHANGES_REQUESTED (submitted as COMMENT due to author constraint)

**P0 — Reusable workflow no longer exists on `@main`**

`.github/workflows/auto-add-to-project.yml:14` calls:
```yaml
uses: Mininglamp-OSS/.github/.github/workflows/auto-add-to-project.yml@main
```

This file was **deleted from `main`** in commit `d5ab928c` (Mininglamp-OSS/.github#11 — "chore: remove auto-add-to-project workflow"). The workflow will fail at runtime with a 404.

The file still exists on branch `feat/restore-auto-add-reusable` and at the pre-deletion commit `3d90d992`. Options:

1. **Merge `feat/restore-auto-add-reusable` first** to restore the reusable workflow, then this caller works.
2. **Pin to `3d90d992`** — but fragile since the file was intentionally removed.

This PR cannot merge as-is because the reusable target does not exist.

**Non-blocking (agrees with Allen's finding):**
- `@main` should be SHA-pinned for supply-chain safety once the target is restored.

**Highlights:**
- `permissions: {}`, trigger pair, and caller structure are all correct.

yujiawei (approved):
# Code Review — PR #8 (octo-lib)

## Summary

Adds a 16-line caller workflow that delegates issue/PR-to-project automation to the org-level reusable workflow at `Mininglamp-OSS/.github/.github/workflows/auto-add-to-project.yml`, pinned to commit `239d0b4`. The change is small, focused, and structurally sound. One configuration risk worth confirming before relying on this in production (see P1 below).

## 1. Verification

| Check | Result | Evidence |
|---|---|---|
| Pinned SHA exists upstream and matches the intended file | ✅ | `239d0b47cfe1947d5e6de826e795feac4ed90562` is the merge commit of `Mininglamp-OSS/.github#12` into `main` (merged 2026-05-15T00:03:07Z). The file at that ref declares `workflow_call` with `secrets.PROJECT_TOKEN` required, matching the caller's `secrets:` block. |
| Dependency PR merged before this PR ships | ✅ | `Mininglamp-OSS/.github#12` is `MERGED` and the head commit (`5843021…`) of this PR was authored after that merge. |
| `pull_request_target` token-exfiltration risk mitigated | ✅ | The upstream reusable workflow does **not** invoke `actions/checkout` and does not run any code from the PR head ref — only `actions/add-to-project@244f685` operating on metadata. Forks cannot reach `PROJECT_TOKEN`. The upstream file even includes an explicit security comment forbidding `actions/checkout` here. |
| Mutable `@main` → pinned SHA fix is correct | ✅ | Force-commit `5843021` replaces `@main` with `@239d0b47cfe1947d5e6de826e795feac4ed90562`, eliminating the "rewrite `main` upstream → leak `PROJECT_TOKEN`" supply-chain class. |
| Caller `permissions: {}` correctness | ✅ | The called workflow uses only `PROJECT_TOKEN` (passed via `secrets:`) and never touches `GITHUB_TOKEN`, so empty default permissions is the right minimum. |
| YAML well-formed and triggers correct | ✅ | `issues: [opened]` + `pull_request_target: [opened]` mirrors the upstream direct-trigger surface; reusable-workflow `uses:` is at the correct `jobs.<id>` position. |

## 2. Findings

### P1 — `PROJECT_TOKEN` may not be available to this repo at runtime

The reusable workflow declares `secrets.PROJECT_TOKEN` as `required: true`. In octo-lib, only `OCTO_BOT_TOKEN` is currently registered as a repository-level Actions secret (no `PROJECT_TOKEN`). For the workflow to succeed, `PROJECT_TOKEN` must be set as either an organization-level secret with octo-lib in scope, or as a repo-level secret here.

Empirical signal that this hasn't been done org-wide yet: in `Mininglamp-OSS/.github` itself, every recent run of the reusable workflow (direct-trigger path, last 5 runs since 2026-05-13) has failed with:

```
Input required and not supplied: github-token
```

That error originates inside `actions/add-to-project`, where `github-token: ${{ secrets.PROJECT_TOKEN }}` resolved to an empty string. If `PROJECT_TOKEN` is not configured at the org (or for this repo), this PR will merge cleanly but every triggered run will fail the same way, and any new issue/PR opened in octo-lib will surface a red Actions notification to its author.

**Recommendation:** before merging (or immediately after), confirm `PROJECT_TOKEN` is configured as an org secret with `Mininglamp-OSS/octo-lib` included in the repository scope, and then trigger one issue/PR to validate. If it's a repo-level secret pattern instead, add it here. This is a deployment concern, not a code defect — the PR itself is correct.

### P2 — None.

### P3 / informational

- **Node.js 20 deprecation in transitive action.** The upstream reusable workflow uses `actions/add-to-project@v1.0.2`, which still ships on Node 20. GitHub will force Node 24 starting 2026-06-02; if `add-to-project` doesn't release a Node 24-compatible version by then, runs may break. Not actionable in this PR — the fix belongs upstream in `.github`. Flagging for awareness.
- **No caller-side timeout.** The called workflow already sets `timeout-minutes: 5` on its job, so this is fine. Worth noting only because the caller has no defense if the upstream were to remove that setting; not worth changing here.

## 3. Suggestions

None blocking. Optional, only if you want belt-and-suspenders:

- Add a one-line comment in the caller pointing to the upstream PR/SHA and the rationale for the pin, so future readers don't have to dig through git blame to understand why the SHA isn't `@main`. Example:
  ```yaml
  # Pinned to Mininglamp-OSS/.github#12 (commit 239d0b4) instead of @main —
  # pull_request_target receives PROJECT_TOKEN, so a moving ref upstream
  # would be a supply-chain risk. Bump deliberately, with review.
  ```
  Optional; the commit message already explains it.

## 4. Verdict

**APPROVED with one operational follow-up.** Code is correct, security posture is right (pinned SHA + no checkout under `pull_request_target` + minimum permissions). The only gate before this is actually useful is confirming `PROJECT_TOKEN` is provisioned for octo-lib — same gap that's currently breaking the workflow inside the `.github` repo. Resolving that is a secrets-config task, not a code change.

Jerry-Xin (approved):
Summary: The PR is relevant repository automation for `octo-lib`, and the new caller workflow is ready to merge.

🔴 Blocking: None.

💬 Non-blocking:
- 🔵 Suggestion `.github/workflows/auto-add-to-project.yml:14`: The PR description says the workflow calls `@main`, but the actual workflow pins a commit SHA. The pinned SHA is preferable for security and matches existing workflow practice, so only the description should be corrected if desired.

✅ Highlights:
- `.github/workflows/auto-add-to-project.yml:7`: Using `pull_request_target` is appropriate here because adding forked PRs to an org project requires access to `PROJECT_TOKEN`.
- `.github/workflows/auto-add-to-project.yml:10`: `permissions: {}` correctly minimizes the default `GITHUB_TOKEN` permissions since the workflow passes an explicit project token.
- `.github/workflows/auto-add-to-project.yml:14`: The reusable workflow is pinned to a full commit SHA, reducing supply-chain risk compared with tracking a mutable branch.

lml2468 (commented):
Summary

Re-reviewed current head 584302103f1a45fa91b7249bf2c68bd87bddf399. Both previous findings are resolved: the caller now references an existing reusable workflow file at Mininglamp-OSS/.github commit 239d0b47cfe1947d5e6de826e795feac4ed90562, and the mutable @main reference has been replaced with a full commit SHA while passing PROJECT_TOKEN.

I attempted to submit this as APPROVED, but GitHub rejected it because the authenticated CLI user is lml2468, the PR author. This is an approval-equivalent review result from my side; no blockers remain.

Blocking issues

None.

Non-blocking suggestions

The PR description still says the reusable workflow is called at @main, while the code correctly uses the pinned SHA. Please update the description when convenient to avoid confusion.

Highlights

- permissions: {} keeps the default GITHUB_TOKEN privileges minimized.
- The pinned reusable workflow avoids the prior supply-chain risk from a moving branch ref.
- The pinned upstream workflow exists at the referenced SHA and does not check out untrusted PR code under pull_request_target.
- Build, Vet, and Test are green, and the PR is mergeable.

PR #9, opened 2026-05-15
feat(config): add NewPersonalMsgSendReq builder for authoritative payload.space_id

### LINKED ISSUE #33: feat(config): add Redis TLS options for AWS ElastiCache / Azure Cache compatibility
## Background

The `DB` config in `config/config.go` currently exposes only `RedisAddr` and `RedisPass` for Redis connections:

```go
DB struct {
    ...
    RedisAddr          string // redis 地址
    RedisPass          string // redis 密码
    AsynctaskRedisAddr string // 异步任务的 redis 地址
}
```

This is sufficient for plaintext Redis (self-hosted, intranet), but **cannot connect to managed Redis services that enforce TLS (in-transit encryption)**, including:

- **AWS ElastiCache for Redis** with in-transit encryption enabled
- **Azure Cache for Redis** (TLS required by default)
- **GCP Memorystore for Redis** with TLS
- Any self-hosted Redis behind stunnel / TLS proxy

Downstream projects (e.g. `octo-server`) need to migrate to AWS ElastiCache but are currently blocked because there is no way to express TLS connection options through `octo-lib` config.

## Proposed change

Add three new optional fields to the `DB` struct, all defaulting to zero values so existing deployments are unaffected:

```go
DB struct {
    ...
    RedisAddr                  string
    RedisPass                  string
    RedisTLS                   bool   // enable TLS (rediss://)
    RedisTLSInsecureSkipVerify bool   // skip cert verification (test / self-signed)
    RedisTLSCAFile             string // optional custom CA cert path
    AsynctaskRedisAddr         string
}
```

Corresponding viper bindings (alongside existing `db.redisAddr` / `db.redisPass`):

```go
c.DB.RedisTLS                   = c.getBool(\"db.redisTLS\", c.DB.RedisTLS)
c.DB.RedisTLSInsecureSkipVerify = c.getBool(\"db.redisTLSInsecureSkipVerify\", c.DB.RedisTLSInsecureSkipVerify)
c.DB.RedisTLSCAFile             = c.getString(\"db.redisTLSCAFile\", c.DB.RedisTLSCAFile)
```

## Backward compatibility

- All new fields default to `false` / `\"\"`, so existing configs without these keys behave identically.
- No changes to existing field names or types.
- Downstream consumers opt in by setting `db.redisTLS: true` in their config.

## Out of scope

- This issue covers **config surface only**. Actually building the `*tls.Config` and wiring it into the `go-redis` client is the consumer's responsibility (e.g. `octo-server` will translate these flags into `redis.Options.TLSConfig` in its own wrapper).
- Not upgrading the `go-redis` client version — `go-redis/redis v6` already supports `Options.TLSConfig`, so no SDK upgrade is required for TLS support.

## Follow-up PR

I will open a PR implementing the fields and viper bindings once this issue is acknowledged.

yujiawei:
## Summary

Adds `config.NewPersonalMsgSendReq(...)` — a defense-in-depth builder for
PERSONAL (channel_type=1) `MsgSendReq` construction.

The builder encodes the authoritative `payload.space_id` contract from
[Mininglamp-OSS/octo-server PR#35 (Fixes #33)][1] into a single octo-lib API so
PERSONAL DM dispatchers no longer need to remember to call enrichment helpers
manually.

## Contract

| `senderSpaceID` | `payload["space_id"]` outcome |
|---|---|
| non-empty | overwritten with `senderSpaceID` (server is authoritative) |
| empty | **stripped** (fail-closed) |

The fail-closed empty branch closes the YUJ-660 High-3 fail-open hole:
dispatchers without Space context (orphan bot, missing X-Space-ID middleware,
non-Space deployment) MUST NOT trust a client-supplied value, since WuKongIM
has no Space concept on PERSONAL channels and the receiver-side SpaceFilter
has no other authoritative signal to fall back on.

`ChannelType` is always set to `ChannelTypePerson`; callers MUST verify the
channel is in fact PERSONAL before invoking. The optional `PersonalMsgOptions`
struct exposes `Header` / `Setting` / `StreamNo` / `Subscribers` without
re-exposing `payload` or `channel_type`.

## Tests

`config/msg_personal_test.go` covers:
- override (client value replaced)
- inject (no client value)
- strip (empty senderSpaceID + client value present → fail-closed)
- leave-absent (both empty)
- nil-payload materialisation
- optional-field propagation
- `channel_type=Person` invariant

```
$ go test ./config/...
ok  	github.com/Mininglamp-OSS/octo-lib/config	2.5s
```

`go build ./...` and `go vet ./...` clean.

## Why a builder, not a free function

A naked free function on `[]byte` payload was considered (matches issue body
proposal) but rejected: the authoritative override needs to mutate the JSON
map, and round-tripping through `[]byte → map → []byte` is what the helper is
hiding from callers. The map-in / `*MsgSendReq`-out signature lets the
encoding live in one place.

## Out-of-scope (handled by the matching octo-server PR)

- Migration of the ~10 PERSONAL `MsgSendReq{}` call sites listed in the issue.
- CI lint that rejects new direct `MsgSendReq{... ChannelType: ChannelTypePerson ...}`
  literals outside the builder.
- Per-call-site `senderSpaceID` resolution (re-uses the existing
  `enrichBotPayloadWithSpaceID` / `spacepkg.GetSpaceID` / robot resolver
  helpers added in PR#35).

## Refs

- Refs Mininglamp-OSS/octo-server#37
- Background: Mininglamp-OSS/octo-server#35 (closes #33), R3/R4 review
  iterations recorded in YUJ-644 / YUJ-660.

[1]: https://github.com/Mininglamp-OSS/octo-server/pull/35

### DIFF
--- a/config/msg_personal.go
+++ b/config/msg_personal.go
@@ -1,0 +1,77 @@
+package config
+
+import (
+	"github.com/Mininglamp-OSS/octo-lib/common"
+	"github.com/Mininglamp-OSS/octo-lib/pkg/util"
+)
+
+// PersonalMsgOptions carries the optional MsgSendReq fields beyond the
+// mandatory authoritative-payload contract enforced by NewPersonalMsgSendReq.
+//
+// Only fields a PERSONAL DM dispatcher legitimately controls are exposed.
+// channel_type is set to ChannelTypePerson by the builder unconditionally;
+// payload is encoded by the builder after authoritative space_id semantics
+// have been applied.
+type PersonalMsgOptions struct {
+	Header      MsgHeader
+	Setting     uint8
+	StreamNo    string
+	Subscribers []string
+}
+
+// NewPersonalMsgSendReq returns a *MsgSendReq for a PERSONAL (channel_type=1)
+// dispatch with authoritative payload.space_id semantics applied to payload
+// before encoding.
+//
+// Contract:
+//   - senderSpaceID != ""  → payload["space_id"] = senderSpaceID (server overrides
+//     any client-supplied value; this is the only authoritative source the
+//     receiver-side SpaceFilter can trust on PERSONAL channels because WuKongIM
+//     has no Space concept on channel_type=1).
+//   - senderSpaceID == ""  → payload["space_id"] is stripped (fail-closed).
+//     The empty case means the dispatcher has no Space context (orphan bot,
+//     non-Space deployment, missing X-Space-ID middleware); we MUST NOT trust
+//     a client-supplied value in that branch — see Mininglamp-OSS/octo-server
+//     PR#35 R3 review (YUJ-660 High-3) for the fail-open hole this closes.
+//
+// Channel type is always set to ChannelTypePerson; callers MUST verify the
+// channel is in fact PERSONAL before invoking this builder. Use the regular
+// MsgSendReq literal for GROUP / COMMUNITY_TOPIC / RTC / etc.
+//
+// payload may be nil; an empty map is materialised. The map is mutated in
+// place (consistent with octo-server's enrichPayloadWithSpaceID helpers) and
+// then JSON-encoded into MsgSendReq.Payload.
+//
+// This builder is the defense-in-depth layer for Mininglamp-OSS/octo-server#37
+// (PR#35 follow-up Medium-1). New PERSONAL DM call sites added in either
+// repository SHOULD use this builder so payload.space_id correctness becomes
+// a property of construction rather than caller diligence.
+func NewPersonalMsgSendReq(
+	channelID string,
+	fromUID string,
+	payload map[string]interface{},
+	senderSpaceID string,
+	opts PersonalMsgOptions,
+) *MsgSendReq {
+	if payload == nil {
+		payload = make(map[string]interface{})
+	}
+	if senderSpaceID != "" {
+		payload["space_id"] = senderSpaceID
+	} else {
+		// Fail-closed: server has no authoritative Space context, so any
+		// client-supplied space_id is untrusted. Strip it. Receivers will
+		// fall through to their own (cache-warmed) SpaceFilter logic.
+		delete(payload, "space_id")
+	}
+	return &MsgSendReq{
+		Header:      opts.Header,
+		Setting:     opts.Setting,
+		FromUID:     fromUID,
+		ChannelID:   channelID,
+		ChannelType: common.ChannelTypePerson.Uint8(),
+		StreamNo:    opts.StreamNo,
+		Subscribers: opts.Subscribers,
+		Payload:     []byte(util.ToJson(payload)),
+	}
+}
--- a/config/msg_personal_test.go
+++ b/config/msg_personal_test.go
@@ -1,0 +1,145 @@
+package config
+
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/Mininglamp-OSS/octo-lib/common"
+)
+
+func decodePayload(t *testing.T, b []byte) map[string]interface{} {
+	t.Helper()
+	out := map[string]interface{}{}
+	if err := json.Unmarshal(b, &out); err != nil {
+		t.Fatalf("payload not JSON: %v (raw=%s)", err, string(b))
+	}
+	return out
+}
+
+// senderSpaceID non-empty → server overrides any client-supplied value.
+func TestNewPersonalMsgSendReq_OverridesClientWhenSenderSpaceIDProvided(t *testing.T) {
+	req := NewPersonalMsgSendReq(
+		"peer_uid",
+		"sender_uid",
+		map[string]interface{}{
+			"content":  "hi",
+			"type":     1,
+			"space_id": "client_forged_space",
+		},
+		"server_authoritative_space",
+		PersonalMsgOptions{Header: MsgHeader{RedDot: 1}},
+	)
+	if req == nil {
+		t.Fatal("nil req")
+	}
+	if req.ChannelType != common.ChannelTypePerson.Uint8() {
+		t.Errorf("ChannelType=%d, want PERSONAL=%d", req.ChannelType, common.ChannelTypePerson.Uint8())
+	}
+	if req.ChannelID != "peer_uid" || req.FromUID != "sender_uid" {
+		t.Errorf("identity fields not preserved: %+v", req)
+	}
+	got := decodePayload(t, req.Payload)
+	if got["space_id"] != "server_authoritative_space" {
+		t.Errorf("space_id=%v, want server_authoritative_space (client value MUST be overridden)", got["space_id"])
+	}
+	if got["content"] != "hi" {
+		t.Errorf("content not preserved, got %v", got["content"])
+	}
+}
+
+// senderSpaceID non-empty → injects when payload has no space_id at all.
+func TestNewPersonalMsgSendReq_InjectsWhenAbsent(t *testing.T) {
+	req := NewPersonalMsgSendReq(
+		"peer_uid",
+		"sender_uid",
+		map[string]interface{}{"content": "hi"},
+		"server_authoritative_space",
+		PersonalMsgOptions{},
+	)
+	got := decodePayload(t, req.Payload)
+	if got["space_id"] != "server_authoritative_space" {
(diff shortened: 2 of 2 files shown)


Jerry-Xin (approved):
Summary: The PR is in scope for `octo-lib` and adds a focused `config` helper that matches the stated PERSONAL message payload contract.

🔴 Blocking

None.

💬 Non-blocking

🔵 Suggestion: [config/msg_personal.go](/private/tmp/pr-review-uUQjFB/repo/config/msg_personal.go:75) uses `util.ToJson`, which silently returns an empty string on JSON marshal failure. This is consistent with existing project style, but because this is a new public builder, consider whether a future error-returning variant would be useful for unsupported payload values.

🔵 Suggestion: [config/msg_personal.go](/private/tmp/pr-review-uUQjFB/repo/config/msg_personal.go:74) passes `Subscribers` through without copying. That is normal for this codebase, but callers can still mutate the slice after construction and affect the request.

✅ Highlights

The relevance gate passes: this directly extends `octo-lib` message construction functionality.

The authoritative `space_id` behavior is implemented clearly: override when `senderSpaceID` is present, strip when absent, and always force `ChannelTypePerson`.

Tests cover the key contract branches, nil payload materialization, optional field propagation, and channel type invariant.

Verification passed: `go test ./...` and `go vet ./...` both completed successfully.

yujiawei (commented):
# Code Review — PR #9 (octo-lib)

## Verdict

**APPROVED (posted as COMMENTED — GitHub blocks self-approval since the reviewer account matches the PR author)** with three minor observations. The contract is correctly encoded, branches are exhaustively tested, and `go build` / `go vet` / `go test ./config/...` all pass clean against `d21edd1`.

## 1. Verification

| Item | Status | Evidence |
|---|---|---|
| `payload["space_id"] = senderSpaceID` when non-empty (override client-supplied) | ✅ | `config/msg_personal.go:64-65` |
| `delete(payload, "space_id")` when empty (fail-closed) | ✅ | `config/msg_personal.go:69` |
| `ChannelType` set to `ChannelTypePerson.Uint8()` unconditionally | ✅ | `config/msg_personal.go:77`; `common/constant.go:19` (PERSONAL = 1 via iota) |
| nil payload materialised to empty map before mutation | ✅ | `config/msg_personal.go:60-62` |
| Optional fields (`Header` / `Setting` / `StreamNo` / `Subscribers`) propagate verbatim | ✅ | `config/msg_personal.go:71-79` |
| Tests cover all four authoritative-payload branches plus invariants | ✅ | `config/msg_personal_test.go` — 7 tests, including the strip-on-empty-sender case (`TestNewPersonalMsgSendReq_StripsClientSpaceIDWhenSenderEmpty`, lines 71-92) |
| `go build ./config/...` | ✅ | clean |
| `go vet ./config/...` | ✅ | clean |
| `go test ./config/...` | ✅ | `ok  github.com/Mininglamp-OSS/octo-lib/config  0.014s` |

The contract documented in the PR body matches the code line-for-line. The `delete` call closes the YUJ-660 fail-open hole as claimed: a dispatcher with no Space context can no longer pass a client-supplied `space_id` through to the receiver-side `SpaceFilter`.

## 2. Findings

### P2 — In-place mutation of the caller's payload map

`config/msg_personal.go:64-69` mutates the input `payload` map directly (writing or deleting `space_id`). The doc comment notes this ("The map is mutated in place"), but since the builder is positioned as the canonical PERSONAL DM construction path, callers who reuse a payload map across multiple sends — for example a templated scaffold modified per-recipient — will observe the side effect.

Two reasonable options:

1. **Copy on entry** (one-line shallow clone before mutation) — bullet-proof, ~zero overhead for a typical small payload, and lets the doc just say "payload is not modified".
2. **Keep current behavior**, but escalate the doc note from a parenthetical to a **WARNING** block, since the in-place semantics are easy to miss at a glance.

Not blocking — payloads are constructed per-call in nearly every existing site I checked (`msg_group.go`, `msg.go:169`, `msg.go:190`) — but worth deciding before more callers adopt the builder.

### P2 — `util.ToJson` swallows marshal errors silently

`pkg/util/json.go:9-15` returns `""` on `json.Marshal` failure. `config/msg_personal.go:79` wraps the result in `[]byte(util.ToJson(payload))`, so a marshal failure produces `Payload = []byte("")` — not nil, not an error, just an empty body that will be sent as if it were valid.

This is consistent with the existing repo pattern (every `MsgSendReq` literal in `msg.go` and `msg_group.go` does the same), so it's not a regression — but a builder that promises "authoritative payload semantics" inherits the limitation. Two paths if you want to tighten:

- Return `(*MsgSendReq, error)` and surface the marshal error, OR
- Use `json.Marshal` directly inside the builder and panic / log on failure.

Lowest-risk: leave as-is and track separately. Realistic marshal failures on `map[string]interface{}` with simple JSON values are rare.

### P2 — Defense-in-depth gap: existing PERSONAL call sites in this same library

The PR body states "migration of the ~10 PERSONAL `MsgSendReq{}` call sites … is out-of-scope (handled by the matching octo-server PR)". That accounts for the octo-server side, but **two PERSONAL call sites already exist in octo-lib itself**:

- `config/msg.go:169` — `SendFriendApply` (`ChannelType: common.ChannelTypePerson.Uint8()`)
- `config/msg.go:190` — `SendFriendSure` (`ChannelType: common.ChannelTypePerson.Uint8()`)

Neither carries a `space_id` field today, so there is no current correctness bug. But the defense-in-depth narrative ("PERSONAL DM dispatchers no longer need to remember to call enrichment helpers manually") has a small asymmetry: the builder lives in the same package as two non-migrated PERSONAL constructions. If a future change adds Space-aware fields to friend-apply/sure payloads without going through the builder, the bug class returns silently.

Suggestion: either migrate `SendFriendApply` / `SendFriendSure` in this PR (small, mechanical), or open a follow-up issue tracking specifically the **octo-lib** internal call sites — distinct from the octo-server migration tracked in the matching PR.

## 3. Smaller observations (Nits, non-blocking)

- The doc rationale block on `NewPersonalMsgSendReq` is excellent — captures the *why* (WuKongIM has no Space concept on PERSONAL, receiver-side SpaceFilter has no other authoritative signal). Keep this style for future builders.
- `PersonalMsgOptions` cleanly excludes `Payload` and `ChannelType` from the optional surface — the right call. This is what makes the "authoritative" claim defensible: callers cannot accidentally bypass the builder by reaching into the options struct.
- `decodePayload` test helper is reasonable; if you grow more PERSONAL builders, consider promoting to a `config/internal/testutil` to share with future tests.
- Test `TestNewPersonalMsgSendReq_OptionalFieldsPropagate` (lines 119-141) does not assert that `space_id` is correctly set in the same call — minor, but a single combined assertion would tighten it.

## 4. Summary

This is a clean, narrow, well-tested defense-in-depth addition. The authoritative-`space_id` contract is correctly mechanised, the fail-closed branch is explicitly tested, and the design rationale (builder over free function, options struct over variadic) is justified. Three P2 observations above are worth resolving before broader adoption — especially the in-place mutation note and the two non-migrated PERSONAL sites in this same package — but none rise to a blocker for merge.

lml2468 (approved):
Summary
This PR adds a focused constructor for PERSONAL DM `MsgSendReq` payloads and keeps the authoritative `payload.space_id` rule inside construction. I do not have any new blocking findings beyond the existing reviews.

Verification
I checked the PR head SHA directly. `MsgSendReq` has the fields this builder populates (`Header`, `Setting`, `FromUID`, `ChannelID`, `ChannelType`, `StreamNo`, `Subscribers`, `Payload`), `common.ChannelTypePerson` is the personal-channel enum used elsewhere for DMs, and the builder follows the repository's existing `[]byte(util.ToJson(payload))` payload encoding convention.

I also checked the nearby message constructors and the new tests. The tests cover override, injection, fail-closed stripping, nil payload materialization, optional field propagation, and the PERSONAL channel-type invariant. The signature matches the existing direct-`MsgSendReq` construction style: required routing identity is passed explicitly, while less common send fields stay in an options struct.

New Findings
None. I specifically checked empty `channelID` / `fromUID` validation and do not see a PR-specific blocker: current `MsgSendReq` construction does not validate these fields, and some existing notification paths intentionally omit `FromUID`. Making this builder validate identity fields would be a wider API contract change rather than necessary defense for the `payload.space_id` issue.

Verdict
Approved.

PR #11, opened 2026-05-15
chore(workflows): pin reusable workflow refs to @v1

lml2468:
## Summary

Pin all reusable workflow references to `@v1` for stable, version-locked behavior.

**Before:** Mixed `@main` and hardcoded commit SHAs
**After:** All callers use `Mininglamp-OSS/.github/...@v1`

## Why

The `.github` repo has been tagged `v1` as the stable baseline. Using a version tag
ensures callers get a predictable, reviewed version of the shared workflows.

No caller-side logic changes — this is a reference update only.

### DIFF
--- a/.github/workflows/auto-add-to-project.yml
+++ b/.github/workflows/auto-add-to-project.yml
@@ -11,6 +11,6 @@ permissions: {}
 
 jobs:
   add-to-project:
-    uses: Mininglamp-OSS/.github/.github/workflows/auto-add-to-project.yml@239d0b47cfe1947d5e6de826e795feac4ed90562
+    uses: Mininglamp-OSS/.github/.github/workflows/auto-add-to-project.yml@v1
     secrets:
       PROJECT_TOKEN: ${{ secrets.PROJECT_TOKEN }}
--- a/.github/workflows/issue-welcome.yml
+++ b/.github/workflows/issue-welcome.yml
@@ -8,4 +8,4 @@ jobs:
   welcome:
     permissions:
       issues: write
-    uses: Mininglamp-OSS/.github/.github/workflows/issue-welcome.yml@3d90d99236e0427a4fbe5b8f25b209ac0b799b3e
+    uses: Mininglamp-OSS/.github/.github/workflows/issue-welcome.yml@v1
--- a/.github/workflows/octo-issue-feed.yml
+++ b/.github/workflows/octo-issue-feed.yml
@@ -6,7 +6,7 @@ on:
 
 jobs:
   notify:
-    uses: Mininglamp-OSS/.github/.github/workflows/octo-issue-feed.yml@main
+    uses: Mininglamp-OSS/.github/.github/workflows/octo-issue-feed.yml@v1
     with:
       repo_name: ${{ github.event.repository.name }}
       issue_number: ${{ github.event.issue.number }}
--- a/.github/workflows/octo-pr-feed.yml
+++ b/.github/workflows/octo-pr-feed.yml
@@ -8,7 +8,7 @@ permissions: {}
 
 jobs:
   notify:
-    uses: Mininglamp-OSS/.github/.github/workflows/octo-pr-feed.yml@c6c51348fccac99161166dce1b52dec1063aeea2
+    uses: Mininglamp-OSS/.github/.github/workflows/octo-pr-feed.yml@v1
     with:
       repo_name: ${{ github.event.repository.name }}
       pr_number: ${{ github.event.pull_request.number }}


Jerry-Xin (changes requested):
This PR is in scope for `octo-lib`, but the `octo-pr-feed` caller can lose required token permissions after switching to `@v1`.

🔴 Blocking

🟡 Warning — `.github/workflows/octo-pr-feed.yml:7` and `.github/workflows/octo-pr-feed.yml:11`: the caller sets `permissions: {}`, but the `@v1` reusable `octo-pr-feed` workflow now checks out shared scripts and declares `contents: read`. GitHub only allows reusable workflows to maintain or reduce caller-provided `GITHUB_TOKEN` permissions, not elevate them, so the called workflow cannot recover `contents: read` from this caller configuration. This can break the new `@v1` target at its checkout step. Add job-level `permissions: contents: read` for `jobs.notify` or remove the empty top-level permissions block while keeping least privilege. ([docs.github.com](https://docs.github.com/en/actions/reference/workflows-and-actions/reusing-workflow-configurations))

💬 Non-blocking

🔵 Suggestion — `.github/workflows/octo-issue-feed.yml:8`: consider explicitly adding `permissions: contents: read` here too, since the reusable workflow also checks out shared scripts. It is not a regression from this PR because this caller already used the moving `@main` implementation, but making the requirement explicit avoids dependence on repository defaults.

✅ Highlights

The reusable workflow paths exist at `Mininglamp-OSS/.github@v1`, and the caller inputs/secrets match the `workflow_call` declarations. The `auto-add-to-project` and `issue-welcome` updates do not show caller-side compatibility issues.

yujiawei (approved):
# Code Review — PR #11 (octo-lib)

## Summary

Pure reference update: 4 caller workflows in `octo-lib/.github/workflows/` switch their `uses:` target from a mix of full commit SHAs and a moving `@main` ref to the new `Mininglamp-OSS/.github` `@v1` tag. No caller-side logic, inputs, or secrets change. +4/-4 lines across 4 files.

**Verdict: Approve.** Low-risk hygiene change. The `@v1` tag exists, contains all four target reusable workflows, has compatible `workflow_call` signatures, and includes the previously pinned commits as ancestors. Replacing `@main` (a moving branch ref) with a tag is a clear correctness improvement.

## 1. Verification

| Check | Result | Evidence |
|---|---|---|
| `v1` tag exists in `Mininglamp-OSS/.github` | ✅ | Annotated tag `v1` → commit `1aab7b304de30a61f1b1e950c66c72bd558a97c7` ("v1: stable reusable workflows baseline") |
| All 4 referenced reusable workflows exist at `@v1` | ✅ | `auto-add-to-project.yml`, `issue-welcome.yml`, `octo-issue-feed.yml`, `octo-pr-feed.yml` all resolvable under `.github/workflows/` at the `v1` ref |
| Previously pinned SHAs are ancestors of `v1` | ✅ | `239d0b4` (auto-add), `3d90d99` (issue-welcome), `c6c5134` (octo-pr-feed) all reachable from `v1` — version moves forward, not backward |
| `workflow_call` input/secret contracts match callers | ✅ | See per-file table below |
| `@main` → `@v1` migration | ✅ | `octo-issue-feed.yml` was previously on the moving `@main` ref; pinning to a tag eliminates surprise updates |

### Caller ↔ reusable-workflow contract alignment

| Caller (octo-lib) | Reusable workflow @ v1 | Inputs/secrets match? |
|---|---|---|
| `auto-add-to-project.yml` (passes `secrets.PROJECT_TOKEN`) | requires `secrets.PROJECT_TOKEN` | ✅ |
| `issue-welcome.yml` (no inputs/secrets) | `workflow_call:` with no inputs/secrets | ✅ |
| `octo-issue-feed.yml` (8 inputs + `OCTO_BOT_TOKEN`) | declares same 8 inputs + `OCTO_BOT_TOKEN` required | ✅ |
| `octo-pr-feed.yml` (11 inputs + `OCTO_BOT_TOKEN`) | declares same 11 inputs + `OCTO_BOT_TOKEN` required | ✅ |

### Behavior delta vs previously pinned SHAs

- **`octo-pr-feed.yml`**: `c6c5134` → `v1` includes the inline-Python → `_shared/scripts/octo_pr_feed_notify.py` refactor (PR #8 in `.github`) and the `job.workflow_sha`-locked checkout hardening (PR #14). Adds `permissions: contents: read` and `timeout-minutes: 5`. Caller-visible interface unchanged.
- **`auto-add-to-project.yml`** and **`issue-welcome.yml`**: previously pinned commits are the latest changes touching those files up to `v1`, so the behavior is effectively identical.
- **`octo-issue-feed.yml`**: moves off `@main` to a stable tag — strictly an improvement.

## 2. Findings

No P0/P1 issues.

### P2 — Tag mutability (nitpick / informational)

`@v1` is a Git tag, and Git tags are mutable: the upstream maintainer can force-push the `v1` ref to a different commit at any time, and downstream consumers will silently pick that up on next run. Full-SHA pinning was the strictly stronger guarantee.

This is a known trade-off and is consistent with how the org has structured the `v1` baseline (the tag's own annotation describes it as the "stable reviewed baseline"). Since `Mininglamp-OSS/.github` is a first-party, org-controlled repo and the rationale (predictable, reviewed shared workflows) is sound, this is acceptable. Worth mentioning so future readers don't assume `@v1` is immutable.

If a stronger guarantee is desired later, two options:
- Pin to the underlying commit (`@1aab7b304de30a61f1b1e950c66c72bd558a97c7`) and bump on each baseline release.
- Treat `@v1` as a major-version moving target (semver convention) and additionally enable a branch-protection / required-review rule on the `v1` ref in `Mininglamp-OSS/.github`.

No action requested for this PR.

### P2 — Out of scope, FYI

The `v1` annotation lists `octo-ci-status.yml` as part of the baseline, but `octo-lib` does not currently consume it. Not a defect of this PR; just worth confirming whether `octo-lib` intends to wire up CI-status notifications under the same versioning policy in a follow-up.

## 3. Recommendations

- ✅ Merge as-is. Clean, minimal, and improves the worst case (`@main` → `@v1`).
- (Optional, follow-up) Add a one-line note in `CONTRIBUTING.md` or the repo README stating the org's reusable-workflow versioning policy is `@v1`, so future workflow additions consistently use the same scheme rather than re-introducing SHA pins or `@main`.

## 4. Additional Observations

- The PR description's "no caller-side logic changes" claim was verified by diffing the `workflow_call` signatures of all four reusable workflows at `v1` against the inputs/secrets each caller passes — they line up exactly.
- The change is event-triggered: `auto-add-to-project.yml` and `octo-pr-feed.yml` use `pull_request_target`, so this PR's own merge will not trigger a self-test of the new refs (they fire on `opened`, not on push to `main`). The first real validation will be the next issue/PR opened post-merge. Low risk given the input contract is verified, but worth keeping an eye on the next issue/PR opened in `octo-lib`.

lml2468 (commented):
I cross-checked this against `Mininglamp-OSS/.github@v1`.

I agree with Allen's blocker. `octo-pr-feed.yml` previously called `octo-pr-feed.yml@c6c51348fccac99161166dce1b52dec1063aeea2`; that reusable workflow had no checkout step and only ran inline Python, so the caller's top-level `permissions: {}` was compatible with the old callee behavior. At `@v1`, the reusable workflow now declares `permissions: contents: read` and adds `actions/checkout` to fetch shared scripts from `Mininglamp-OSS/.github`. Since reusable workflow token permissions are constrained by the caller, this caller should grant `contents: read` before switching to `@v1` (job-level on `notify` is enough).

Other checks: the `v1` tag exists, and the four callers match the corresponding reusable workflow inputs/secrets. `octo-issue-feed.yml` currently relies on the repo default read token permission; making `contents: read` explicit there would be cleaner, but the clear regression is the PR feed caller that explicitly sets `{}`.

PR #12, opened 2026-05-15
fix(ci): upgrade Go 1.24, add lint, add CI status notifications, pin action SHAs

lml2468:
### 🟠 H-5: Add octo-ci-status.yml
Notify Octo IM `📦 octo-lib` channel when CI status changes on main branch.

### 🟡 M-2: Upgrade Go to 1.24
Go 1.20 is EOL. Upgrade to current stable 1.24.

### 🟡 M-5: Add golangci-lint job
octo-lib is the shared foundation for all Go services in the org — it deserves the highest code-quality bar. Add `golangci-lint v1.64.5` to match octo-server and octo-smart-summary.

### 🟡 M-4 alignment: Remove `needs: build` from test job
`go test` rebuilds anyway; run test/vet/lint in parallel.

### 🟢 L-1: Pin action SHAs
Pin `actions/checkout` and `actions/setup-go` to verified commit SHAs for supply-chain security.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  GO_VERSION: '1.24'
+  GO_VERSION: '1.25.x'
 
 jobs:
   build:
@@ -59,7 +59,7 @@ jobs:
         with:
           go-version: ${{ env.GO_VERSION }}
       - name: golangci-lint
-        uses: golangci/golangci-lint-action@55c2c1448f86e01eaae002a5a3a9624417608d84 # v6
+        uses: golangci/golangci-lint-action@1e7e51e771db61008b38414a730f564565cf7c20 # v9.2.0
         with:
-          version: v1.64.5
+          version: v2.12.2
           args: --timeout=5m
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,15 +14,15 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  GO_VERSION: '1.20'
+  GO_VERSION: '1.24'
 
 jobs:
   build:
     name: Build
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
-      - uses: actions/setup-go@v5
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+      - uses: actions/setup-go@40f1582b2485089dde7abd97c1529aa768e1baff # v5
         with:
           go-version: ${{ env.GO_VERSION }}
       - name: Build
@@ -31,10 +31,9 @@ jobs:
   test:
     name: Test
     runs-on: ubuntu-latest
-    needs: build
     steps:
-      - uses: actions/checkout@v4
-      - uses: actions/setup-go@v5
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+      - uses: actions/setup-go@40f1582b2485089dde7abd97c1529aa768e1baff # v5
         with:
           go-version: ${{ env.GO_VERSION }}
       - name: Test
@@ -44,9 +43,23 @@ jobs:
     name: Vet
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
-      - uses: actions/setup-go@v5
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+      - uses: actions/setup-go@40f1582b2485089dde7abd97c1529aa768e1baff # v5
         with:
           go-version: ${{ env.GO_VERSION }}
       - name: Vet
         run: go vet ./...
+
+  lint:
+    name: Lint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+      - uses: actions/setup-go@40f1582b2485089dde7abd97c1529aa768e1baff # v5
+        with:
+          go-version: ${{ env.GO_VERSION }}
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@55c2c1448f86e01eaae002a5a3a9624417608d84 # v6
+        with:
+          version: v1.64.5
+          args: --timeout=5m
--- a/.github/workflows/octo-ci-status.yml
+++ b/.github/workflows/octo-ci-status.yml
@@ -6,6 +6,8 @@ on:
     types: [completed]
     branches: [main]
 
+permissions: {}
+
 jobs:
   notify:
     uses: Mininglamp-OSS/.github/.github/workflows/octo-ci-status.yml@v1
--- a/.github/workflows/octo-ci-status.yml
+++ b/.github/workflows/octo-ci-status.yml
@@ -6,7 +6,8 @@ on:
     types: [completed]
     branches: [main]
 
-permissions: {}
+permissions:
+  actions: read
 
 jobs:
   notify:
--- a/.github/workflows/octo-ci-status.yml
+++ b/.github/workflows/octo-ci-status.yml
@@ -1,0 +1,21 @@
+name: Octo CI Status
+
+on:
+  workflow_run:
+    workflows: ["CI"]
+    types: [completed]
+    branches: [main]
+
+jobs:
+  notify:
+    uses: Mininglamp-OSS/.github/.github/workflows/octo-ci-status.yml@v1
+    with:
+      repo_name: ${{ github.event.workflow_run.repository.name }}
+      workflow_name: ${{ github.event.workflow_run.name }}
+      conclusion: ${{ github.event.workflow_run.conclusion }}
+      run_id: ${{ github.event.workflow_run.id }}
+      run_url: ${{ github.event.workflow_run.html_url }}
+      project_group_id: "6095b6ded00046009e72882ce4067289"
+      workflow_id: ${{ github.event.workflow_run.workflow_id }}
+    secrets:
+      OCTO_BOT_TOKEN: ${{ secrets.OCTO_BOT_TOKEN }}
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,7 @@
+version: "2"
+run:
+  timeout: 5m
+linters:
+  disable-all: true
+  enable:
+    - govet
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -2,6 +2,6 @@ version: "2"
 run:
   timeout: 5m
 linters:
-  disable-all: true
+  default: none
   enable:
     - govet


Jerry-Xin (changes requested):
Summary: The PR is relevant to `octo-lib`, but the new status notification workflow undermines the stated SHA-pinning security goal while passing a repository secret to an external reusable workflow.

🔴 Blocking

🟡 Warning — `.github/workflows/octo-ci-status.yml:11` calls `Mininglamp-OSS/.github/.github/workflows/octo-ci-status.yml@v1` while passing `OCTO_BOT_TOKEN` at lines 20-21. `@v1` is a mutable tag, so anyone able to move that tag in the shared `.github` repository can change the code that receives this secret. Since this PR explicitly pins actions for supply-chain security, the newly added reusable workflow should also be pinned to an immutable commit SHA.

💬 Non-blocking

🔵 Suggestion — `.github/workflows/ci.yml:17` upgrades only the CI toolchain to Go 1.24, while `go.mod:3` still declares `go 1.20`. That may be intentional to preserve downstream compatibility, but the PR title/description currently reads like the module itself is being upgraded. Consider either updating `go.mod` or clarifying that this is a CI runner upgrade only.

✅ Highlights

The PR is in scope for this repository: it updates CI, linting, and project notification workflows for `octo-lib`.

`actionlint` passes for the changed workflows.

`go test ./...` passes locally in the available Go toolchain.

yujiawei (commented):
# Code Review — PR #12 (octo-lib)

## Summary

Solid CI hygiene PR: Go 1.20 → 1.24 toolchain bump, parallelization of build/test/vet/lint, action SHA pinning for supply-chain safety, and a new state-change notification workflow. Verified mechanics check out. One operational concern that is likely to bite on first merge (lint will almost certainly fail), and a few smaller items worth tightening.

---

## 1. Verification

| Item | Result | Evidence |
| --- | --- | --- |
| `actions/checkout@34e1148...` is the canonical v4 commit | ✅ | `refs/tags/v4` → `34e114876b0b11c390a56381ad16ebd13914f8d5` |
| `actions/setup-go@40f1582...` is the canonical v5 commit | ✅ | `refs/tags/v5` → `40f1582b2485089dde7abd97c1529aa768e1baff` |
| `golangci/golangci-lint-action@55c2c14...` is the canonical v6 commit | ✅ | `refs/tags/v6` → `55c2c1448f86e01eaae002a5a3a9624417608d84` (release `6.5.2`) |
| Reusable workflow `Mininglamp-OSS/.github/.github/workflows/octo-ci-status.yml@v1` exists and input/secret schema matches caller | ✅ | All 7 inputs (`repo_name`, `workflow_name`, `conclusion`, `run_id`, `workflow_id`, `run_url`, `project_group_id`) and `OCTO_BOT_TOKEN` secret align |
| `name: CI` in `ci.yml` matches `workflows: ["CI"]` filter in `octo-ci-status.yml` | ✅ | `ci.yml:1` |
| `golangci-lint v1.64.5` is compatible with `golangci-lint-action@v6` and supports Go 1.24 | ✅ | v1.64.x added Go 1.24 support; v6 of the action covers golangci-lint v1.x |
| Removing `needs: build` from `test` is sound | ✅ | `go test` re-compiles via the test pipeline; `build` is `CGO_ENABLED=0` while `test -race` requires CGO — they validate different surfaces, parallelization is safe |

---

## 2. Findings

### P1 — Lint job will almost certainly fail on first run

**Where:** `.github/workflows/ci.yml:46-65` (new `lint` job)

**Issue:** The repo has no `.golangci.yml` and has never been linted, so this PR introduces lint with **defaults** (`errcheck`, `gosimple`, `govet`, `ineffassign`, `staticcheck`, `unused`) onto a sizeable, pre-existing codebase. With ~20 packages including older code (`go-redis v6`, `olivere/elastic v6+incompatible`, etc.), `staticcheck` and `errcheck` typically surface dozens of legacy issues on first run.

If the new `lint` job becomes a required check before this PR is itself merged, this PR will not pass its own lint job. Even if it merges as a non-required check, the next `main` push will mark the CI red and trigger an immediate `octo-ci-status` notification.

**Suggestions (pick one):**
- Add a minimal `.golangci.yml` in this PR scoping linters to a known-clean set (e.g. `govet`, `ineffassign`) and explicitly disable noisy ones until the codebase is cleaned up.
- Or run `golangci-lint run --timeout=5m` locally first and either fix or `//nolint:` the existing findings before merge.
- Or land the lint job as `continue-on-error: true` for one or two iterations to surface the backlog without gating.

### P2 — `go.mod` still declares `go 1.20`

**Where:** `go.mod:3`

The PR raises CI's `GO_VERSION` to 1.24 but leaves the module directive at `go 1.20`. Go is forward-compatible so builds still work, but:
- Downstream consumers of `octo-lib` won't be nudged to 1.24+ tooling.
- The "upgrade to Go 1.24" intent isn't fully reflected on the package metadata.

Library repos sometimes intentionally keep a low `go` directive to widen consumer compatibility — if that's the case here, it's worth a one-line comment in the PR description so reviewers don't flag it again. Otherwise, bumping to `go 1.22` or `go 1.24` in `go.mod` would align the two.

### P2 — Node.js 20 actions deprecation lands ~2 weeks after this PR

**Where:** `.github/workflows/ci.yml` (all four jobs)

The most recent CI run on `main` already prints:
> Node.js 20 actions are deprecated... Actions will be forced to run with Node.js 24 by default starting June 2nd, 2026.

The pinned SHAs for `actions/checkout@v4` (commit dated 2025-11-13) and `actions/setup-go@v5` (commit dated 2025-12-15) are the latest v4/v5 tags, which still ship the Node.js 20 runtime. After **2026-06-02** GitHub will force these to Node.js 24, which has historically caused breakage for actions not explicitly tested on the new runtime.

This is not a blocker for this PR — the upgrade window is the same regardless — but worth tracking as a follow-up. Long-term, pinning to `actions/checkout@v6` SHA (Node 24 native) would survive the cutover unchanged.

### P2 — Caller for `octo-ci-status.yml` could use `secrets: inherit` and concurrency

**Where:** `.github/workflows/octo-ci-status.yml`

Two small operational improvements:

1. **Secret passing.** The current explicit `secrets: OCTO_BOT_TOKEN: ${{ secrets.OCTO_BOT_TOKEN }}` works but is brittle: if the reusable workflow ever requires an additional secret, every caller must be updated. `secrets: inherit` is the conventional choice for org-internal reusable workflows and matches how most Mininglamp-OSS callers will use this.

2. **Concurrency.** No `concurrency:` block on the caller. If `main` lands two pushes in quick succession and both CI runs complete close together, two notification jobs fire. The reusable workflow's state-change logic should de-dupe most cases, but adding:
   ```yaml
   concurrency:
     group: octo-ci-status-${{ github.event.workflow_run.id }}
     cancel-in-progress: false
   ```
   would harden against races without losing recovery notifications.

### P3 — `cancel-in-progress: true` on push to `main`

**Where:** `.github/workflows/ci.yml:11-13` (pre-existing, not changed by this PR)

Pre-existing observation rather than a finding against this diff: the concurrency group cancels in-progress runs on `main` pushes too. With `octo-ci-status.yml` watching `workflow_run.completed` events, a cancelled run will fire `conclusion: cancelled`, which the reusable workflow may or may not classify as a state transition. Worth confirming the reusable workflow's behavior on `cancelled` since it isn't `success` or `failure`.

---

## 3. Pre-existing observations (not blockers for this PR)

- **`CGO_ENABLED=0 go build -v ./...` and `pkg/db/sqlite.go`.** `pkg/db/sqlite.go:8` blank-imports `github.com/mattn/go-sqlite3`, which is CGO-only. The build job passes today (and will continue to under Go 1.24), but only a subset of packages appears in `-v` output, suggesting `pkg/db` may be silently excluded somehow or simply cached. Worth tracing once-over to confirm `pkg/db` actually compiles in CI rather than being a no-op. Out of scope for this PR.

- **Test job `go test -race ./...` already covers the build surface that matters for correctness; the dedicated `build` job exists primarily to validate the CGO-disabled artifact.** Keeping it parallel rather than serial is the right call.

---

## 4. Verdict

`COMMENTED` — no P0 blockers, mechanics all verify. **Strongly recommend resolving the P1 lint configuration item before merge** (or making `lint` non-blocking for the first one or two iterations); the rest can be addressed as follow-ups.

lml2468 (commented):
Independent review on `6be011d`.

I disagree with the existing `@v1` blocker. For org-owned reusable workflows, `@v1` is already the established caller pattern; `octo-server` and `octo-admin` both call this same `octo-ci-status` reusable workflow at `@v1`. This PR's SHA-pinning scope is third-party actions (`actions/checkout`, `actions/setup-go`, `golangci-lint-action`). A two-tier policy, with immutable SHAs for external actions and a version tag for org-owned reusable workflows, is reasonable. If we want org workflow SHA pinning, that should be a uniform org-wide policy change, not a blocker only for this PR.

Findings:

- **P1: The new Lint job is already failing on PR CI.** Since this PR only changes workflows, the lint findings are pre-existing code issues (`errcheck`, `unused`, `gosimple`, `staticcheck`, including the deprecated jaeger/rest/curve25519 calls). Merging as-is leaves `main` with red CI and makes every future PR inherit the failed Lint check. Please either clean up the baseline here, add `.golangci.yml` / targeted `nolint` suppressions for accepted legacy issues, or explicitly land this as a documented first-lint baseline/temporarily non-gating job.

- **P1: `.github/workflows/octo-ci-status.yml` should declare explicit permissions.** The reusable workflow needs `actions: read` for workflow-run history/state-transition detection and `contents: read` for checkout, but the caller currently has no `permissions:` block, so this depends on repo default token settings. Add the minimum permissions explicitly:
  ```yaml
  permissions:
    actions: read
    contents: read
  ```

- **P2: The pinned action SHAs are still Node.js 20 actions.** The current CI log warns that GitHub will force JavaScript actions to Node.js 24 on **2026-06-02** and remove Node.js 20 from runners on **2026-09-16**. This is not a merge blocker, but these SHAs will need a follow-up bump soon.

- **Non-blocking: `go.mod` still declares `go 1.20` while CI runs Go 1.24.** That may be intentional for library compatibility, but it means Go 1.21+ language/library usage could slip through CI and break downstream consumers using the module's declared version. Either align `go.mod` or document that this is a CI-toolchain-only upgrade.

Verdict: `COMMENTED`.

yujiawei (changes requested):
# Code Review — PR #12 (octo-lib)

## Summary

Re-review on the new head `fa8585b`, which bumped `GO_VERSION` from 1.24 → 1.25 and `golangci-lint` from `v1.64.5` → `v2.12.2` since the previous round.

The bumps introduced a hard regression: **the new lint job is currently failing on this exact head SHA** because `golangci-lint-action@v6` does not support golangci-lint v2.x. This is a real blocker, not theoretical — the failure is already visible in CI. Recommending `CHANGES_REQUESTED` until the action/version mismatch is resolved. Other items below are smaller.

---

## 1. Verification

| Item | Result | Evidence |
| --- | --- | --- |
| `actions/checkout@34e1148...` resolves to canonical v4 | ✅ | `refs/tags/v4` → `34e114876b0b11c390a56381ad16ebd13914f8d5` |
| `actions/setup-go@40f1582...` resolves to canonical v5 | ✅ | `refs/tags/v5` → `40f1582b2485089dde7abd97c1529aa768e1baff` |
| `golangci/golangci-lint-action@55c2c14...` resolves to canonical v6 | ✅ | `refs/tags/v6` → `55c2c1448f86e01eaae002a5a3a9624417608d84` |
| `golangci-lint v2.12.2` compatible with `golangci-lint-action@v6` | ❌ | See P0 below — action prints `golangci-lint v2 is not supported by golangci-lint-action v6, you must update to golangci-lint-action v7` |
| Reusable workflow input/secret schema matches caller | ✅ | All 7 inputs (`repo_name`, `workflow_name`, `conclusion`, `run_id`, `workflow_id`, `run_url`, `project_group_id`) plus `OCTO_BOT_TOKEN` align with `Mininglamp-OSS/.github/.github/workflows/octo-ci-status.yml@v1` |
| `name: CI` in `ci.yml` matches `workflows: ["CI"]` filter in `octo-ci-status.yml` | ✅ | `ci.yml:1` |
| Removing `needs: build` from `test` is sound | ✅ | `go test` recompiles its own surface; `build` is `CGO_ENABLED=0` while `test -race` requires CGO — they cover different artifacts and parallelize safely |
| Go 1.25 toolchain works for this codebase | ✅ | Build/Vet jobs were green on this SHA before lint cancelled the run; `GOVERSION='go1.25.10'` in CI logs |

---

## 2. Findings

### 🔴 P0 — Lint job is broken on this head: `golangci-lint v2` is not supported by `golangci-lint-action@v6`

**Where:** `.github/workflows/ci.yml:62-65`

```yaml
uses: golangci/golangci-lint-action@55c2c1448f86e01eaae002a5a3a9624417608d84 # v6
with:
  version: v2.12.2
  args: --timeout=5m
```

**Evidence (from the live PR run on this head SHA):**

```
##[error]Failed to run: Error: invalid version string 'v2.12.2',
golangci-lint v2 is not supported by golangci-lint-action v6,
you must update to golangci-lint-action v7.
```

Run: <https://github.com/Mininglamp-OSS/octo-lib/actions/runs/25928740551/job/76216907500>

`golangci-lint v2` shipped a breaking config-format change and is intentionally gated behind `golangci-lint-action@v7`. The pinned `@v6` SHA explicitly rejects any `v2.x` version string. This is not a warning — the job exits non-zero on every run, so the new `Lint` check will be permanently red until fixed.

**Fix (pick one):**

1. **Upgrade the action to v7 (matches stated v2.12.2 intent):**
   ```yaml
   uses: golangci/golangci-lint-action@<v7-commit-sha> # v7
   with:
     version: v2.12.2
     args: --timeout=5m
   ```
   Note: golangci-lint v2 also requires a v2-format `.golangci.yml` if any config exists; the repo currently has none, so default ruleset applies — see P1 below.

2. **Pin to golangci-lint v1.x (matches the v6 action):**
   ```yaml
   uses: golangci/golangci-lint-action@55c2c1448f86e01eaae002a5a3a9624417608d84 # v6
   with:
     version: v1.64.5   # or any v1.x
     args: --timeout=5m
   ```

The commit message claims "1.24 is EOL" as justification for bumping to Go 1.25 and lint v2 — note that per Go's release policy the two most recent majors are supported, so Go 1.24 is **not** EOL until Go 1.26 ships. The Go 1.25 bump itself is fine; the lint v2 bump is the part that broke.

---

### 🟡 P1 — Even after the action/version mismatch is fixed, default-ruleset lint will surface a backlog

**Where:** `.github/workflows/ci.yml:53-65`, repo has no `.golangci.yml`

This was raised in the previous round and remains relevant: `octo-lib` has never been linted, contains older dependencies (`go-redis v6`, `olivere/elastic v6+incompatible`, deprecated jaeger client, etc.), and golangci-lint v2's default linters (`errcheck`, `govet`, `ineffassign`, `staticcheck`, `unused`) typically produce dozens of legacy findings on first run.

The author already saw this happen in the previous CI run on `6be011d` (lint failed there too, before the v2 bump masked it with a different error). With v2's stricter `staticcheck` and the newly-bundled `unused`/`ineffassign` defaults, the failure backlog will likely be similar or worse.

**Suggested approach:**
- Add a minimal `.golangci.yml` (v2 format — `version: "2"`, `linters.enable`, `linters.exclusions`) scoped to a known-clean linter set.
- Or land `lint:` with `continue-on-error: true` for one or two iterations to surface the backlog without gating merges.
- Or run `golangci-lint run --timeout=5m` locally on `fa8585b` first and either fix or `//nolint:` legacy findings before merge.

Whichever path is chosen, please verify CI is green before merging — leaving `Lint` red on `main` will trigger `octo-ci-status` notifications on every push.

---

### 🟡 P1 — `.github/workflows/octo-ci-status.yml` has no explicit `permissions:` block

**Where:** `.github/workflows/octo-ci-status.yml:1-21`

The caller workflow is missing a `permissions:` block, so it inherits the repo's default `GITHUB_TOKEN` permissions. The reusable workflow needs `actions: read` to query `workflow_run` history for state-transition detection and `contents: read` for checkout. If the org default permission ever tightens to `contents: read` only (the security-recommended baseline), this caller will silently lose state-change detection.

**Fix:** add an explicit minimum-permissions block at workflow scope:

```yaml
permissions:
  actions: read
  contents: read
```

This was raised in the prior review and has not been addressed in `fa8585b`.

---

### 🟢 P2 — Reusable workflow `Mininglamp-OSS/.github/.github/workflows/octo-ci-status.yml@v1` is a mutable tag

**Where:** `.github/workflows/octo-ci-status.yml:11`

The PR explicitly pins external actions (`actions/checkout`, `actions/setup-go`, `golangci-lint-action`) to immutable commit SHAs for supply-chain safety — but the new caller passes `OCTO_BOT_TOKEN` to a reusable workflow at the mutable `@v1` tag. Anyone with permission to move the `v1` tag in the org-shared `.github` repo can change the code that receives this secret.

The author's previous response (two-tier policy: SHAs for third-party, tags for org-internal) is defensible, and `octo-server`/`octo-admin` already follow that pattern. I'd accept this as a non-blocker for *this* PR, but flagging that the inconsistency does undercut the stated supply-chain goal slightly. Whether to harmonize is an org-wide policy call, not a single-PR call.

---

### 🟢 P2 — `go.mod` still declares `go 1.20` while CI runs Go 1.25

**Where:** `go.mod:3`

The PR raises the CI toolchain by **five minor versions** (1.20 → 1.25) but leaves the module directive at `go 1.20`. With such a wide gap, Go 1.21+ language and stdlib usage (e.g. `min`/`max`/`clear` builtins, `slices`/`maps` stdlib packages, range-over-func in 1.22, generic type aliases in 1.24) can land on `main` without ever failing CI, but will break downstream consumers who use the module's declared toolchain.

If the intent is genuinely "library compatibility, downstream consumers may pin Go 1.20" then leaving `go 1.20` is correct — but please add a one-line comment in the PR description so future reviewers don't re-flag this. Otherwise, bumping `go.mod` to `go 1.22` or `go 1.24` would close the gap.

---

### 🟢 P2 — `octo-ci-status.yml` could use `secrets: inherit` and a `concurrency:` block

**Where:** `.github/workflows/octo-ci-status.yml`

1. **Secret passing.** Explicit `secrets: OCTO_BOT_TOKEN: ${{ secrets.OCTO_BOT_TOKEN }}` works but every future secret added to the reusable workflow requires updating every caller. `secrets: inherit` is the conventional pattern for org-internal reusable workflows.

2. **Concurrency.** Two `main` pushes landing close together both fire `workflow_run.completed`, producing two notifications. The reusable workflow's state-change logic likely de-dupes most cases, but a caller-side guard is cheap insurance:
   ```yaml
   concurrency:
     group: octo-ci-status-${{ github.event.workflow_run.id }}
     cancel-in-progress: false
   ```

---

### 🟢 P2 — Pinned action SHAs are still Node.js 20 actions; deprecation lands ~2 weeks after merge

**Where:** `.github/workflows/ci.yml` (all four jobs)

CI logs on this head SHA print:

> Node.js 20 actions are deprecated... Actions will be forced to run with Node.js 24 by default starting June 2nd, 2026. Node.js 20 will be removed from the runner on September 16th, 2026.

The pinned `actions/checkout@v4` (commit dated 2025-11-13) and `actions/setup-go@v5` (2025-12-15) are still Node 20-based. Not a blocker for this PR — the cutover window is the same regardless — but worth scheduling a follow-up bump to `actions/checkout@v6` (Node 24 native) before `2026-06-02`.

---

### 🟢 P3 — `cancel-in-progress: true` interacts with `workflow_run` notifications

**Where:** `.github/workflows/ci.yml:11-14` (pre-existing, unchanged by this PR)

Concurrency cancels in-progress runs on `main` pushes too. A cancelled run fires `conclusion: cancelled`, which the reusable workflow may or may not classify as a state transition — worth confirming in the reusable workflow's logic since neither `success` nor `failure` covers it. Out of scope for this PR.

---

## 3. Highlights

- All third-party action SHAs verified against canonical v4/v5/v6 tags.
- Reusable workflow input/secret schema fully matches the caller.
- Parallelization (removing `needs: build` from `test`) is the right call — `build` and `test` validate different artifacts (CGO-disabled vs `-race` with CGO).
- `name: CI` ↔ `workflows: ["CI"]` correctly aligned for `workflow_run` triggering.
- Action SHA pinning meaningfully reduces supply-chain risk on third-party actions.

---

## 4. Verdict

`CHANGES_REQUESTED` — the lint job is currently failing on this exact head SHA due to the `golangci-lint-action@v6` ↔ `golangci-lint v2.12.2` incompatibility (P0). Once that is fixed and the resulting lint output is either clean or scoped via `.golangci.yml`/`continue-on-error`, this is ready to land. The remaining items (P1 permissions block, P2 `go.mod` gap, etc.) can be addressed in this PR or as fast-follows depending on how aggressive you want to be.

Jerry-Xin (changes requested):
This PR is in scope for `octo-lib`, but the Go upgrade is incomplete.

🔴 Blocking

🟡 Warning — `.github/workflows/ci.yml:17` upgrades CI to Go `1.25`, but `go.mod:3` still declares `go 1.20`. If the intent is to move this library off EOL Go, the module directive must be updated as part of the same PR. Otherwise consumers and Go tooling still treat the module as a Go 1.20 module while CI no longer validates that version. Please either update `go.mod` to the supported target version or explicitly state that this PR only changes the CI test toolchain while preserving Go 1.20 compatibility.

💬 Non-blocking

🔵 Suggestion — `.github/workflows/octo-ci-status.yml:11` calls a reusable workflow via mutable `@v1` while this PR’s security theme is pinning workflow dependencies. Since this job passes `OCTO_BOT_TOKEN`, consider pinning the reusable workflow to a full commit SHA too, or document why internal org workflow tags are the accepted exception.

🔵 Suggestion — `.github/workflows/octo-ci-status.yml` could set `permissions: {}` unless the reusable workflow requires a scoped `GITHUB_TOKEN`. That would match the least-privilege style already used by some existing workflows.

✅ Highlights

The PR is project-relevant and limited to CI/workflow infrastructure. The pinned SHAs for `actions/checkout@v4`, `actions/setup-go@v5`, and `golangci/golangci-lint-action@v6` match their major tags. Local verification passed with `go test ./...` and `go test -race -count=1 -timeout 5m ./...`.

lml2468 (commented):
Incremental re-review on `fa8585b`.

Verified: the only incremental code diff since `6be011d` is `GO_VERSION: 1.24 -> 1.25` and `golangci-lint: v1.64.5 -> v2.12.2` in `.github/workflows/ci.yml`.

Findings:
- **P0:** `golangci-lint-action@55c2c144...` is still v6, but the workflow now requests `version: v2.12.2`. The live Lint job fails with: `golangci-lint v2 is not supported by golangci-lint-action v6, you must update to golangci-lint-action v7`. Fix by pinning the action to a v7 SHA, or downgrade golangci-lint back to a v1.x release.
- **P1:** `go.mod` still declares `go 1.20` while CI now runs Go `1.25`. The gap widened and Allen's blocker is valid unless the PR explicitly documents this as CI-toolchain-only while preserving Go 1.20 module compatibility.
- **P1:** `.github/workflows/octo-ci-status.yml` still has no explicit `permissions:` block. Add the minimum required `actions: read` and `contents: read` instead of depending on repo defaults.
- **P2:** The Node.js 20 actions deprecation note still applies as a follow-up.

The previous `@v1` reusable-workflow concern is resolved/non-blocking for this PR.

Verdict: `COMMENTED`; existing `CHANGES_REQUESTED` reviews should stand until the CI breakage and version/permissions gaps are fixed.

Jerry-Xin (changes requested):
One-line summary: The PR is relevant to `octo-lib`, but the newly added lint job currently fails on the existing codebase and would make CI red immediately.

🔴 Blocking

- 🟡 Warning: [.github/workflows/ci.yml](/private/tmp/pr-review-rl6VTR/repo/.github/workflows/ci.yml:61) adds `golangci-lint` with `version: v2.12.2` and no repository config/baseline, but running the configured command fails with 37 findings. Examples include unchecked writes in [pkg/markdown/markdown.go](/private/tmp/pr-review-rl6VTR/repo/pkg/markdown/markdown.go:32), unchecked `resp.Body.Close()` in [pkg/network/network.go](/private/tmp/pr-review-rl6VTR/repo/pkg/network/network.go:120), deprecated API usage in [config/tracer.go](/private/tmp/pr-review-rl6VTR/repo/config/tracer.go:53), and unused symbols in [pkg/log/logger.go](/private/tmp/pr-review-rl6VTR/repo/pkg/log/logger.go:17). This will block every PR and main push once merged. Fix the reported issues or add an intentional `.golangci.yml` baseline/configuration before enabling the job as required CI.

💬 Non-blocking

- 🔵 Suggestion: [.github/workflows/octo-ci-status.yml](/private/tmp/pr-review-rl6VTR/repo/.github/workflows/octo-ci-status.yml:13) uses the reusable workflow at `@v1`. That matches existing repo workflow style, but if the goal is strict supply-chain pinning, consider whether reusable workflows should also be pinned or governed by a documented org policy.

✅ Highlights

- The PR is in scope for this repository because it modifies CI and repository workflow behavior.
- The pinned SHAs for `actions/checkout@v4`, `actions/setup-go@v5`, and `golangci/golangci-lint-action@v9.2.0` resolve to the matching tags.
- `go test ./...` and `go vet ./...` pass locally; the blocker is specifically the new lint job.

lml2468 (commented):
## Incremental re-review on `9cac5f2`

I checked the full PR diff, read the changed files from the head branch ref `fix/workflow-improvements`, verified the existing reviews, and checked the reusable workflow at `Mininglamp-OSS/.github@v1`.

Note: GitHub does not allow this authenticated account to request changes on its own PR, so this is submitted as `COMMENTED`. The review verdict is still **CHANGES_REQUESTED**.

## Previous findings

- **Resolved:** The previous P0 action/version mismatch is fixed. `golangci/golangci-lint-action@1e7e51e771db61008b38414a730f564565cf7c20` resolves exactly to `refs/tags/v9.2.0`, and this version supports running `golangci-lint v2.12.2`. The current failure is now the real lint output, not the old v6/v2 incompatibility.
- **Still unresolved:** `go.mod:3` still declares `go 1.20` while CI now runs `GO_VERSION: '1.25.x'`. If this library intentionally keeps Go 1.20 compatibility, the PR should say so explicitly and CI should still have a way to validate that compatibility. Otherwise, update the module directive to the supported target. As written, newer standard-library usage can compile in CI and still break consumers using the module's declared Go version.
- **Still blocking:** `.github/workflows/octo-ci-status.yml` now has `permissions: {}`, which is not sufficient. The called reusable workflow declares `actions: read` and `contents: read`, checks out `Mininglamp-OSS/.github`, and its script calls the Actions API for the current run and workflow-run history. Reusable workflow token permissions can only be maintained or reduced through the call chain, not elevated by the called workflow, so an empty caller permission set will make the notification workflow fail on `main` workflow-run completions. Please set the caller permissions explicitly:

```yaml
permissions:
  actions: read
  contents: read
```

## Additional current-head blocker

- **P1:** The `Lint` check is still failing on the current head. I agree this is pre-existing code debt exposed by adding lint for the first time, but merging this workflow as a required failing job leaves `main` permanently red. Please either fix/baseline the current findings with an intentional `.golangci.yml`, or make the first lint rollout non-blocking until the baseline is cleaned up.

## No new issues found

- `actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5` matches `v4`.
- `actions/setup-go@40f1582b2485089dde7abd97c1529aa768e1baff` matches `v5`.
- `GO_VERSION: '1.25.x'` itself is not causing the current Build/Test/Vet failures; those checks are green.

## Verdict

**CHANGES_REQUESTED** until the token permissions and red lint rollout are fixed, and the Go module/CI version gap is either aligned or explicitly documented with compatibility validation.

Jerry-Xin (changes requested):
Project relevance gate passed: this PR is in scope because it updates CI, linting, and repository workflow automation for `octo-lib`. However, the new lint job is currently broken.

🔴 Blocking

- 🟡 Warning: `.golangci.yml:5` uses `linters.disable-all`, which is invalid for `golangci-lint` v2 config. I verified with `golangci-lint v2.12.2 config verify -c .golangci.yml`, and it fails with: `additional properties 'disable-all' not allowed`. This means the new `.github/workflows/ci.yml:61` lint job will fail before linting. Use the v2 syntax, for example `linters.default: none` with `linters.enable`.

💬 Non-blocking

- 🔵 Suggestion: `.github/workflows/ci.yml:17` sets CI to Go `1.25.x`, but `go.mod:3` still declares `go 1.20`. If the project is actually upgrading its supported Go baseline, update `go.mod` too. If only the CI runner is changing, clarify that in the PR description.
- 🔵 Suggestion: `.github/workflows/ci.yml:61` adds golangci-lint but enables only `govet`, while `.github/workflows/ci.yml:50` already runs `go vet ./...`. After fixing the config, consider enabling additional linters or dropping the duplicate vet-only lint job.
- 🔵 Suggestion: `.github/workflows/octo-ci-status.yml:14` calls a reusable workflow via mutable `@v1` while this PR’s stated goal includes pinning action SHAs. Since this workflow receives `OCTO_BOT_TOKEN`, consider pinning the reusable workflow to an immutable commit SHA as well.

✅ Highlights

- The PR correctly keeps build, test, vet, and lint jobs parallel.
- The `actions/checkout`, `actions/setup-go`, and `golangci/golangci-lint-action` SHAs match the tagged versions noted in comments.

lml2468 (commented):
Verdict: CHANGES_REQUESTED

Incremental re-review for `bf9894608409af954eec919f87a2c8ec8ddcadae` against previous `9cac5f2be53402b70bbe5575947985061fd7aee6`.

Blocking

- P1 Lint remains unresolved. I agree with Jerry-Xin latest review rather than filing a duplicate finding: `.golangci.yml` declares `version: "2"` but still uses `linters.disable-all`. The current GitHub `Lint` check fails in `golangci-lint config verify` with `additional properties disable-all not allowed`. I also verified locally with `golangci-lint 2.12.2`; replacing it with `linters.default: none` validates successfully.

Previous findings status

- P0 notification permissions: resolved. The caller now grants `actions: read`; the reusable workflow needs `actions:read` for the Actions API. `Mininglamp-OSS/.github` visibility is public, and its reusable workflow checkout targets that public repo, so I do not see a remaining `contents:read` blocker here.
- P1 go.mod: unresolved. `go.mod` still says `go 1.20` while CI uses `GO_VERSION: 1.25.x`. Jerry-Xin already called this out in the latest review as a suggestion, so I am only confirming the state here.
- P1 lint: unresolved as above; current CI `Lint` is still red.

New issues

- No new issues found in the incremental changes beyond the already-reviewed golangci-lint v2 config failure.

Jerry-Xin (changes requested):
Summary: Project relevance gate passed; these CI changes are in scope for `octo-lib`, but two issues need author action before merge.

🔴 Blocking

- 🟡 Warning: `.github/workflows/octo-ci-status.yml:14` calls `Mininglamp-OSS/.github/.github/workflows/octo-ci-status.yml@v1` by mutable tag while passing `OCTO_BOT_TOKEN` at `.github/workflows/octo-ci-status.yml:23-24`. This weakens the PR’s supply-chain hardening because future movement of `v1` can change secret-consuming workflow code without a change in this repo. Pin the reusable workflow to an immutable commit SHA, or document why this repo intentionally trusts the mutable org-level tag.

- 🟡 Warning: `.github/workflows/ci.yml:17` sets `GO_VERSION: '1.25.x'`, while the PR description says Go 1.24 and `go.mod:3` still declares `go 1.20`. For a shared library, this leaves the module advertising Go 1.20 compatibility while CI no longer tests that compatibility. Either update `go.mod` to the intended minimum Go version and align the PR text, or keep a CI matrix that includes the supported minimum Go version.

💬 Non-blocking

- 🔵 Suggestion: `.golangci.yml:5-7` enables only `govet`, which duplicates the existing `go vet ./...` job. If the goal is a stronger code-quality gate matching peer repos, consider enabling the same linter set used there or removing the separate `vet` job once golangci-lint owns that check.

✅ Highlights

- The pinned SHAs for `actions/checkout`, `actions/setup-go`, and `golangci/golangci-lint-action` resolve to the tagged versions noted in comments.
- Removing `needs: build` is reasonable because `go test` rebuilds packages as needed.
- Local verification passed: `go test -race -count=1 -timeout 5m ./...`, `go vet ./...`, and `golangci-lint v2.12.2 run --timeout=5m`.

lml2468 (commented):
Verdict: **APPROVED**

Incremental re-review on `468b2c6` against previous `bf98946`.

## Previous findings

All prior blocking issues are resolved:

- **golangci-lint-action v6/v2.x incompatibility** (P0, round 2): Fixed — now `v9.2.0`, SHA verified.
- **`permissions: {}`** (P0, round 2): Fixed — caller now grants `actions: read`; `.github` repo is public so `contents: read` not needed for checkout.
- **`.golangci.yml` v1 syntax** (P1, round 3): Fixed — `disable-all: true` replaced with `default: none`. Verified locally with `golangci-lint v2.12.2 config verify` — exit 0.
- **CI Lint red** (P1, rounds 2–3): Fixed — all four checks (Build, Test, Vet, Lint) are now green.

## Remaining items (non-blocking)

- **P2 `go.mod` declares `go 1.20`, CI uses `1.25.x`.** This is pre-existing technical debt, not introduced by this PR. The module directive was `go 1.20` on `main` before this PR. Updating it would change the module's minimum Go version for all consumers and is better handled as a separate dedicated PR with its own testing. I recommend a fast-follow.
- **P2 `@v1` mutable tag for reusable workflow.** Allen flagged this as blocking. My independent assessment: this is an org-internal workflow in `Mininglamp-OSS/.github` (public repo, org-controlled). Only org members with write access can move the tag. The PR's supply-chain hardening goal — pinning *third-party* actions to immutable SHAs — is fully achieved. Org-internal workflows operate within a different trust boundary where mutable version tags are the standard pattern and pinning to SHA would make routine maintenance harder. Acceptable trade-off; can be revisited as an org-wide policy decision.
- **P2 `govet`-only lint duplicates `go vet` job.** Allen noted this as non-blocking. Agree — the lint job currently only runs `govet`, which the existing Vet job already covers. This is a reasonable starting point; additional linters can be enabled incrementally.

## Why APPROVED

This PR has been through 4 review rounds. Every code-quality blocker has been fixed. CI is fully green. The PR achieves its stated goals: action SHA pinning, lint integration, CI status notifications, and Go version upgrade in CI. The remaining items are pre-existing debt or architectural preferences that don't block merge.

lml2468:
Closing: workflow/action PR cleanup per org policy.

PR #16, opened 2026-05-19
feat(ci): refine CI + add stale and PR labeler

lml2468:
## Changes

- **ci.yml**: draft PRs are now skipped; docs/markdown-only changes no longer trigger full CI
- **stale.yml**: four-lane stale management (unassigned/assigned × issues/PRs) via shared reusable workflow
- **labeler.yml**: auto size labels (XS/S/M/L/XL) + dependency-changed detection + maintainer checklist comment

## Dependencies

Requires `Mininglamp-OSS/.github#feat/automation-p0-shared-workflows` (PR #25) to be merged first — the reusable workflows `reusable-stale.yml` and `reusable-pr-labeler.yml` must be on `main` for stale and labeler to be functional.

## Part of P0 automation rollout

Modeled after openclaw best practices.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,6 +59,7 @@ jobs:
           filters: |
             code:
               - '**/*.go'
+              - '**/*.proto'
               - 'go.mod'
               - 'go.sum'
               - 'Makefile'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,9 +37,11 @@ jobs:
       contents: read
       pull-requests: read
     outputs:
-      # For push events the paths-filter step is skipped; default to 'true'
-      # so downstream jobs always run on direct main pushes.
-      code: ${{ github.event_name != 'pull_request' && 'true' || steps.filter.outputs.code }}
+      # For push events the paths-filter step is skipped; default to 'false'
+      # (code=false is not used for push — push always runs). We use a forward
+      # positive pattern list so the logic is unambiguous: code=true when at
+      # least one changed file matches a known source pattern.
+      code: ${{ github.event_name == 'pull_request' && steps.filter.outputs.code || 'true' }}
     steps:
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
         if: github.event_name == 'pull_request'
@@ -49,17 +51,18 @@ jobs:
         id: filter
         if: github.event_name == 'pull_request'
         with:
-          # Negation patterns (! prefix) correctly exclude matched files even
-          # when '**' would otherwise match them — a .md file matches '**' but
-          # is then eliminated by '!**/*.md', leaving code: false for docs-only
-          # PRs. This is standard minimatch behaviour; predicate-quantifier
-          # defaults to 'any' which is exactly what we want.
+          # Positive patterns only — no negation, no predicate-quantifier needed.
+          # code=true when at least one changed file matches a source pattern.
+          # Docs-only PRs (only *.md / docs/) won't match any pattern → code=false
+          # and downstream jobs are skipped (skipped counts as passing for required
+          # status checks).
           filters: |
             code:
-              - '**'
-              - '!**/*.md'
-              - '!docs/**'
-              - '!.github/ISSUE_TEMPLATE/**'
+              - '**/*.go'
+              - 'go.mod'
+              - 'go.sum'
+              - 'Makefile'
+              - '.github/workflows/**'
 
   build:
     needs: [changes]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,9 +2,18 @@ name: CI
 
 on:
   pull_request:
+    types: [opened, reopened, synchronize, ready_for_review, converted_to_draft]
     branches: [main]
+    paths-ignore:
+      - "**/*.md"
+      - "docs/**"
+      - ".github/ISSUE_TEMPLATE/**"
   push:
     branches: [main]
+    paths-ignore:
+      - "**/*.md"
+      - "docs/**"
+      - ".github/ISSUE_TEMPLATE/**"
 
 permissions:
   contents: read
@@ -18,6 +27,7 @@ env:
 
 jobs:
   build:
+    if: github.event_name != 'pull_request' || !github.event.pull_request.draft
     name: Build
     runs-on: ubuntu-latest
     steps:
@@ -29,6 +39,7 @@ jobs:
         run: CGO_ENABLED=0 go build -v ./...
 
   test:
+    if: github.event_name != 'pull_request' || !github.event.pull_request.draft
     name: Test
     runs-on: ubuntu-latest
     needs: build
@@ -41,6 +52,7 @@ jobs:
         run: go test -race -count=1 -timeout 5m ./...
 
   vet:
+    if: github.event_name != 'pull_request' || !github.event.pull_request.draft
     name: Vet
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,11 +3,15 @@ name: CI
 on:
   pull_request:
     types: [opened, reopened, synchronize, ready_for_review, converted_to_draft]
+    # converted_to_draft is kept intentionally: it creates "skipped" check entries
+    # that satisfy required-check branch-protection rules when a PR is demoted
+    # back to draft, preventing stale "pending" statuses from the last real push.
     branches: [main]
-    paths-ignore:
-      - "**/*.md"
-      - "docs/**"
-      - ".github/ISSUE_TEMPLATE/**"
+    # paths-ignore is intentionally NOT set here — workflow-level path filtering
+    # prevents GitHub from creating any check run, which deadlocks required-status
+    # checks on docs-only PRs. Path-based skipping is handled by the `changes`
+    # preflight job below (dorny/paths-filter produces 'skipped' statuses that
+    # GitHub counts as passing for required checks).
   push:
     branches: [main]
     paths-ignore:
@@ -26,8 +30,42 @@ env:
   GO_VERSION: '1.20'
 
 jobs:
+  changes:
+    name: Detect changed paths
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: read
+    outputs:
+      # For push events the paths-filter step is skipped; default to 'true'
+      # so downstream jobs always run on direct main pushes.
+      code: ${{ github.event_name != 'pull_request' && 'true' || steps.filter.outputs.code }}
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+        if: github.event_name == 'pull_request'
+        with:
+          persist-credentials: false
+      - uses: dorny/paths-filter@6852f92c20ea7fd3b0c25de3b5112db3a98da050 # v3
+        id: filter
+        if: github.event_name == 'pull_request'
+        with:
+          # Negation patterns (! prefix) correctly exclude matched files even
+          # when '**' would otherwise match them — a .md file matches '**' but
+          # is then eliminated by '!**/*.md', leaving code: false for docs-only
+          # PRs. This is standard minimatch behaviour; predicate-quantifier
+          # defaults to 'any' which is exactly what we want.
+          filters: |
+            code:
+              - '**'
+              - '!**/*.md'
+              - '!docs/**'
+              - '!.github/ISSUE_TEMPLATE/**'
+
   build:
-    if: github.event_name != 'pull_request' || !github.event.pull_request.draft
+    needs: [changes]
+    if: |
+      (github.event_name != 'pull_request' || !github.event.pull_request.draft) &&
+      needs.changes.outputs.code == 'true'
     name: Build
     runs-on: ubuntu-latest
     steps:
@@ -39,10 +77,12 @@ jobs:
         run: CGO_ENABLED=0 go build -v ./...
 
   test:
-    if: github.event_name != 'pull_request' || !github.event.pull_request.draft
+    if: |
+      (github.event_name != 'pull_request' || !github.event.pull_request.draft) &&
+      needs.changes.outputs.code == 'true'
     name: Test
     runs-on: ubuntu-latest
-    needs: build
+    needs: [changes]
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-go@v5
@@ -52,7 +92,10 @@ jobs:
         run: go test -race -count=1 -timeout 5m ./...
 
   vet:
-    if: github.event_name != 'pull_request' || !github.event.pull_request.draft
+    needs: [changes]
+    if: |
+      (github.event_name != 'pull_request' || !github.event.pull_request.draft) &&
+      needs.changes.outputs.code == 'true'
     name: Vet
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -1,0 +1,19 @@
+name: PR Labeler
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+
+permissions: {}
+
+jobs:
+  label:
+    uses: Mininglamp-OSS/.github/.github/workflows/reusable-pr-labeler.yml@main
+    with:
+      pr_number: ${{ github.event.pull_request.number }}
+      repo_owner: ${{ github.repository_owner }}
+      repo_name: ${{ github.event.repository.name }}
+    permissions:
+      issues: write
+      pull-requests: write
+    secrets: inherit
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -1,7 +1,15 @@
 name: PR Labeler
 
 on:
-  pull_request:
+  # pull_request_target runs in the context of the base repo and retains write
+  # access even for fork PRs — required because the reusable labeler calls the
+  # GitHub REST API to set labels and post comments (both need write tokens).
+  #
+  # Security: this workflow does NOT check out or execute any code from the PR
+  # branch; it only passes the PR number / owner / repo name to a reusable
+  # workflow that performs only API calls against the base repo. This matches
+  # the pattern used by auto-add-to-project.yml and octo-pr-feed.yml.
+  pull_request_target: # zizmor: ignore[dangerous-triggers] metadata-only automation; no PR code executed
     types: [opened, synchronize, reopened]
 
 permissions: {}
@@ -16,4 +24,6 @@ jobs:
     permissions:
       issues: write
       pull-requests: write
-    secrets: inherit
+    # secrets: intentionally omitted — reusable-pr-labeler.yml uses only the
+    # inherited GITHUB_TOKEN granted via the job-level permissions above.
+
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -26,4 +26,3 @@ jobs:
       pull-requests: write
     # secrets: intentionally omitted — reusable-pr-labeler.yml uses only the
     # inherited GITHUB_TOKEN granted via the job-level permissions above.
-
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,15 @@
+name: Stale
+
+on:
+  schedule:
+    - cron: "0 3 * * *"
+  workflow_dispatch:
+
+permissions: {}
+
+jobs:
+  stale:
+    uses: Mininglamp-OSS/.github/.github/workflows/reusable-stale.yml@main
+    permissions:
+      issues: write
+      pull-requests: write


yujiawei (dismissed):
# Code Review — PR #16 (octo-lib)

## Summary

Three focused CI changes: skip draft PRs and docs-only changes in the existing CI workflow, plus two thin caller workflows that delegate to org-shared reusable workflows for stale management and PR labeling. The changes are small (46 additions, 0 deletions, 3 files), well-scoped, and follow GitHub Actions best practices. No blockers found.

## Verification

| Check | Result | Notes |
|---|---|---|
| Draft skip pattern correctness | ✅ | `if: github.event_name != 'pull_request' \|\| !github.event.pull_request.draft` correctly skips on draft PRs and lets push events through. GitHub treats jobs skipped via `if:` as success for required-check evaluation, so this does not accidentally block merges. |
| `pull_request.types` covers draft transitions | ✅ | `ready_for_review` and `converted_to_draft` are included, so flipping draft status correctly re-triggers / re-evaluates CI. |
| `paths-ignore` does not break branch protection | ✅ | Verified `octo-lib/main` ruleset: only `code-review` is a required status check. `Build` / `Test` / `Vet` are advisory, so a docs-only PR that never fires CI cannot deadlock the merge gate. |
| Top-level `permissions: {}` (least privilege) | ✅ | Both new workflows default to no permissions; required scopes (`issues: write`, `pull-requests: write`) are granted only at the job level. |
| Reusable-workflow inputs match callee signature | ✅ | `reusable-pr-labeler.yml` declares `pr_number` (number), `repo_owner` (string), `repo_name` (string) as required — caller passes all three with correct types. `reusable-stale.yml` has no required inputs; defaults apply. |
| Concurrency for CI | ✅ | Existing `concurrency` block (workflow + PR number || ref) is unchanged and still correct. |
| Cron schedule | ✅ | `0 3 * * *` UTC = 11:00 Beijing — reasonable off-peak window. `workflow_dispatch` added for manual runs. |
| Dependency on `.github#25` | ⚠️ | Documented in PR body. Until that PR lands, `stale.yml` and `labeler.yml` runs will fail with "workflow file not found." Acceptable as long as #25 merges first; worth re-verifying in the merge sequence. |

## Findings

### P0 / P1
None.

### P2 / Suggestions (non-blocking)

1. **`labeler.yml` `secrets: inherit` is unnecessary.** The called workflow (`reusable-pr-labeler.yml`) does not declare any `secrets:` inputs and uses only the default `GITHUB_TOKEN` via `actions/github-script`. `secrets: inherit` is a no-op here and slightly inconsistent with `stale.yml`, which omits it. Removing it tightens the surface and improves consistency:

   ```yaml
   # .github/workflows/labeler.yml
   permissions:
     issues: write
     pull-requests: write
   # secrets: inherit   ← can be removed
   ```

2. **Reusable workflows pinned to `@main` rather than to a tag/SHA.** For an org-internal shared workflow this is conventional, but it does mean any future change to `Mininglamp-OSS/.github`'s `main` immediately runs across every consumer with `issues: write` + `pull-requests: write` on whatever PR fires the trigger. Once `.github#25` lands and the shared-workflow surface stabilizes, consider cutting a tag (e.g. `v1`) and pinning to `@v1` (or a commit SHA) across all consumers. Not required to land this PR.

3. **`converted_to_draft` will produce empty workflow runs.** When a ready PR is moved back to draft, the CI workflow triggers and every job evaluates `if:` to false → an Actions run with three skipped jobs appears in the PR check list. Functionally fine (treated as success), but cosmetically noisy. If that bothers you, the simpler form `types: [opened, reopened, synchronize, ready_for_review]` (drop `converted_to_draft`) achieves the same skip behavior without the empty run, because the prior CI status remains attached to the head SHA. Minor preference — keeping it is also defensible.

4. **`paths-ignore` on `synchronize`.** Worth being aware of: a PR whose final commit is markdown-only will not re-run CI on that synchronize event; the previous CI status on an earlier SHA is what GitHub displays for that older commit, while the new head SHA shows no CI run. Because `code-review` is the only required check this does not block merges, but reviewers shouldn't read "no CI on head" as "CI is broken." No action required.

## Additional Observations

- The existing `concurrency` group key `${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}` continues to do the right thing under the new event types.
- `permissions: {}` at the top of `stale.yml` and `labeler.yml` correctly forces explicit per-job grants — good pattern to keep as the template for future reusable-workflow callers in this repo.
- `repo_owner` and `repo_name` could equivalently be derived inside the reusable workflow from `github.repository`, but passing them explicitly from the caller is clearer and lets the reusable workflow stay context-free. No change requested.

## Verdict

Approving. The PR is a clean, defensible CI hardening + automation rollout. The four P2 items above are stylistic and can be addressed in a follow-up (or ignored). Please confirm `Mininglamp-OSS/.github#25` is merged before (or together with) this PR so the reusable-workflow references resolve on `main`.

Jerry-Xin (changes requested):
This PR is relevant to `octo-lib`, but one security issue should be fixed before merge.

🔴 Blocking

- 🔴 Critical — [`.github/workflows/labeler.yml`](/private/tmp/pr-review-YO9tgP/repo/.github/workflows/labeler.yml:19): `secrets: inherit` exposes all available repository/org secrets to the reusable PR labeler workflow, but the called workflow only needs `GITHUB_TOKEN` permissions to label/comment. This is unnecessary secret exposure, especially because the workflow is referenced via mutable `@main`. Remove `secrets: inherit` unless a specific secret is required, and pass only that named secret explicitly.

💬 Non-blocking

- 🟡 Warning — [`.github/workflows/labeler.yml`](/private/tmp/pr-review-YO9tgP/repo/.github/workflows/labeler.yml:4): using `pull_request` means label/comment writes will generally not work for forked PRs because the token is read-only. If OSS fork PR labeling is required, consider `pull_request_target` with a pinned trusted reusable workflow and no checkout of untrusted head code.

- 🟡 Warning — [`.github/workflows/labeler.yml`](/private/tmp/pr-review-YO9tgP/repo/.github/workflows/labeler.yml:11) and [`.github/workflows/stale.yml`](/private/tmp/pr-review-YO9tgP/repo/.github/workflows/stale.yml:12): the reusable workflows are not currently on `Mininglamp-OSS/.github@main`; this PR must not be merged before that dependency lands, or the new workflows will fail to resolve.

✅ Highlights

- `ci.yml` correctly skips draft PR jobs and re-runs on `ready_for_review`.
- `paths-ignore` is scoped to markdown/docs/issue templates, which matches the stated CI optimization.
- Workflow syntax validates with `actionlint`.

lml2468 (commented):
This PR adds three valuable CI improvements (draft skip, docs-only skip, stale + labeler automation). One blocking issue found independently, plus additional findings beyond existing reviews.

**Verdict: CHANGES_REQUESTED** (submitted as COMMENT due to self-approve limitation)

🔴 Blocking issues

1. **`secrets: inherit` in `labeler.yml` — unnecessary secret exposure (independently confirmed)**
   `.github/workflows/labeler.yml:19` passes `secrets: inherit`, but the called `reusable-pr-labeler.yml` declares zero `secrets:` inputs and operates entirely via the default `GITHUB_TOKEN` through `actions/github-script`. This exposes all repository/org secrets (including `OCTO_BOT_TOKEN`) to a reusable workflow referenced via mutable `@main`. Remove `secrets: inherit` entirely — the workflow will still function correctly since it only needs the job-level `permissions` grants.

   This is independently verified by reading the reusable workflow source: no `secrets:` block in `workflow_call`, no `${{ secrets.* }}` usage anywhere in the script.

🟡 Non-blocking suggestions

1. **Fork PR labeling will silently fail**
   `labeler.yml` uses `pull_request` trigger. For a public repo, fork PRs get a read-only `GITHUB_TOKEN`, so `issues.addLabels()` and `issues.createComment()` calls will 403. The workflow will show as failed on fork PRs. Two options:
   - Accept this (fork PRs just won't get labels — functional but noisy)
   - Switch to `pull_request_target` with the current pinned-SHA reusable workflow (safe since no untrusted code checkout occurs)

2. **Mutable `@main` ref for labeler and stale reusable workflows**
   Both `labeler.yml` and `stale.yml` pin to `@main` while the other workflows in this repo (PR feed, issue feed, CI status) pin to `@v1`. Inconsistent pinning strategy. Once the `.github` repo stabilizes these workflows, consider tagging and aligning to `@v1` for consistency and blast-radius control.

3. **`converted_to_draft` produces empty workflow runs**
   Including `converted_to_draft` in `pull_request.types` means moving a ready PR back to draft triggers CI with all three jobs skipped (evaluating `if:` to false). Cosmetically noisy — an Actions run with 3 skipped jobs appears. Dropping `converted_to_draft` from the types list achieves the same skip behavior without the empty run, since the prior CI status remains attached to the head SHA.

4. **Dependency ordering: `.github` repo PR must merge first**
   Both `labeler.yml` and `stale.yml` reference reusable workflows that may not exist on `Mininglamp-OSS/.github@main` yet. Confirm the dependency PR (`.github#25` per PR description) is merged before this one, or these workflows will fail with "workflow file not found."

✅ Highlights
- Draft skip logic is correct and handles push events transparently
- `ready_for_review` + `converted_to_draft` types ensure draft transitions re-evaluate CI
- Top-level `permissions: {}` on both new workflows with per-job grants is the right pattern
- `paths-ignore` scope (markdown, docs, issue templates) is reasonable
- Stale cron at `0 3 * * *` UTC (11:00 Beijing) is a sensible off-peak window

yujiawei (dismissed):
# Code Review — PR #16 (octo-lib)

Scope: three workflow files — `ci.yml` (refinements), `stale.yml` (new), `labeler.yml` (new). Reviewed against current `main`, with cross-checks against the reusable workflows in `Mininglamp-OSS/.github` (`reusable-stale.yml`, `reusable-pr-labeler.yml`), the repo's active branch ruleset, and the live PR check state.

## 1. Verification

| Item | Status | Evidence |
|---|---|---|
| Dependency on `Mininglamp-OSS/.github#25` | ✅ | PR #25 is merged; `reusable-stale.yml` and `reusable-pr-labeler.yml` are present on `main`. |
| Caller↔callee input contract for labeler | ✅ | `labeler.yml:20-23` passes `pr_number` / `repo_owner` / `repo_name`; matches `reusable-pr-labeler.yml` `workflow_call.inputs`. |
| Stale defaults are accepted by callee | ✅ | `stale.yml` calls reusable with no inputs; all four `days_*` inputs in `reusable-stale.yml` have defaults. |
| `pull_request_target` does not execute PR code | ✅ | `labeler.yml` only forwards metadata to the reusable workflow, which runs `actions/github-script` (REST API only) — no `actions/checkout` of PR ref anywhere in the chain. |
| `GITHUB_TOKEN` is sufficient without `secrets: inherit` | ✅ | The reusable workflow does not reference `secrets.*`; the inherited token plus caller-granted `issues: write` / `pull-requests: write` is all it needs. |
| Pinned third-party actions | ✅ | `actions/stale@5bef64f…` and `actions/github-script@f28e40c7…` are pinned by commit SHA in the reusable workflows. First-party `actions/checkout@v4` / `actions/setup-go@v5` already in `ci.yml`. |
| Draft handling does not deadlock required checks | ✅ | `types: [..., ready_for_review, converted_to_draft]` re-triggers the workflow on draft transitions; jobs run on ready, are skipped on draft. |
| Branch protection impact of `paths-ignore` | ⚠️ | Today only `code-review` is a required status check (per repo ruleset `16278090`). Build/Test/Vet are not required, so docs-only PRs are not blocked. See finding below — this is a future-proofing concern if any of them become required. |
| CI currently green on the PR head | ✅ | `Build` pass, `Vet` pass, `Test` still pending at review time; no syntax errors in the new YAML. |

## 2. Findings

No P0 / P1 issues. The change is mergeable as-is.

### P2 — `paths-ignore` is a foot-gun if CI jobs are ever marked required

`ci.yml:7-10,13-16` skips the entire workflow for `**/*.md`, `docs/**`, `.github/ISSUE_TEMPLATE/**`. Today only `code-review` is required by the ruleset, so this is safe. The moment a maintainer adds `Build` / `Test` / `Vet` to required status checks (a natural next step), a docs-only PR will never trigger the workflow and the required checks will stay in *Expected* indefinitely — branch protection will silently block the PR with no obvious cause.

Two robust options:
1. Replace `paths-ignore` with a `paths-filter` job (e.g. `dorny/paths-filter`) that emits outputs, and gate `build` / `test` / `vet` via `if:` — the jobs then exist and report `skipped`, which protection treats as success.
2. Keep `paths-ignore`, but add a tiny always-running `ci-required` summary job that other workflows can be required against.

Not a blocker, but worth a follow-up issue before adding any of these to required checks.

### P2 — Cron at exactly `0 3 * * *` lands on a GitHub scheduler magnet hour

`stale.yml:5` uses `"0 3 * * *"`. GitHub explicitly warns that schedules at the top of the hour are often delayed by 10–30+ minutes due to congestion (and may be skipped during incidents). For a daily stale sweep this is mostly cosmetic, but a small offset (e.g. `"17 3 * * *"`) gives more predictable runs and is the convention applied to similar daily jobs in other Mininglamp-OSS repos.

### Nit — trailing blank line in `labeler.yml`

`labeler.yml` ends with an extra blank line after line 28 (final two newlines). Harmless, but `yamllint`/`prettier` configs will flag it.

### Nit — `secrets: intentionally omitted` deserves a one-line contrast with PR #25's usage example

`labeler.yml:27-28` correctly notes the reusable workflow only needs `GITHUB_TOKEN`. PR #25's how-to example shows `secrets: inherit`. Both work; the divergence is deliberate (tighter blast radius for fork PRs invoked via `pull_request_target`). Calling that out in the comment would save a future reader from "wait, the docs said `secrets: inherit`…" confusion. Suggestion only.

### Nit — Consider documenting the `ready_for_review` + `converted_to_draft` rationale inline

The `types:` expansion at `ci.yml:5` is correct and non-obvious. A one-line comment ("re-trigger when draft is lifted; allow concurrency-group cancellation when reverted to draft") would help future maintainers avoid trimming the list back to the default.

## 3. Suggestions

- File a follow-up issue: "Make `Build` / `Test` / `Vet` safely required" — pre-condition is fixing the `paths-ignore` interaction described above. Cheap to do now, painful to debug later.
- When other repos in the org adopt the same `stale.yml` / `labeler.yml`, stagger their cron minutes (e.g. octo-lib `17`, octo-server `23`, dmwork-adapters `29`) so a single org-wide scheduler hiccup doesn't drop every repo's run.
- Optional: pass explicit values to the reusable stale call rather than relying on defaults (`days_before_issue_stale: 14`, etc.) — makes the caller self-documenting and avoids silent behavior shifts if the reusable workflow's defaults ever change.

## 4. Additional Observations

- **Security model for `pull_request_target` is sound.** The inline comment in `labeler.yml:4-11` is exactly the rationale reviewers should see; the workflow never checks out the PR ref, the reusable callee only consumes API inputs (`pr_number`, `repo_owner`, `repo_name`), and `github-script` is pinned by SHA. This is the right pattern for fork-PR labeling and is consistent with `auto-add-to-project.yml` / `octo-pr-feed.yml`.
- **Concurrency group** at `ci.yml:21-23` already exists and pairs correctly with the new `converted_to_draft` type — converting ready→draft now cancels in-flight jobs (CI-minute savings).
- **Top-level `permissions: {}` + per-job `permissions:`** in both new workflows is the recommended least-privilege pattern. Good.
- **No `secrets: inherit` exposure on a `pull_request_target` workflow** — important hardening for fork-PR scenarios; do not regress this in future edits.

## Verdict

APPROVED. Findings are all P2 / nit and can be addressed in follow-up PRs without blocking this change. The CI refinements are correct, the new automation workflows match the merged reusable contracts, and the security boundary around `pull_request_target` is properly handled and documented.

Jerry-Xin (dismissed):
Summary: The PR is relevant to `octo-lib` and the workflow changes are scoped and generally consistent with the repo’s shared automation model.

💬 Non-blocking

🟡 Warning — [.github/workflows/ci.yml](/private/tmp/pr-review-EKedPv/repo/.github/workflows/ci.yml:7): `paths-ignore` skips the whole CI workflow for docs/markdown-only PRs. If `CI`, `Build`, `Test`, or `Vet` are required branch protection checks, GitHub can leave skipped path-filtered checks pending and block merge. Consider a lightweight always-reporting fallback check, or confirm these checks are not required for ignored paths. ([docs.github.com](https://docs.github.com/en/actions/how-tos/manage-workflow-runs/skip-workflow-runs))

🟡 Warning — [.github/workflows/labeler.yml](/private/tmp/pr-review-EKedPv/repo/.github/workflows/labeler.yml:19), [.github/workflows/stale.yml](/private/tmp/pr-review-EKedPv/repo/.github/workflows/stale.yml:12): The new reusable workflow calls use `@main`, while existing repo callers pin shared workflows to `@v1`. This is acceptable for the stated dependency rollout, but it should be pinned to a released tag or SHA once the shared workflows are available, especially for the `pull_request_target` labeler with write permissions.

✅ Highlights

✅ Draft PR handling is correctly implemented with job-level `if` guards, and `ready_for_review` is included so CI resumes when drafts become reviewable.

✅ Permissions are kept narrow: default workflow permissions are empty/read-only, with write scopes granted only where label/stale automation needs them.

✅ The `pull_request_target` caller does not check out PR code or pass secrets, which keeps the fork-PR labeler pattern appropriately constrained.

lml2468 (commented):
**Re-review after new commit** (`69f0f39` → `5e810c1`)

Incremental change: single file `.github/workflows/labeler.yml`

**Previous blocking findings — resolution status:**

| # | Finding | Status |
|---|---|---|
| 🔴 | `secrets: inherit` unnecessary secret exposure | ✅ **Resolved** — removed entirely, with clear comment explaining why omitted |
| 🟡 | Fork PR labeling 403 | ✅ **Resolved** — switched to `pull_request_target` with thorough security rationale in comments. No checkout of PR code, metadata-only automation. Matches existing `auto-add-to-project.yml` and `octo-pr-feed.yml` pattern |

**Remaining non-blocking items (unchanged, accepted as-is):**
- `@main` ref pinning — to be aligned when `.github` repo cuts a tag
- `converted_to_draft` empty runs — cosmetic, defensible either way

**Verdict: APPROVED** (submitted as COMMENT due to self-approve limitation)

Both blocking issues are properly resolved. The `pull_request_target` migration is well-documented with inline security justification.

Jerry-Xin (changes requested):
The PR is relevant to `octo-lib`, but the CI path filter does not implement the advertised docs-only skip behavior.

🔴 Blocking

- 🔴 Critical: `.github/workflows/ci.yml:48-62` uses `dorny/paths-filter` with a broad positive pattern (`'**'`) plus negated exclusions, but it does not set `predicate-quantifier: every`. In `dorny/paths-filter@v3`, the default is `some`, so a docs-only file such as `README.md` still matches `'**'`, making `steps.filter.outputs.code == 'true'`. As a result, markdown/docs-only PRs will still run `Build`, `Test`, and `Vet`, contradicting the PR’s main CI change. Add:
  ```yaml
  predicate-quantifier: every
  ```
  under the `with:` block for `dorny/paths-filter`.

💬 Non-blocking

- 🟡 Warning: `.github/workflows/labeler.yml:19` and `.github/workflows/stale.yml:12` depend on reusable workflows at `Mininglamp-OSS/.github@main`. This is called out in the PR description, but these workflows will fail if this PR is merged before the shared workflow PR lands on `main`. For a privileged `pull_request_target` workflow, consider whether the org prefers a stable tag like the existing `@v1` automation workflows.

✅ Highlights

- The PR correctly keeps PR-level CI path filtering inside jobs instead of `pull_request.paths-ignore`, avoiding missing required check runs.
- `pull_request_target` is documented and scoped to metadata/API automation, with no PR checkout in this wrapper workflow.
- `actionlint` passes for the changed workflow files.

yujiawei (dismissed):
# Code Review — PR #16 (octo-lib)

## Summary

Refines CI (draft-skip, docs-skip via preflight) and adds two automation workflows (`stale.yml`, `labeler.yml`) that delegate to reusable workflows in `Mininglamp-OSS/.github`. The design is well thought out, comments document the non-obvious choices (workflow-level `paths-ignore` deadlock on PRs, `pull_request_target` safety rationale, `converted_to_draft` producing skipped checks), and SHA pinning is applied to the new third-party action calls. Dependency PR #25 in `.github` is merged on `main`, so the reusable workflow refs resolve. CI on this branch is green (Build/Test/Vet/Detect changed paths all passed).

No P0 or P1 blockers. A handful of P2/P3 suggestions below — none block merge.

## Verification

- ✅ Dependency PR `.github#25` is merged to `main` (verified via API, merged 2026-05-19T08:12:09Z); both `reusable-pr-labeler.yml` and `reusable-stale.yml` are present on `.github` `main`.
- ✅ `pull_request_target` in `labeler.yml` does not check out PR code; the reusable workflow only calls REST APIs with `actions/github-script@<sha>` — the documented safety claim holds.
- ✅ Workflow-level `permissions: {}` in both new workflows (least privilege at top); writes are scoped per-job to `issues: write` and `pull-requests: write`.
- ✅ `changes` preflight: `dorny/paths-filter@v3` is SHA-pinned (`6852f92c…`), `actions/checkout` is SHA-pinned (`34e11487…`), and the `code` output's fallback `(github.event_name != 'pull_request' && 'true' || steps.filter.outputs.code)` correctly defaults to `'true'` for push events where the `filter` step is gated off.
- ✅ Negation patterns in `dorny/paths-filter` (`!**/*.md`, `!docs/**`, `!.github/ISSUE_TEMPLATE/**`) work as the inline comment claims — micromatch evaluates `!` exclusions after positive matches, and the default `predicate-quantifier: any` is what's wanted here.
- ✅ Symmetry: `push.paths-ignore` mirrors the PR-side filter list, so docs-only pushes to `main` and docs-only PRs both skip the heavy jobs.
- ✅ Current PR CI: Build / Test / Vet / Detect changed paths all pass on `7b02d14`.

## Findings

### P2 — `@main` floating ref for reusable workflows

`stale.yml:12` and `labeler.yml:19` reference `Mininglamp-OSS/.github/.github/workflows/reusable-*.yml@main`, while existing precedent in this repo (`auto-add-to-project.yml`, `octo-pr-feed.yml`) pins to `@v1`. The `v1` tag in `.github` predates these new reusables (verified: `reusable-pr-labeler.yml` and `reusable-stale.yml` 404 at `?ref=v1`), so `@main` is a pragmatic shortcut, but it means any merge to `.github@main` silently changes behavior here — including changes that touch label management, comment creation, and stale-close decisions.

Suggestion: after `.github` cuts a new tag (e.g. `v2`), update both files to the tagged ref; or pin to the merge SHA of `.github` PR #25 in the interim. Even a comment noting the intent to migrate to a tagged ref would help future maintainers.

### P2 — `test` no longer depends on `build`

`ci.yml` `test` job changes from `needs: build` to `needs: [changes]` (line 85). Tests now run in parallel with build instead of after it, which is faster but means:
- A build failure no longer cancels in-flight tests; both jobs will execute the same compilation work, doubling CI minutes on broken-build PRs.
- If `go test` happens to fail with a confusing message rooted in the same compile error that `go build` would surface clearly, that signal is muddier.

If the parallelism was intentional (faster CI on the happy path is the more common case), consider mentioning it in the commit message; otherwise restoring `needs: [changes, build]` keeps the build-as-fail-fast property without losing the path-skip behavior.

### P3 — Action pinning inconsistency within `ci.yml`

The new `changes` job pins to SHAs (`actions/checkout@34e11487…`, `dorny/paths-filter@6852f92c…`). The pre-existing `build`/`test`/`vet` jobs still use `actions/checkout@v4` and `actions/setup-go@v5` (lines 72–73, 87–88, 102–103). This isn't introduced by this PR, but since you're already touching these jobs it's worth either pinning them in this PR for consistency or filing a follow-up so the practice is uniform across the file.

### P3 — `labeler.yml` does not trigger on `edited`

`labeler.yml:13` listens for `[opened, synchronize, reopened]`. Edits to PR title or body don't trigger relabeling. This is usually fine because the reusable labeler keys off file changes (size + dependency files), and a `synchronize` will re-run on any push. Flagging only because if title/body conventions later drive labels (e.g. conventional-commit prefix → label), `edited` will need to be added.

### P3 — `concurrency.cancel-in-progress` for push to `main`

`ci.yml:25-27` (pre-existing) cancels in-progress runs of the same workflow on push to `main`. With rapid commits this can cancel an in-progress validation of an earlier commit, leaving no recorded "Build passed on SHA X" for the canceled SHA. Acceptable for low-traffic `main`, but worth knowing if you ever wire required-status checks against historical SHAs.

### P3 — Naming nit in the reusable labeler (out-of-scope, but flagging)

The reusable workflow in `.github` comments `// Calculate net lines excluding lockfiles` then computes `(f.additions || 0) + (f.deletions || 0)`, which is churn (additions + deletions), not net (additions − deletions). The size buckets are calibrated around churn so the behavior is fine; only the comment/naming is misleading. Cosmetic, lives in the other repo, mentioning here so it's not lost.

## Recommendation

The PR is mergeable as-is. The two P2 items (`@main` ref pinning, `test` no longer needing `build`) are worth a thought before merge but neither is a correctness or security blocker. The P3 items are housekeeping.

Nice touches worth calling out: the inline comments on `paths-ignore` deadlock and `pull_request_target` safety are exactly the kind of context that prevents the next maintainer from "simplifying" this into a footgun.

lml2468 (commented):
**Re-review after new commit** (`5e810c1` → `7b02d14`)

Incremental change: `.github/workflows/ci.yml` — replaced workflow-level `paths-ignore` with a `dorny/paths-filter` preflight job (`changes`). 

**Verdict: CHANGES_REQUESTED** (submitted as COMMENT due to self-approve limitation)

🔴 Blocking

1. **`predicate-quantifier` must be `every` — independently confirmed Allen's finding**

   The `dorny/paths-filter` config at lines 48–62 uses negation patterns (`!**/*.md`, `!docs/**`, etc.) with `predicate-quantifier` defaulting to `some`. The official docs explicitly state:

   > "The filters below will match markdown files despite the exclusion syntax UNLESS you specify 'every' as the predicate-quantifier parameter."

   With `some` (default), a file only needs to match ANY one pattern. Since `README.md` matches `'**'`, the negation `'!**/*.md'` is irrelevant — `code` will be `true` for docs-only PRs, defeating the entire purpose of the preflight job.

   The inline comment (lines 53–56) claiming "predicate-quantifier defaults to 'any' which is exactly what we want" is incorrect on both counts — the default is `some`, and it is NOT what we want.

   **Fix:** add `predicate-quantifier: 'every'` to the `dorny/paths-filter` step config.

🟡 Non-blocking

1. **`test` job lost `needs: build` dependency**
   Previous version: `test` had `needs: build` (sequential). New version: `needs: [changes]` (parallel with build). For Go projects `go test` does its own compilation, so this is a valid optimization. But it means a build failure won't prevent test from running (wasting ~30s of CI minutes on a doomed run). Minor — parallel is defensible for faster feedback on the common success path.

✅ Highlights
- Replacing workflow-level `paths-ignore` with job-level `dorny/paths-filter` is the right architecture — skipped jobs still create check entries that satisfy branch protection
- `persist-credentials: false` on checkout is good security hygiene
- Push event default to `code: true` via `github.event_name != 'pull_request' && 'true' || steps.filter.outputs.code` is correct
- `paths-ignore` retained on `push` trigger (main branch) is appropriate — push events don't need branch protection check entries
- Action pinned to full commit SHA (`dorny/paths-filter@6852f92c...`) with version comment — excellent supply chain practice

lml2468:
## Fix pushed (41adae6) — filter strategy overhaul

**Blocking finding (predicate-quantifier semantics)**

After iterating through three approaches, settled on the simplest and most unambiguous solution:

**Previous (rejected): negation + `some`**
```yaml
code: ['**', '!**/*.md', ...]  # some mode
```
Reviewers raised concerns that negation may not correctly exclude files in `some` mode.

**Attempted (rejected): `docs_only` + `every`**
```yaml
docs_only: ['**/*.md', 'docs/**', ...]  # every mode
```
Rejected after consulting dorny/paths-filter semantics: `predicate-quantifier: every` means a single file must match ALL listed patterns simultaneously — so `README.md` (matches `**/*.md` but not `docs/**`) would always yield `docs_only=false`.

**Final fix (current): explicit positive patterns, no negation**
```yaml
code:
  - '**/*.go'
  - 'go.mod'
  - 'go.sum'
  - 'Makefile'
  - '.github/workflows/**'
```
`code=true` when at least one changed file matches a known source pattern (`some` default). No negation, no predicate-quantifier edge cases.

Truth table:
| Scenario | code | CI |
|---|---|---|
| Only README.md changed | false | skipped ✅ |
| Only main.go changed | true | runs ✅ |
| README.md + main.go | true | runs ✅ |
| Push to main | true (default) | runs ✅ |

Jerry-Xin (changes requested):
One-line summary: The PR is relevant to `octo-lib`, but the CI path filter can skip CI for a real source file type already present in the repository.

🔴 Blocking

- 🔴 Critical: `.github/workflows/ci.yml:59`-`.github/workflows/ci.yml:65` defines `code` changes as only Go files, Go modules, `Makefile`, and workflow files. This repository contains `pkg/wkhook/webhook.proto`, which is project source/API schema, but a PR changing only that `.proto` file would set `code=false` and skip `Build`, `Test`, and `Vet`. Since skipped jobs count as passing, this creates a CI bypass for non-doc source changes. Add `**/*.proto` at minimum, or invert the logic so only known documentation/markdown-only paths skip full CI.

💬 Non-blocking

- 🟡 Warning: `.github/workflows/labeler.yml:19` and `.github/workflows/stale.yml:12` call reusable workflows from `Mininglamp-OSS/.github@main`. The PR description documents the dependency, but `@main` means behavior can change without a corresponding PR in this repo. Consider moving to a stable tag such as `@v1` once the shared workflows are released.

- 🔵 Suggestion: `.github/workflows/labeler.yml:29` has an extra blank line flagged by `git diff --check`. This is minor, but worth cleaning up while touching the file.

✅ Highlights

- ✅ The PR correctly avoids workflow-level `paths-ignore` for pull requests, which prevents required checks from getting stuck with no check run.
- ✅ The `pull_request_target` labeler includes a clear security comment and only passes metadata to the reusable workflow, matching the existing shared-workflow style in this repo.

yujiawei (dismissed):
# Code Review — PR #16 (octo-lib)

## Summary

Three CI changes: (1) refine `ci.yml` to skip draft PRs and docs-only changes via a `dorny/paths-filter` preflight; (2) add `stale.yml` calling the org-shared reusable stale workflow; (3) add `labeler.yml` calling the reusable PR labeler via `pull_request_target`.

Head SHA `41adae6` resolves the prior blocking findings from earlier reviews (predicate-quantifier semantics on a negation-based filter, `secrets: inherit` over-exposure, fork-PR write-permission). The current implementation uses positive patterns only — no negation, no `predicate-quantifier` edge cases — and the security surface around the privileged `pull_request_target` workflow is correctly minimized. No P0/P1 issues remain.

## 1. Verification

| Item | Status | Evidence |
|---|---|---|
| Previous blocker: predicate-quantifier on negation patterns | ✅ Resolved | `ci.yml:59-65` switched to positive patterns (`**/*.go`, `go.mod`, `go.sum`, `Makefile`, `.github/workflows/**`). `code=true` when ≥1 changed file matches a source pattern; docs-only PRs match nothing → `code=false` → jobs skipped. No `predicate-quantifier` needed; default `some` is the intended semantics here. |
| Previous blocker: `secrets: inherit` over-exposure | ✅ Resolved | `labeler.yml:27-28` documents the deliberate omission. Verified against `Mininglamp-OSS/.github/.github/workflows/reusable-pr-labeler.yml@main`: no `secrets:` block on `workflow_call`, no `${{ secrets.* }}` references — the inherited `GITHUB_TOKEN` plus caller-granted `issues: write` / `pull-requests: write` is sufficient. |
| Previous blocker: fork-PR labeling 403 | ✅ Resolved | `labeler.yml:12` uses `pull_request_target` with documented `zizmor: ignore[dangerous-triggers]` rationale. No `actions/checkout` of PR ref anywhere in the caller or reusable chain; the reusable workflow only calls `actions/github-script@<sha>` against the REST API. |
| Dependency PR `.github#25` is merged on `main` | ✅ | Both `reusable-pr-labeler.yml` and `reusable-stale.yml` are present on `Mininglamp-OSS/.github@main` (verified via API). |
| Caller↔callee input contract for labeler | ✅ | `labeler.yml:20-23` passes `pr_number: number`, `repo_owner: string`, `repo_name: string` — exact match to the reusable workflow's `workflow_call.inputs`. |
| Caller↔callee for stale (no inputs) | ✅ | All four `days_*` inputs in `reusable-stale.yml` have defaults (14 / 7 / 14 / 7). Caller relies on defaults — acceptable. |
| `paths-filter` is step-level, not workflow-level | ✅ | The preflight `changes` job always runs and emits a status. Downstream `Build`/`Test`/`Vet` get `skipped` via the `needs.changes.outputs.code == 'true'` gate, which GitHub counts as passing for required-check evaluation. No workflow-level `paths-ignore` on `pull_request` → no deadlock risk. |
| Push events default `code=true` | ✅ | `ci.yml:44` ternary: `github.event_name == 'pull_request' && steps.filter.outputs.code || 'true'`. For push, the LHS is `false` → returns `'true'`; for PR with `code=false`, the string `'false'` is truthy in GH expressions but is still the value `'false'`, so the gate evaluates correctly. |
| `paths-ignore` on `push` mirrors PR filter | ✅ | `ci.yml:17-20` (`**/*.md`, `docs/**`, `.github/ISSUE_TEMPLATE/**`) is symmetric to the PR-side skip set, so docs-only pushes to `main` and docs-only PRs are treated consistently. |
| Required-check impact | ✅ | Repo ruleset `16278090` lists only `code-review` as required. `Build`/`Test`/`Vet` are advisory, so a docs-only PR that skips them cannot deadlock merge. |
| Top-level `permissions: {}` + per-job grants | ✅ | Both new workflows. Standard least-privilege pattern. |
| Action pinning in `changes` job | ✅ | `actions/checkout@34e11487…`, `dorny/paths-filter@6852f92c…` — SHA-pinned. Reusable workflow callees pin `actions/stale@5bef64f…` and `actions/github-script@f28e40c7…`. |
| `persist-credentials: false` on PR-context checkout | ✅ | `ci.yml:49`. Good hardening since `changes` runs before downstream gates. |
| CI on this SHA | ✅ | Build / Vet / Detect changed paths pass on `41adae6`; Test in progress (no syntactic issue, only runtime). |
| `converted_to_draft` re-trigger behavior | ✅ | `ci.yml:5` types list includes `ready_for_review` and `converted_to_draft`. Draft transitions re-evaluate; jobs run on ready, skip on draft. |

## 2. Findings

No P0 or P1 issues. The PR is mergeable as-is.

### P2 — `Makefile` pattern in filter list has no corresponding file in this repo

`ci.yml:64` includes `Makefile` in the source-pattern list, but `Mininglamp-OSS/octo-lib` doesn't currently ship a `Makefile`. This is harmless (a pattern that matches nothing simply doesn't contribute to `code=true`) and arguably good forward-defense if one is added later. Worth mentioning only so a future maintainer doesn't read it as evidence that the filter is stale or copy-pasted from a sibling repo by accident — a one-line comment ("kept for future use; this repo doesn't have one yet") would make the intent obvious.

### P2 — `test` no longer depends on `build`

In the prior workflow, `test` had `needs: build`. In `ci.yml:88` it is now `needs: [changes]`, running in parallel with `build`. Two consequences:

- A broken-build PR now consumes test minutes that would previously have been short-circuited (Go runs its own compile inside `go test`, so the test will fail the same way the build would — minutes wasted but no functional issue).
- The test failure message on a broken build may be noisier than the equivalent `go build` error, which can momentarily mask the real cause for a reviewer who only opens the Test log.

Both are minor and the parallelism is defensible on the happy path. If you want belt-and-braces, `needs: [changes, build]` keeps the fast-fail property without losing the path-skip behavior. No change required.

### P2 — Reusable workflows pinned to `@main` rather than a tag

`stale.yml:12` and `labeler.yml:19` reference `Mininglamp-OSS/.github@main`. Other automation workflows in this repo (`auto-add-to-project.yml`, `octo-pr-feed.yml`, `octo-ci-status.yml`) pin to `@v1`. The new reusables are not on the `v1` tag yet (verified: 404 at `?ref=v1`), so `@main` is a pragmatic interim choice, but it means any merge to `.github@main` immediately affects label management, comment creation, and stale-close decisions across consumers with `issues: write` + `pull-requests: write`. The risk is amplified for `labeler.yml` because the trigger is `pull_request_target`.

Suggestion: once a new tag (e.g. `v2`) is cut on `.github`, pin both files to that tag; in the meantime, pin to the merge SHA of `.github#25` if you want a stable surface today. Not a merge blocker.

### P2 — Cron at `0 3 * * *` lands on a GitHub scheduler magnet hour

`stale.yml:5` uses `"0 3 * * *"`. GitHub explicitly notes that schedules at the top of the hour are commonly delayed by 10–30 minutes due to scheduler congestion, and may be skipped entirely during incidents. For a daily stale sweep this is mostly cosmetic, but a small offset (e.g. `"17 3 * * *"`) gives more predictable runs and would let other Mininglamp-OSS repos stagger their own daily sweeps without colliding when they adopt the same template (octo-lib `17`, octo-server `23`, dmwork-adapters `29`, etc.).

### P3 — `converted_to_draft` produces empty workflow runs

Including `converted_to_draft` in `pull_request.types` means moving a ready PR back to draft triggers the workflow with all three jobs skipped (the draft gate evaluates `if:` to false). This appears as an Actions run with 0/3 jobs run, which is functionally fine but cosmetically noisy in the checks list. The inline comment at `ci.yml:6-8` explains the rationale (avoiding stale "pending" statuses from the last real push); keeping it is defensible. If the noise becomes a problem, dropping `converted_to_draft` from the types list achieves the same final state because the prior CI status remains attached to the head SHA.

### P3 — `labeler.yml` does not trigger on `edited`

`labeler.yml:13` listens for `[opened, synchronize, reopened]`. Edits to PR title or body don't trigger relabeling. Today this is fine because the reusable labeler keys off file diffs (size buckets + dependency files). If title/body conventions ever drive labels (e.g. conventional-commit prefix → label), `edited` will need to be added.

### Nit — Key ordering across the three Go jobs is inconsistent

`build` and `vet` open with `needs:` → `if:` → `name:`; `test` opens with `if:` → `name:` → `runs-on:` → `needs:`. All correct semantically; just reads less uniformly when diffing the three side-by-side. Worth a one-line cleanup if you touch this file again.

### Nit — Trailing blank line in `labeler.yml`

`labeler.yml` ends with two newlines after the comment block at lines 27-28. `yamllint`/`prettier` configs flag it; harmless otherwise.

## 3. Suggestions

- File a follow-up issue: pin the reusable workflows to a tagged ref once `.github` cuts one. Same change should land across all consuming repos at the same time.
- File a follow-up issue: when adding `Build` / `Test` / `Vet` to required status checks (a natural next step), verify the `changes` preflight produces `skipped` for docs-only PRs as expected — easy to do with a one-line README-only test PR before flipping required status.
- Optional: pass explicit `days_*` inputs to the reusable stale call instead of relying on defaults — makes the caller self-documenting and protects against silent shifts if defaults are tuned upstream.
- Optional: align the pre-existing `build`/`test`/`vet` jobs' `actions/checkout@v4` / `actions/setup-go@v5` to SHA pins for consistency with the new `changes` job's pinning style. Out of scope for this PR.

## 4. Additional Observations

- The inline rationale comments at `ci.yml:6-8`, `ci.yml:10-14`, `ci.yml:40-43`, `ci.yml:54-58`, and `labeler.yml:4-11` are exactly the kind of context that prevents the next maintainer from "simplifying" this into a footgun (workflow-level `paths-ignore` deadlock, `pull_request_target` safety boundary, push-event default for the `code` output). Keep this style.
- The concurrency group at `ci.yml:25-27` correctly pairs with the new `converted_to_draft` type — converting ready→draft now cancels in-flight jobs (CI-minute savings).
- `pull_request_target` security boundary is properly handled: no PR ref checkout, metadata-only inputs (`pr_number`, `repo_owner`, `repo_name`), reusable callee is SHA-pinned at the action level. Do not regress this in future edits.
- The history of this PR (negation → `every` quantifier → positive patterns) is a useful artifact — the final approach is the right one and the inline comment at `ci.yml:54-58` captures why.

## Verdict

APPROVED. All previously raised blockers are properly addressed at `41adae6`. The remaining items are P2 stylistic / future-proofing concerns and small nits that can be batched into a follow-up PR without blocking this change. Please confirm the merge sequence (`.github#25` already on `main` — verified) before merging.

lml2468 (commented):
**Re-review after new commit** (`7b02d14` → `41adae6`)

Incremental change: `.github/workflows/ci.yml` — complete rewrite of path filter from negation-based to positive-only patterns.

**Previous blocking findings — resolution status:**

| # | Finding | Status |
|---|---|---|
| 🔴 `predicate-quantifier` must be `every` | ✅ **Resolved (differently)** — switched to positive-only patterns (`**/*.go`, `go.mod`, `go.sum`, `Makefile`, `.github/workflows/**`), eliminating the need for negation and `predicate-quantifier` entirely. Cleaner approach. |

**New findings this round:**

🟡 Non-blocking (respectfully disagree with Allen's severity)

1. **`.proto` file not covered by code filter**
   `pkg/wkhook/webhook.proto` exists in the repo and is project source (API schema). A PR touching only this file would set `code=false` and skip CI.

   However, I disagree this is blocking because:
   - The current CI has **no protoc step** — `go build/test/vet` cannot validate `.proto` consistency
   - In practice, `.proto` changes should always come with regenerated `.pb.go` files (which would trigger CI via `**/*.go`)
   - If someone changes `.proto` without regenerating, CI running wouldn't catch it anyway — Go toolchain ignores `.proto` files
   - The real fix for proto drift detection is a separate CI job running `protoc --check` or similar

   That said, adding `'**/*.proto'` is a 1-line change with zero downside — it acts as a canary (confirms Go code still compiles after proto-adjacent changes). **Recommend adding it, but not blocking on it.**

2. **`test` job still has `needs: [changes]` instead of `needs: [changes, build]`** (carried forward from previous round, non-blocking)

✅ Highlights
- Positive-only pattern approach is significantly cleaner than the negation-based filter. No `predicate-quantifier` ambiguity, easy to reason about, easy to extend.
- Comments accurately describe the behavior now (unlike the previous incorrect comment about predicate-quantifier).
- Output expression `github.event_name == 'pull_request' && steps.filter.outputs.code || 'true'` correctly handles push events.

**Verdict: APPROVED** (submitted as COMMENT due to self-approve limitation)

The `predicate-quantifier` blocking issue is resolved. The `.proto` gap is real but has no practical impact without a protoc CI job — recommend adding the pattern but not blocking.

lml2468:
## Fix pushed (30ce8ea)

**Blocking (Jerry-Xin) — `**/*.proto` missing from code filter**

Added `'**/*.proto'` to the positive-pattern filter list.

```yaml
filters: |
  code:
    - '**/*.go'
    - '**/*.proto'   # ← added
    - 'go.mod'
    - 'go.sum'
    - 'Makefile'
    - '.github/workflows/**'
```

This covers `pkg/wkhook/webhook.proto` (confirmed present). In practice proto changes should always accompany regenerated `.pb.go` files (which trigger via `**/*.go`), but the filter should be accurate regardless.

**Nit — trailing blank line in `labeler.yml`**

Trimmed.

yujiawei (approved):
# Code Review — PR #16 (octo-lib)

## Summary

Three focused CI changes at head `30ce8ea`:

1. `ci.yml`: skip draft PRs; replace workflow-level `paths-ignore` (which would deadlock required-check protection on PRs) with a `dorny/paths-filter` preflight job that gates `Build`/`Test`/`Vet` via `if:`. Push to `main` retains `paths-ignore` (safe — no required-check requirement on push).
2. `stale.yml`: new caller that delegates to the org-shared `reusable-stale.yml`.
3. `labeler.yml`: new caller that delegates to the org-shared `reusable-pr-labeler.yml` via `pull_request_target`, with no PR-ref checkout and no `secrets: inherit`.

The PR has iterated through the prior blocking findings (predicate-quantifier semantics on negation patterns, `secrets: inherit` over-exposure, fork-PR write-token 403, missing `**/*.proto`) and the current head resolves all of them. No P0/P1 issues remain. Findings below are P2/P3/nit and can land in a follow-up.

## 1. Verification

| Item | Status | Evidence |
|---|---|---|
| Prior blocker: predicate-quantifier on negation patterns | ✅ Resolved | `ci.yml:59-66` uses positive patterns only (`**/*.go`, `**/*.proto`, `go.mod`, `go.sum`, `Makefile`, `.github/workflows/**`). Default `some` is correct here — `code=true` iff ≥1 changed file matches a source pattern; docs-only PRs match nothing → `code=false` → downstream jobs skipped. |
| Prior blocker: `secrets: inherit` over-exposure | ✅ Resolved | `labeler.yml:27-28` documents the deliberate omission. Verified against `Mininglamp-OSS/.github/.github/workflows/reusable-pr-labeler.yml@main`: no `secrets:` block on `workflow_call`, no `${{ secrets.* }}` references — inherited `GITHUB_TOKEN` plus caller-granted `issues: write` / `pull-requests: write` is sufficient. |
| Prior blocker: fork-PR labeling 403 | ✅ Resolved | `labeler.yml:12` uses `pull_request_target` with documented `zizmor: ignore[dangerous-triggers]` rationale. No `actions/checkout` of PR ref anywhere in the caller or reusable chain; the reusable callee only calls `actions/github-script@<sha>` against REST APIs. |
| Prior blocker: `**/*.proto` missing from filter | ✅ Resolved | `ci.yml:62` adds `'**/*.proto'`. `pkg/wkhook/webhook.proto` is confirmed present in the repo, so a proto-only change now triggers CI. |
| Prior nit: trailing blank line in `labeler.yml` | ✅ Resolved | File now ends after line 28. |
| Dependency PR `.github#25` is merged on `main` | ✅ | Both `reusable-pr-labeler.yml` and `reusable-stale.yml` are present on `Mininglamp-OSS/.github@main`. |
| Caller↔callee input contract for labeler | ✅ | `labeler.yml:20-23` passes `pr_number: number`, `repo_owner: string`, `repo_name: string` — exact match to the reusable workflow's `workflow_call.inputs` (all three required). |
| Caller↔callee for stale (no inputs) | ✅ | All four `days_*` inputs in `reusable-stale.yml` have defaults (14/7/14/7). Caller relies on defaults — acceptable. |
| `paths-filter` is step-level, not workflow-level on PR | ✅ | The `changes` preflight always runs and emits a status; `Build`/`Test`/`Vet` get `skipped` via the `needs.changes.outputs.code == 'true'` gate, which GitHub counts as passing for required-check evaluation. No workflow-level `paths-ignore` on `pull_request` → no deadlock risk. |
| Push event default `code=true` | ✅ | `ci.yml:44` ternary `github.event_name == 'pull_request' && steps.filter.outputs.code || 'true'`: for push the LHS is `false`, so the expression returns `'true'`. |
| `paths-ignore` on `push` mirrors PR filter | ✅ | `ci.yml:17-20` (`**/*.md`, `docs/**`, `.github/ISSUE_TEMPLATE/**`) is symmetric to the PR-side skip set. |
| Required-check impact | ✅ | Only `code-review` is required on `main`. `Build`/`Test`/`Vet` are advisory, so a docs-only PR that skips them cannot deadlock merge today. |
| Top-level `permissions: {}` + per-job grants | ✅ | Both new workflows. Standard least-privilege pattern. |
| Action pinning in `changes` job | ✅ | `actions/checkout@34e11487…`, `dorny/paths-filter@6852f92c…` are SHA-pinned. Reusable callees pin `actions/stale@5bef64f…` and `actions/github-script@f28e40c7…`. |
| `persist-credentials: false` on PR-context checkout | ✅ | `ci.yml:49`. Good hardening since `changes` runs before the downstream gates. |
| `converted_to_draft` re-trigger behavior | ✅ | `ci.yml:5` types list includes `ready_for_review` and `converted_to_draft`. Draft transitions re-evaluate; jobs run on ready, skip on draft. |
| CI status on `30ce8ea` | ✅ | Build pass, Vet pass, Detect changed paths pass, Test in progress at review time. No syntactic issue. |

## 2. Findings

No P0 or P1 issues. The PR is mergeable as-is.

### P2 — Reusable workflows pinned to `@main` rather than a tag

`stale.yml:12` and `labeler.yml:19` reference `Mininglamp-OSS/.github@main`, while other automation workflows in this repo (`auto-add-to-project.yml`, `octo-pr-feed.yml`, `octo-ci-status.yml`) pin to `@v1`. The new reusables are not on `v1` yet, so `@main` is a pragmatic interim choice, but any merge to `.github@main` immediately affects label management, comment creation, and stale-close decisions across consumers with `issues: write` + `pull-requests: write`. The blast-radius is amplified for `labeler.yml` because the trigger is `pull_request_target`.

Suggestion: once a new tag (e.g. `v2`) is cut on `.github`, pin both files to that tag; in the meantime, pinning to the merge SHA of `.github#25` would give a stable surface today. Not a merge blocker.

### P2 — `test` no longer depends on `build`

`ci.yml:89` is `needs: [changes]` (parallel with `build`) rather than the prior `needs: build`. Two consequences:

- A broken-build PR consumes test minutes that would previously have been short-circuited (Go runs its own compile inside `go test`, so the test will fail the same way the build would — minutes wasted but no functional issue).
- Test-log failures on a broken-build PR may be noisier than the equivalent `go build` error, momentarily masking the real cause for a reviewer who only opens the Test log.

Both minor. Parallelism is defensible on the happy path. If you want belt-and-braces, `needs: [changes, build]` keeps the fast-fail property without losing the path-skip behavior. Author has already noted they want this kept as-is; flagging for the record only.

### P2 — Cron at `0 3 * * *` lands on a GitHub scheduler magnet hour

`stale.yml:5` uses `"0 3 * * *"`. GitHub explicitly notes that schedules at the top of the hour are commonly delayed by 10–30 minutes due to scheduler congestion, and may be skipped entirely during incidents. A small offset (e.g. `"17 3 * * *"`) gives more predictable runs and would let other Mininglamp-OSS repos stagger daily sweeps when they adopt the same template (e.g. octo-lib `17`, octo-server `23`, dmwork-adapters `29`). Cosmetic for stale management; do as part of the same follow-up that addresses tag pinning.

### P3 — `converted_to_draft` produces empty workflow runs

Including `converted_to_draft` in `pull_request.types` means moving a ready PR back to draft triggers the workflow with all three jobs skipped (the draft gate evaluates `if:` to false). This appears as an Actions run with 0/3 jobs run, which is functionally fine but cosmetically noisy in the checks list. The inline comment at `ci.yml:6-8` explains the rationale (avoiding stale "pending" statuses from the last real push); keeping it is defensible. If the noise becomes a problem, dropping `converted_to_draft` from the types list achieves the same final state because the prior CI status remains attached to the head SHA.

### P3 — `labeler.yml` does not trigger on `edited`

`labeler.yml:13` listens for `[opened, synchronize, reopened]`. Edits to PR title or body don't trigger relabeling. Today this is fine because the reusable labeler keys off file diffs (size buckets + dependency files). If title/body conventions ever drive labels (e.g. conventional-commit prefix → label), `edited` will need to be added.

### P3 — `Makefile` pattern in filter list has no corresponding file in this repo

`ci.yml:65` includes `Makefile`, but `Mininglamp-OSS/octo-lib` doesn't currently ship a `Makefile`. Harmless — a pattern that matches nothing simply doesn't contribute to `code=true` — and arguably good forward-defense. A one-line comment ("kept for future use; this repo doesn't have one yet") would make the intent obvious to the next maintainer reading the filter list.

### P3 — Action pinning inconsistency within `ci.yml`

The new `changes` job pins to SHAs. The pre-existing `build`/`test`/`vet` jobs still use `actions/checkout@v4` and `actions/setup-go@v5` (lines 76–77, 91–92, 106–107). Not introduced by this PR, but since you're touching the file, either pin them here for consistency or file a follow-up so the practice is uniform across the file.

### Nit — Key ordering across the three Go jobs is inconsistent

`build` and `vet` open with `needs:` → `if:` → `name:`; `test` opens with `if:` → `name:` → `runs-on:` → `needs:`. All correct semantically; just reads less uniformly when diffing the three side-by-side. Worth a one-line cleanup if you touch this file again.

## 3. Suggestions

- File a follow-up issue to pin the reusable workflows to a tagged ref once `.github` cuts one — the same change should land across all consuming repos at the same time.
- Before adding `Build` / `Test` / `Vet` to required status checks (a natural next step), verify the `changes` preflight produces `skipped` for docs-only PRs as expected — easy to confirm with a one-line README-only test PR before flipping required status.
- Optional: pass explicit `days_*` inputs to the reusable stale call instead of relying on defaults — makes the caller self-documenting and protects against silent shifts if defaults are tuned upstream.
- Optional: align the pre-existing `build`/`test`/`vet` jobs' `actions/checkout@v4` / `actions/setup-go@v5` to SHA pins for consistency with the new `changes` job's pinning style. Out of scope for this PR.

## 4. Additional Observations

- The inline rationale comments at `ci.yml:6-8`, `ci.yml:10-14`, `ci.yml:40-43`, `ci.yml:54-58`, and `labeler.yml:4-11` are exactly the kind of context that prevents the next maintainer from "simplifying" this into a footgun (workflow-level `paths-ignore` deadlock, `pull_request_target` safety boundary, push-event default for the `code` output). Keep this style.
- The concurrency group at `ci.yml:25-27` correctly pairs with the new `converted_to_draft` type — converting ready→draft now cancels in-flight jobs (CI-minute savings).
- `pull_request_target` security boundary is properly handled: no PR-ref checkout, metadata-only inputs (`pr_number`, `repo_owner`, `repo_name`), reusable callee is SHA-pinned at the action level. Do not regress this in future edits.
- The iteration history of this PR (negation → `every` quantifier → positive patterns; `pull_request` → `pull_request_target`; +`**/*.proto`) is a useful artifact — the final approach is the right one and the inline comment at `ci.yml:54-58` captures why.

## Verdict

APPROVED. All previously raised blockers are resolved at `30ce8ea`. Remaining items are P2/P3/nit and can be batched into a follow-up PR without blocking this change. Please confirm the merge sequence (`.github#25` already on `main` — verified) before merging.

Jerry-Xin (approved):
Summary: The PR is relevant to `octo-lib` and the workflow changes are structurally sound; I found no blocking issues.

💬 Non-blocking

🟡 Warning — `.github/workflows/labeler.yml:19`, `.github/workflows/stale.yml:12`: Both reusable workflows are referenced with `@main`, while existing org reusable workflows in this repo use `@v1`. This is acceptable for rollout, especially since PR #25 is now present on `main`, but it leaves future behavior mutable without a PR in this repo. Consider moving these to a stable tag or pinned SHA after the automation contract settles.

🔵 Suggestion — `.github/workflows/ci.yml:59-66`: The positive allow-list is clear and currently matches this Go/proto repo, but it will skip CI for any future behavior-affecting file types unless the list is updated. If the project later adds scripts, Dockerfiles, generated config, or other build inputs, they should be added here.

✅ Highlights

🔵 The project relevance gate passes: CI, stale handling, and PR labeling are repository automation for this project.

🔵 `.github/workflows/ci.yml:68-102` correctly skips draft PRs and docs-only PRs at the job level, which should still produce check conclusions for required-check workflows.

🔵 `.github/workflows/labeler.yml:12-19` uses `pull_request_target` without checking out PR code and grants only issue/PR write permissions, which keeps the fork-PR labeling use case reasonably constrained.

Verification performed: `go test ./...` passed, and `actionlint` passed for the three changed workflow files.

lml2468 (commented):
**Re-review after new commit** (`41adae6` → `30ce8ea`)

Incremental diff: 2 files, both trivial.

1. `.github/workflows/ci.yml` — added `'**/*.proto'` to the code filter. Resolves the `.proto` gap noted in last round.
2. `.github/workflows/labeler.yml` — removed trailing blank line. Cosmetic.

All previous findings resolved. CI green. No new issues.

**Verdict: APPROVED** (submitted as COMMENT due to self-approve limitation)

PR #17, opened 2026-05-19
feat(ci): add CodeQL security scan and Workflow Sanity check

lml2468:
### `codeql.yml` (new)
Go CodeQL security analysis via shared reusable workflow.
- queries: `security-extended`
- Triggers: push to main / PR opened-synced-reopened / daily cron 06:00 UTC / manual
- Skip draft PRs and docs-only changes
- Concurrency group prevents parallel duplicate runs

### `workflow-sanity.yml` (new)
Validates workflow quality on CI file changes:
- No-tabs scan (Python3 across \*.yml + \*.yaml)
- actionlint v1.7.12 static analysis (SHA-256 verified download)

## Merge order
**Merge `Mininglamp-OSS/.github` PR #26 first** — `reusable-codeql.yml` and `workflow-sanity.yml` must be on `main` before these callers work.

## Part of P1 automation rollout

PR #15, opened 2026-05-19
ci: add Octo CI Status notification workflow

lml2468:
## 变更

补充 `octo-ci-status.yml` caller workflow。

octo-lib 的 CI 本身是通过的，但缺少状态变化推送。合并后当 main 分支 CI 从 success→failure 或 failure→success 时，会自动推送告警到 ci-status 群和 octo-lib 项目群。

使用 `Mininglamp-OSS/.github` 的 reusable workflow `@v1`（已在 PR #24 修复了 startup_failure 根因）。

PR #18, opened 2026-05-19
feat(ci): simplify octo-issue-feed to new/reopen only

lml2468:
## Summary

Simplifies the `octo-issue-feed.yml` caller following the product decision to:
1. Only forward **new issue (opened)** and **reopened** events — close/label/etc. are ignored
2. Send to the shared **issue-feed group only** (per-repo project group removed)
3. Remove IssueTriage bot mention (deprecated)

### Changes
- Trigger: `[opened, closed, reopened, labeled]` → `[opened, reopened]`
- Removed: `issue_labels`, `project_group_id` from `with:` block
- Switched: `@v1` → `@main` to pick up new reusable after `.github` PR #28 merges
- Added: `permissions: {}` at workflow level

## Merge order
**Merge `.github` PR #28 first** — the reusable `octo-issue-feed.yml` on main must match this caller interface before these callers work correctly.

### DIFF
--- a/.github/workflows/octo-issue-feed.yml
+++ b/.github/workflows/octo-issue-feed.yml
@@ -2,19 +2,19 @@ name: Octo Issue Feed
 
 on:
   issues:
-    types: [opened, closed, reopened, labeled]
+    types: [opened, reopened]
+
+permissions: {}
 
 jobs:
   notify:
-    uses: Mininglamp-OSS/.github/.github/workflows/octo-issue-feed.yml@v1
+    uses: Mininglamp-OSS/.github/.github/workflows/octo-issue-feed.yml@main
     with:
       repo_name: ${{ github.event.repository.name }}
       issue_number: ${{ github.event.issue.number }}
       issue_title: ${{ github.event.issue.title }}
       issue_url: ${{ github.event.issue.html_url }}
       issue_author: ${{ github.event.issue.user.login }}
-      issue_labels: ${{ toJson(github.event.issue.labels.*.name) }}
       event_action: ${{ github.event.action }}
-      project_group_id: 6095b6ded00046009e72882ce4067289
     secrets:
       OCTO_BOT_TOKEN: ${{ secrets.OCTO_BOT_TOKEN }}


Jerry-Xin (changes requested):
The workflow change is project-relevant, but it is not currently safe to merge because the referenced reusable workflow on `Mininglamp-OSS/.github@main` still requires inputs this caller now omits.

🔴 Blocking

- 🔴 Critical — `.github/workflows/octo-issue-feed.yml:11-18`: this caller now uses `Mininglamp-OSS/.github/.github/workflows/octo-issue-feed.yml@main` while removing `project_group_id`. I checked the current external `main` reusable workflow, and it still declares `project_group_id` as a required `workflow_call` input. If this PR merges before `.github` PR #28 lands, the issue-feed workflow will fail at runtime for opened/reopened issues. The merge-order note is correct, but this PR should not merge until that dependency is actually on `main`, or the reusable reference is pinned to a compatible ref.

💬 Non-blocking

- 🔵 Suggestion — `.github/workflows/octo-issue-feed.yml:11`: using `@main` makes this workflow track future reusable-workflow changes automatically. That may be intentional for centralized CI, but a stable tag or SHA would reduce accidental breakage from later interface changes.

✅ Highlights

- The PR passes the project relevance gate: this is repository CI/automation, so it belongs in `Mininglamp-OSS/octo-lib`.
- `issues.types: [opened, reopened]` matches the stated product decision.
- `permissions: {}` is consistent with the repository’s other reusable-workflow callers.
- `actionlint` reported no syntax issues for `.github/workflows/octo-issue-feed.yml`.

yujiawei (changes requested):
# Code Review — PR #18 (octo-lib)

## Summary

This PR trims the `octo-issue-feed.yml` caller in `octo-lib` to match the upcoming reusable-workflow interface in `.github` PR #28:

- Issue trigger narrowed to `[opened, reopened]`
- `issue_labels` and `project_group_id` removed from the `with:` block
- Reusable ref switched from `@v1` → `@main`
- Workflow-level `permissions: {}` added

The intent is clean and the diff is appropriately small. However, two coupling/durability issues need to be resolved before merge.

---

## Findings

### P1 — Required-input regression if merge order slips (`.github/workflows/octo-issue-feed.yml:11`)

The reusable workflow at `Mininglamp-OSS/.github` on both `main` (`44d499e`) and the `v1` tag (`0a0ec95`) currently declares `project_group_id` as a **required** input. `.github` PR #28 (which would remove that requirement) is still **OPEN** with `reviewDecision=CHANGES_REQUESTED`, so the contract change has not landed.

If this PR is merged before `.github` PR #28 merges, then on the next `issues` event in `octo-lib`:

- The job resolves `uses: Mininglamp-OSS/.github/.github/workflows/octo-issue-feed.yml@main`
- The reusable workflow fails at startup with a missing-required-input error for `project_group_id`
- Issue-feed notifications for `octo-lib` silently break

The PR description says "Merge `.github` PR #28 first", but that ordering is enforced only by human coordination. With an auto-merging dispatcher in the loop this is a real foot-gun.

**Suggested fix:** keep the ref at `@v1` (or a new `@v2`) and move the tag in `.github` *after* PR #28 merges. That way the dependency is encoded in the version, not in the merge queue.

### P1 — Switching to a floating `@main` ref breaks org convention and version pinning (`.github/workflows/octo-issue-feed.yml:11`)

A code search shows **all 8 sibling callers** in the org (`octo-server`, `octo-web`, `octo-admin`, `octo-adapters`, `octo-deployment`, `octo-smart-summary`, `octo-matter`, and `octo-lib` itself today) currently pin to `@v1`. This PR is the lone divergence to `@main`.

Pinning a reusable workflow to a moving branch ref is generally discouraged even for org-internal workflows because:

- Any future commit to `.github`'s `main` immediately changes `octo-lib`'s CI behavior with no review on this side.
- It defeats the value of having a `v1` tag (which is what the rest of the fleet relies on for reproducibility).
- It creates inconsistent versioning across the fleet — fixing `octo-lib` later requires either rewriting back to `@v1`/`@v2`, or bumping every other caller to `@main` (and losing the same guarantees there).

**Suggested fix:** decide on a tag strategy with `.github`:
- Option A: have `.github` PR #28 retag `v1` to the new commit, then this caller stays at `@v1`.
- Option B: cut a `v2` tag for the breaking input change, then pin this caller (and a follow-up fleet-wide migration) to `@v2`.

Either way, ending up on `@main` long-term is the worst option of the three.

### Nit — `permissions: {}` on a reusable-only caller is symbolic (`.github/workflows/octo-issue-feed.yml:6`)

Good hygiene, no objection. Note that since the caller defines no inline steps and the reusable already declares `permissions: {}`, this line is mostly defense-in-depth / signal-of-intent rather than a functional restriction. Worth keeping.

### Trigger filter — fine, redundant with reusable but harmless (`.github/workflows/octo-issue-feed.yml:5`)

Narrowing to `types: [opened, reopened]` is correct. The reusable (per `.github` PR #28's body) also early-exits on non-`{opened,reopened}` actions, so the filter at the caller is belt-and-suspenders. Keeping it at the caller saves a workflow-run start on every `closed`/`labeled` event, which is the right place for the filter — leave as-is.

---

## What's fine / not blocking

- Diff size is appropriate for the stated goal.
- Inputs that remain (`repo_name`, `issue_number`, `issue_title`, `issue_url`, `issue_author`, `event_action`) match the reusable workflow's `workflow_call` interface today and continue to match it after `.github` PR #28.
- `secrets.OCTO_BOT_TOKEN` plumbing is unchanged.
- No new external action references, no SHA pins needed.

---

## Verdict

`CHANGES_REQUESTED`. The change itself is correct in spirit, but as-written it's coupled to a still-open PR in a different repo via a hand-coordinated merge order *and* swaps a versioned ref for a moving branch ref. Pinning back to a tag (`@v1` retagged, or a new `@v2`) resolves both concerns simultaneously and brings this caller back in line with the other 7 sibling repos. Once that's done, this should be a quick re-review.

lml2468 (commented):
**Independent Review — PR #18**

CI green, mergeable. Single file change: simplify `octo-issue-feed.yml` caller.

**Verdict: APPROVED with merge-order constraint** (submitted as COMMENT due to self-approve limitation)

🔴 Blocking: None — **with the explicit condition that `.github` PR #28 merges first.**

Allen flagged the `project_group_id` contract mismatch as blocking. I independently verified:

- ✅ Current `@main` reusable still has `project_group_id: required: true` — correct
- ✅ `.github` PR #28 removes `project_group_id` and `issue_labels` from the contract — correct
- ✅ PR #28 is OPEN (not merged yet) — correct

**However, I classify this as a merge-order dependency, not a code defect:**

The PR description explicitly documents the constraint: "Merge `.github` PR #28 first." The caller code is correct *for the new contract*. If PR #28 merges before PR #18, everything works. If PR #18 merges first, the next opened/reopened issue triggers a workflow failure because GitHub Actions rejects missing required inputs at parse time.

This is the same pattern we accepted in PRs #15, #16, and #17 (all depending on `.github` PRs merging first). The risk is operational (merge ordering), not architectural.

🟡 Non-blocking

1. **`@v1` → `@main` transition**
   Same mutable-ref pattern as other recent PRs. Once `.github` stabilizes, should pin to a tag.

2. **`permissions: {}` added** — good hardening, consistent with other callers.

3. **No `.github/actions/**` path in push trigger** — N/A for this workflow (issue events only, no push trigger).

✅ Highlights
- Clean scope reduction: 4 event types → 2, removing `closed` and `labeled` per product decision
- Removing `project_group_id` centralizes group routing in the reusable workflow (send to issue-feed group only) — simpler, fewer secrets to manage per repo
- Removing `issue_labels` input matches the removal from the reusable's contract in PR #28
- The reusable workflow (PR #28) adds a guard `if action not in ("opened", "reopened"): sys.exit(0)` — defense in depth even if the caller accidentally broadens event types later

lml2468 (commented):
[APPROVE]

Straightforward CI simplification matching the product decision to forward only `opened`/`reopened` events.

**Changes verified:**
- Trigger narrowed: `[opened, closed, reopened, labeled]` → `[opened, reopened]` — removes noise events (close/label)
- Removed `issue_labels` / `project_group_id` — consistent with reusable workflow interface change in `.github` PR #28
- `@v1` → `@main` — intentional, picks up the updated reusable interface
- `permissions: {}` added — explicit zero-permission lock, good security hygiene

🟡 **Merge order**: PR body documents this correctly — merge `.github` PR #28 first; callers will break if merged before the reusable interface is updated.

LGTM — approve when `.github` PR #28 is merged.

PR #23, opened 2026-05-21
fix(ci): use issues: write permission for PR welcome comments

lml2468:
## Problem

The `pr-contributor-welcome.yml` caller that landed in a previous PR declared `permissions: pull-requests: write`, but PR comments are created via the GitHub Issues API (`issues.createComment`) which requires `issues: write`.

Caught by Jerry-Xin in Mininglamp-OSS/octo-adapters#53. Reusable workflow fix: Mininglamp-OSS/.github#33.

## Fix

`pull-requests: write` → `issues: write` in the caller job permissions.

PR #24, opened 2026-05-21
fix(ci): grant both issues: write and pull-requests: write

lml2468:
Follow-up permission fix. Both `issues: write` and `pull-requests: write` are needed — matches `labeler.yml` established pattern.

### DIFF
--- a/.github/workflows/pr-contributor-welcome.yml
+++ b/.github/workflows/pr-contributor-welcome.yml
@@ -17,3 +17,4 @@ jobs:
     uses: Mininglamp-OSS/.github/.github/workflows/reusable-pr-contributor-welcome.yml@main
     permissions:
       issues: write
+      pull-requests: write


Jerry-Xin (approved):
Summary: The PR is relevant to this repository and correctly updates the PR contributor welcome workflow permissions to match the established reusable workflow pattern.

✅ Highlights
- `.github/workflows/pr-contributor-welcome.yml:18-20` keeps default workflow permissions empty and grants only the job-level scopes needed by the reusable workflow.
- `.github/workflows/pr-contributor-welcome.yml:20` aligns with the existing `labeler.yml` and `stale.yml` pattern of granting both `issues: write` and `pull-requests: write`.
- 🔵 Suggestion: No additional tests are needed for this narrow GitHub Actions permission fix.

yujiawei (approved):
# Code Review — PR #24 (octo-lib)

## Summary

One-line change to `.github/workflows/pr-contributor-welcome.yml` re-adding `pull-requests: write` to the caller job that invokes `Mininglamp-OSS/.github/.github/workflows/reusable-pr-contributor-welcome.yml@main`, alongside the `issues: write` that PR #23 just landed.

## Verdict

No correctness, security, or build-break concerns. The change is harmless, but it is also a runtime no-op and the PR description overstates necessity. Approving with P2 / maintainability notes; consider whether this PR is worth merging as-is or could simply be closed.

## Findings

### P2 — Added permission has no runtime effect (caller cannot elevate beyond the reusable workflow's declared scope)

The reusable workflow `Mininglamp-OSS/.github/.github/workflows/reusable-pr-contributor-welcome.yml@main` declares its job-level permissions as:

```yaml
jobs:
  welcome:
    permissions:
      issues: write
```

Per GitHub Actions semantics, the caller's `jobs.<id>.permissions` block sets the **ceiling** for the GITHUB_TOKEN passed to the reusable workflow, but the reusable workflow's own job-level `permissions:` is what actually applies to the job's token. Granting `pull-requests: write` at the caller does not propagate into the reusable job — that job will still only have `issues: write`.

Net effect of this PR at runtime: zero. If the intent is to expand the token scope for the welcome job, the reusable workflow's job permissions need to be updated first (in `Mininglamp-OSS/.github`), not the caller.

### P2 — PR description's "both are needed" claim is incorrect for the current reusable workflow

PR body states: *"Both `issues: write` and `pull-requests: write` are needed — matches `labeler.yml` established pattern."*

What the reusable welcome workflow actually does (verified by reading `reusable-pr-contributor-welcome.yml@main`):

1. `GET /orgs/{org}/members/{username}` — uses default GITHUB_TOKEN access; no scoped write needed.
2. `search.issuesAndPullRequests` — read-only.
3. `issues.listComments` — covered by `issues: write`.
4. `issues.createComment` — requires `issues: write`. PR comments are created via the Issues API (PR conversations are issue comments under the hood); `pull-requests: write` is **not** required for this call.

The labeler analogy doesn't carry over: the labeler reusable mutates labels on PRs, which is a documented case where GitHub does require `pull-requests: write` in addition to `issues: write`. The welcome workflow only posts a comment, so the parallel does not hold.

This is exactly what PR #23's body and the upstream fix `Mininglamp-OSS/.github#33` documented when reducing the scope from `pull-requests: write` → `issues: write`. PR #24 partially walks that least-privilege reduction back without a corresponding new API call requiring it.

### P2 — Consider whether this PR is needed at all

Given:
- The reusable workflow caps effective permissions at `issues: write`, so the new line cannot grant anything extra at runtime.
- No new API call has been introduced that requires `pull-requests: write`.
- PR #23 just landed a deliberate least-privilege cleanup (issue spotted by Jerry-Xin in `octo-adapters#53`, reusable fix in `.github#33`).

Two reasonable paths forward:

1. **Close this PR.** The state after #23 is correct and minimal. Re-adding the unused permission re-introduces the very pattern that #23 cleaned up across all 16 caller repos.
2. **Keep this PR but tighten the rationale.** If maintainers want every caller's permission block to look identical to `labeler.yml` for cognitive consistency, that's a defensible stance — but the PR description should say so explicitly ("defensive consistency, not functional necessity") rather than asserting both permissions are required.

Either choice is reasonable; my preference would be (1) for principle-of-least-privilege hygiene, especially given how recent the cleanup was.

## Non-issues / Looks good

- File path, indentation, and YAML structure are correct.
- No change to the `pull_request_target` trigger or to the (correctly empty) top-level `permissions: {}`. The `zizmor: ignore[dangerous-triggers]` comment and the security note about not checking out PR code remain intact.
- No risk of regressions to existing welcome behavior; worst case, the line is dead config.

## Recommendation

Approve. Maintainer to decide whether to merge as defensive consistency or close in favor of the post-#23 minimal state. If merging, recommend updating the PR description to clarify the change is no-op-at-runtime and motivated by uniformity with `labeler.yml`, not by a missing token scope.

lml2468 (commented):
[REQUEST_CHANGES] — reviewer account matches PR author; flagging for a human maintainer. Do not merge.

## Root Cause (from run logs)

The `welcome / welcome` CI check **FAILED** 🔴. Here is the exact failure sequence from job `77143362724`:

```
GITHUB_TOKEN Permissions:
  Issues: write
  Metadata: read

@lml2468 is not an org member, proceeding
RequestError [HttpError]: Resource not accessible by integration
  status: 403
  → at search.issuesAndPullRequests
```

The org-membership check returned 404 (lml2468 not detected as member via the token), so execution fell through to `search.issuesAndPullRequests({q: "type:pr author:..."})`. That call requires **`pull-requests: read`** at minimum — which is not in the token when the job only has `issues: write`.

## Why This PR Does Not Fix the Problem

Adding `pull-requests: write` to the **caller** job won't help, because the **reusable workflow's own job** explicitly restricts permissions to `issues: write` only:

```yaml
# reusable-pr-contributor-welcome.yml (current, after .github#33)
jobs:
  welcome:
    permissions:
      issues: write   ← this is the ceiling for its steps
```

Per GitHub Actions docs: the effective GITHUB_TOKEN permissions in a reusable workflow's job = **intersection** of (caller job permissions) ∩ (reusable job permissions). The caller granting `pull-requests: write` is silently filtered out.

## Correct Fix

The fix must happen in the **reusable workflow** (`Mininglamp-OSS/.github`):

```yaml
# reusable-pr-contributor-welcome.yml
jobs:
  welcome:
    permissions:
      issues: write
      pull-requests: read   # ← needed for search.issuesAndPullRequests(type:pr)
```

`pull-requests: read` is sufficient — the workflow only reads PR count, it never writes to PRs directly.

Alternatively, replace `search.issuesAndPullRequests` with `pulls.list({state: 'all', creator: username, per_page: 2})` which uses the Issues API and may work with `issues: write` alone.

---

🔵 **Note on labeler.yml comparison**: labeler.yml has both permissions because it actively writes labels (a PR-write operation). The welcome workflow never writes to PRs, so the analogy doesn't hold.

lml2468 (commented):
[CHANGES_REQUESTED] — reviewer account matches PR author; submitting as COMMENT.

## Verdict: CHANGES_REQUESTED

Agreeing with lml2468's self-review — this PR alone won't fix the welcome workflow due to the permission intersection model.

## Root Cause Validation

The CI failure log confirms (independently verified):

```
RequestError [HttpError]: Resource not accessible by integration
  status: 403
  → search.issuesAndPullRequests
```

`search.issuesAndPullRequests({q: "type:pr author:..."})" requires `pull-requests: read` — but the **callee** (reusable workflow at .github@main after #33) restricts to `permissions: {issues: write}` only. Per GitHub Actions docs: effective GITHUB_TOKEN = caller ∩ callee. The caller granting `pull-requests: write` is filtered out because the callee doesn't request it.

## Why Both PRs Are Needed

| Layer | Current | Fix PR | What it provides |
|---|---|---|---|
| Caller (this repo) | `issues: write` | **#24** (this) | Grants `pull-requests: write` |
| Callee (.github) | `issues: write` | **.github#34** | Requests `pull-requests: write` |
| Effective | ✗ (no PR perms) | Both merged | ✅ |

This PR IS necessary — the caller must grant the permission for the callee to receive it. But it's not sufficient alone.

## Blocking

1. **Least privilege**: `pull-requests: write` → `pull-requests: read`. The welcome workflow only calls `search.issuesAndPullRequests` (read) and `issues.createComment` (covered by `issues: write`). It never writes to PRs. `.github#34` should mirror this.

## Non-blocking

- Consider noting `.github#34` dependency in the PR description
- The `labeler.yml` pattern comparison is misleading: labeler needs `pull-requests: write` because it sets labels; welcome never writes to PRs

## Summary

Code direction is correct. Fix `write` → `read` for least privilege, coordinate with `.github#34`, and this is good to merge.

PR #34, opened 2026-05-22
feat(config): add Redis TLS options for managed Redis compatibility

an9xyz:
## Summary

Closes #33.

Adds three optional fields to `DB` config in `config/config.go` so downstream services (e.g. `octo-server`) can connect to managed Redis instances that enforce in-transit encryption — AWS ElastiCache, Azure Cache for Redis, GCP Memorystore, or any self-hosted Redis behind a TLS proxy.

- `RedisTLS bool` — enable TLS (`rediss://`)
- `RedisTLSInsecureSkipVerify bool` — skip cert verification (test / self-signed)
- `RedisTLSCAFile string` — optional custom CA cert path

Corresponding viper bindings: `db.redisTLS`, `db.redisTLSInsecureSkipVerify`, `db.redisTLSCAFile`.

## Scope

- **In scope**: config surface only (struct fields + viper bindings + unit tests).
- **Out of scope**: building `*tls.Config` and wiring it into the `go-redis` client — that lives in the consumer (e.g. `octo-server` will translate these flags into `redis.Options.TLSConfig`).
- **No SDK upgrade**: `go-redis/redis v6` already supports `Options.TLSConfig`.

## Backward compatibility

All new fields default to `false` / `""`. Existing configs without these keys behave identically — no breaking changes.

## Test plan

- [x] Unit tests for default values (`TestRedisTLSDefaults`)
- [x] Unit tests for viper YAML binding (`TestRedisTLSFromViper`)
- [x] `go test -race ./config/` passes
- [x] `go build ./...` passes
- [x] **Integration verified locally** against two Redis containers (plaintext :26379 + TLS :26380 with self-signed cert). Full 6-case matrix (plaintext↔plaintext, TLS↔TLS, both mismatch directions, InsecureSkipVerify, CAFile, strict mode) all behave as expected:
  - `TLS=false` → plaintext: PONG ✓
  - `TLS=true, InsecureSkipVerify=true` → TLS: PONG ✓
  - `TLS=true, CAFile=<cert>` → TLS: PONG ✓
  - `TLS=true` (strict, no CA) → TLS: rejects untrusted cert ✓
  - Protocol mismatches in either direction fail as expected ✓

## Follow-up

Consumer-side PR (in `octo-server` or wherever the redis client is instantiated) will translate these fields into `redis.Options.TLSConfig`.

PR #35, opened 2026-05-22
feat(config): add Redis TLS options for managed Redis compatibility

### LINKED ISSUE #33: feat(config): add Redis TLS options for AWS ElastiCache / Azure Cache compatibility

an9xyz:
## Summary

Closes #33.

Adds three optional fields to `DB` config in `config/config.go` so downstream services (e.g. `octo-server`) can connect to managed Redis instances that enforce in-transit encryption — AWS ElastiCache, Azure Cache for Redis, GCP Memorystore, or any self-hosted Redis behind a TLS proxy.

- `RedisTLS bool` — enable TLS (`rediss://`)
- `RedisTLSInsecureSkipVerify bool` — skip cert verification (test / self-signed)
- `RedisTLSCAFile string` — optional custom CA cert path

Corresponding viper bindings: `db.redisTLS`, `db.redisTLSInsecureSkipVerify`, `db.redisTLSCAFile`.

## Scope

- **In scope**: config surface only (struct fields + viper bindings + unit tests).
- **Out of scope**: building `*tls.Config` and wiring it into the `go-redis` client — that lives in the consumer (e.g. `octo-server` will translate these flags into `redis.Options.TLSConfig`).
- **No SDK upgrade**: `go-redis/redis v6` already supports `Options.TLSConfig`.

## Backward compatibility

All new fields default to `false` / `""`. Existing configs without these keys behave identically — no breaking changes.

## Test plan

- [x] Unit tests for default values (`TestRedisTLSDefaults`)
- [x] Unit tests for viper YAML binding (`TestRedisTLSFromViper`)
- [x] `go test -race ./config/` passes
- [x] `go build ./...` passes
- [x] **Integration verified locally** against two Redis containers (plaintext :26379 + TLS :26380 with self-signed cert). Full 6-case matrix (plaintext↔plaintext, TLS↔TLS, both mismatch directions, InsecureSkipVerify, CAFile, strict mode) all behave as expected:
  - `TLS=false` → plaintext: PONG ✓
  - `TLS=true, InsecureSkipVerify=true` → TLS: PONG ✓
  - `TLS=true, CAFile=<cert>` → TLS: PONG ✓
  - `TLS=true` (strict, no CA) → TLS: rejects untrusted cert ✓
  - Protocol mismatches in either direction fail as expected ✓

## Follow-up

Consumer-side PR (in `octo-server` or wherever the redis client is instantiated) will translate these fields into `redis.Options.TLSConfig`.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -97,14 +97,17 @@ type Config struct {
 	}
 	// ---------- db相关配置 ----------
 	DB struct {
-		MySQLAddr            string        // mysql的连接信息
-		MySQLMaxOpenConns    int           // 最大连接数
-		MySQLMaxIdleConns    int           // 最大空闲连接数
-		MySQLConnMaxLifetime time.Duration // 连接最大生命周期
-		Migration            bool          // 是否合并数据库
-		RedisAddr            string        // redis地址
-		RedisPass            string        // redis密码
-		AsynctaskRedisAddr   string        // 异步任务的redis地址 不写默认为RedisAddr的地址
+		MySQLAddr                  string        // mysql的连接信息
+		MySQLMaxOpenConns          int           // 最大连接数
+		MySQLMaxIdleConns          int           // 最大空闲连接数
+		MySQLConnMaxLifetime       time.Duration // 连接最大生命周期
+		Migration                  bool          // 是否合并数据库
+		RedisAddr                  string        // redis地址
+		RedisPass                  string        // redis密码
+		RedisTLS                   bool          // 是否启用 TLS 连接 Redis（rediss://），用于 AWS ElastiCache / Azure Cache 等托管服务
+		RedisTLSInsecureSkipVerify bool          // 是否跳过 TLS 证书校验（仅用于测试 / 自签名证书场景）
+		RedisTLSCAFile             string        // 自定义 CA 证书文件路径（可选）
+		AsynctaskRedisAddr         string        // 异步任务的redis地址 不写默认为RedisAddr的地址
 	}
 	// ---------- 分布式配置 ----------
 	Cluster struct {
@@ -313,14 +316,17 @@ func New() *Config {
 
 		// ---------- db配置 ----------
 		DB: struct {
-			MySQLAddr            string
-			MySQLMaxOpenConns    int
-			MySQLMaxIdleConns    int
-			MySQLConnMaxLifetime time.Duration
-			Migration            bool
-			RedisAddr            string
-			RedisPass            string
-			AsynctaskRedisAddr   string
+			MySQLAddr                  string
+			MySQLMaxOpenConns          int
+			MySQLMaxIdleConns          int
+			MySQLConnMaxLifetime       time.Duration
+			Migration                  bool
+			RedisAddr                  string
+			RedisPass                  string
+			RedisTLS                   bool
+			RedisTLSInsecureSkipVerify bool
+			RedisTLSCAFile             string
+			AsynctaskRedisAddr         string
 		}{
 			MySQLAddr:            "root:demo@tcp(127.0.0.1:3306)/test?charset=utf8mb4&parseTime=true",
 			MySQLMaxOpenConns:    100,
@@ -574,6 +580,9 @@ func (c *Config) ConfigureWithViper(vp *viper.Viper) {
 	c.DB.Migration = c.getBool("db.migration", c.DB.Migration)
 	c.DB.RedisAddr = c.getString("db.redisAddr", c.DB.RedisAddr)
 	c.DB.RedisPass = c.getString("db.redisPass", c.DB.RedisPass)
+	c.DB.RedisTLS = c.getBool("db.redisTLS", c.DB.RedisTLS)
+	c.DB.RedisTLSInsecureSkipVerify = c.getBool("db.redisTLSInsecureSkipVerify", c.DB.RedisTLSInsecureSkipVerify)
+	c.DB.RedisTLSCAFile = c.getString("db.redisTLSCAFile", c.DB.RedisTLSCAFile)
 	c.DB.AsynctaskRedisAddr = c.getString("db.asynctaskRedisAddr", c.DB.AsynctaskRedisAddr)
 
 	//#################### cluster ####################
--- a/config/config_redis_tls_test.go
+++ b/config/config_redis_tls_test.go
@@ -1,0 +1,51 @@
+package config
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/spf13/viper"
+)
+
+func TestRedisTLSDefaults(t *testing.T) {
+	cfg := New()
+	vp := viper.New()
+	cfg.ConfigureWithViper(vp)
+
+	if cfg.DB.RedisTLS {
+		t.Errorf("RedisTLS default should be false, got true")
+	}
+	if cfg.DB.RedisTLSInsecureSkipVerify {
+		t.Errorf("RedisTLSInsecureSkipVerify default should be false, got true")
+	}
+	if cfg.DB.RedisTLSCAFile != "" {
+		t.Errorf("RedisTLSCAFile default should be empty, got %q", cfg.DB.RedisTLSCAFile)
+	}
+}
+
+func TestRedisTLSFromViper(t *testing.T) {
+	const yaml = `
+db:
+  redisTLS: true
+  redisTLSInsecureSkipVerify: true
+  redisTLSCAFile: /etc/ssl/certs/ca.pem
+`
+	vp := viper.New()
+	vp.SetConfigType("yaml")
+	if err := vp.ReadConfig(strings.NewReader(yaml)); err != nil {
+		t.Fatalf("read config: %v", err)
+	}
+
+	cfg := New()
+	cfg.ConfigureWithViper(vp)
+
+	if !cfg.DB.RedisTLS {
+		t.Errorf("expected RedisTLS=true")
+	}
+	if !cfg.DB.RedisTLSInsecureSkipVerify {
+		t.Errorf("expected RedisTLSInsecureSkipVerify=true")
+	}
+	if got, want := cfg.DB.RedisTLSCAFile, "/etc/ssl/certs/ca.pem"; got != want {
+		t.Errorf("RedisTLSCAFile = %q, want %q", got, want)
+	}
+}


Jerry-Xin (approved):
This PR is relevant to `octo-lib` and cleanly extends the shared config surface for Redis TLS without changing runtime behavior by itself.

💬 Non-blocking
- 🟡 Warning: [config/config.go](/private/tmp/pr-review-RDeM94/repo/config/config.go:99) uses an exported anonymous nested struct for `Config.DB`. Adding fields is backward-compatible for config files and `New()`-based usage, but it can break downstream Go code that constructs `config.Config{DB: struct{...}{...}}` literals. This is an existing API-shape limitation, not introduced uniquely by the TLS fields, but the PR’s backward-compatibility note should ideally be scoped to config-file/runtime defaults.
- 🔵 Suggestion: [config/config.go](/private/tmp/pr-review-RDeM94/repo/config/config.go:107) mentions `rediss://` in the comment, but this PR only exposes a boolean flag and does not parse or emit Redis URLs. Consider wording it as “enable TLS for Redis connections” to avoid implying URL scheme handling.

✅ Highlights
- The new fields are initialized with zero-value-safe defaults in `New()` and are bound through `ConfigureWithViper`.
- Tests cover both default behavior and YAML/Viper loading.
- Verified locally: `go test -race ./config/` and `go build ./...` both pass.

lml2468 (commented):
## Code Review — PR #35 (齐静春 📖)

### Summary

This PR replaces the now-closed #34 (identical diff, cleaner branch name `feat/config-redis-tls`). Adds three optional TLS config fields to the `DB` struct with Viper bindings and unit tests. Config-only scope — no runtime behavioral change.

### Verification

- ✅ Diff identical to reviewed #34 — same code, confirmed via `diff`
- ✅ CI: Build ✅ Test ✅ Vet ✅ (CodeQL pending, non-blocking)
- ✅ Defaults: `false`/`false`/`""` — zero behavioral change for existing configs
- ✅ Viper keys consistent with existing naming (`db.redisTLS`, etc.)
- ✅ Tests cover default values and YAML binding round-trip
- ✅ Mergeable, no conflicts

### Findings

Carried forward from my #34 review — all still apply since the code is identical:

**🟡 P1 — Follow-up: octo-lib's own Redis clients ignore TLS fields** (Non-blocking)

`Context.Cache()` → `NewRedisCache(addr, pass)` → `redis.New(addr, pass)` and `NewAsyncTask()` → `redisbroker.NewGR()` are the only two Redis client construction sites in octo-lib. Neither reads the new TLS config. Consumers using these convenience APIs won't get TLS even with `db.redisTLS=true`.

Acceptable for a first step (PR explicitly scopes this out), but the follow-up should update `redis.New()` to accept TLS options — otherwise `Context.Cache()` becomes a leaky abstraction. Suggest tracking as a GitHub issue.

**🔵 P2 — Consumer-side validation** (Non-blocking)

Consumers should log a warning when `InsecureSkipVerify=true` and validate `RedisTLSCAFile` readability before connection.

### Verdict

**LGTM** — clean, well-scoped, backward-compatible. Would APPROVE but cannot on same-account PR.

yujiawei (approved):
# Code Review — PR #35 (octo-lib)

## Summary

Adds three optional config fields (`RedisTLS`, `RedisTLSInsecureSkipVerify`, `RedisTLSCAFile`) to `config.Config.DB` plus matching viper bindings, to enable downstream services to connect to managed Redis offerings (AWS ElastiCache, Azure Cache, GCP Memorystore) that enforce in-transit encryption. Scope is explicitly limited to the config surface; building `*tls.Config` and wiring it into the `go-redis` client is left to the consumer.

The change is small (config struct + bindings + 2 unit tests), backward compatible (zero-value defaults), matches the existing patterns in this file, and links to the design discussion in #33. Tests pass locally (`go test -race ./config/`) and the package builds.

## 1. Verification

| Item | Status | Evidence |
|---|---|---|
| Struct fields added in both type decl and the `New()` literal | ✅ | `config/config.go:105-107` (decl), `config/config.go:326-328` (literal) |
| Field names/types match between decl and literal | ✅ | identical 3-field block in both places |
| viper bindings use the right helpers (`getBool` / `getString`) | ✅ | `config/config.go:583-585` |
| viper key naming consistent with neighbours (`db.redisAddr`, `db.redisPass`) | ✅ | `db.redisTLS` / `db.redisTLSInsecureSkipVerify` / `db.redisTLSCAFile` (camelCase) |
| Default values backward compatible | ✅ | `false` / `false` / `""` — pre-existing configs behave identically |
| Tests cover defaults | ✅ | `config/config_redis_tls_test.go:10-23` |
| Tests cover viper YAML binding | ✅ | `config/config_redis_tls_test.go:25-50` |
| `go test -race ./config/` | ✅ | passes |
| `go build ./...` | ✅ | passes |
| Issue #33 acceptance criteria | ✅ | requested fields and viper keys match the proposal verbatim |

## 2. Findings

No P0 / P1 issues — the change is correctness-safe and additive.

### P2 — `octo-lib`'s own internal Redis instantiation does not honor the new TLS flags

`octo-lib` itself instantiates a Redis client in two places that ignore the new fields:

- `config/context.go:105` — `Context.NewRedisCache()` calls `common.NewRedisCache(c.cfg.DB.RedisAddr, c.cfg.DB.RedisPass)` with no TLS option.
- `config/asynctask.go:38-39` — `redisbroker.NewGR` / `redisbackend.NewGR` are constructed with `cfg.DB.AsynctaskRedisAddr` only.

If a downstream service uses `Context.NewRedisCache()` (or `asynctask`) and sets `db.redisTLS: true` in its config, it will silently get a plaintext connection and the connection will fail (or worse, succeed if a plaintext port is also exposed) — without any indication that the TLS flag was inert.

The PR description correctly identifies the consumer-side wiring as out of scope, but the consumer-side work covers more than just `octo-server`: `octo-lib`'s own `Context.NewRedisCache()` and `asynctask` constructors are also consumers and will need to be updated in the follow-up. Worth tracking as a separate issue so the gap is not forgotten.

Suggested follow-up: extend `common.NewRedisCache` and the asynctask broker construction to accept TLS options derived from `cfg.DB.RedisTLS*`, or document clearly that those internal helpers do not yet support TLS.

### P2 (nit) — No `ServerName` / mTLS knobs

The current set covers the common managed-Redis cases (AWS ElastiCache, Azure, GCP). Two knobs that are sometimes needed but not exposed:

- `RedisTLSServerName` — for SNI override, useful when connecting through a stunnel proxy where the server cert SAN differs from the dial hostname.
- `RedisTLSCertFile` / `RedisTLSKeyFile` — for client mTLS, used by some self-hosted setups.

Neither is required by the linked use case (in-transit encryption to managed services) so leaving them out for now is reasonable. Worth noting in #33 as a possible future extension.

### P2 (nit) — `InsecureSkipVerify` documentation

The struct comment says "仅用于测试 / 自签名证书场景". This is correct, but a self-signed cert is better handled via `RedisTLSCAFile` than via `InsecureSkipVerify` (which disables hostname verification too). A short note in either the struct comment or the README cautioning that `InsecureSkipVerify=true` is **not** the recommended way to use a private CA would help operators pick the right knob. Strictly cosmetic.

### P2 (nit) — Test could exercise viper key case-insensitivity

`viper` is case-insensitive by default, so `db.redisTLS` and `db.redistls` resolve identically. The test currently only exercises the canonical camelCase form. A single extra subtest with snake_case or all-lowercase keys would lock the contract in. Not blocking — just a robustness nice-to-have.

## 3. Recommendations

- **Merge as-is.** The change is small, scoped, and well-tested, and the open questions all live in the explicitly-deferred consumer-side work.
- **Open a tracking issue** for the internal `octo-lib` callers (`Context.NewRedisCache`, `asynctask`) so the silent-plaintext-fallback gap is owned and won't surprise an operator who flips `db.redisTLS: true` and expects a TLS connection from the lib's helpers.
- **Consider a README/CHANGELOG note** describing the new keys and the "config-only" boundary, so consumers know they are responsible for the wiring.

## 4. Additional observations

- The new fields are placed between `RedisPass` and `AsynctaskRedisAddr`, which preserves the existing field ordering and reads naturally. Good call.
- The Go compiler will reject any future drift between the type declaration and the `New()` literal, since they are anonymous structs with the same shape — so this section is self-policing for future field additions.
- No security concerns introduced by this PR itself: defaulting `InsecureSkipVerify` to `false` is the right choice (operators must opt in to skipping verification).

PR #37, opened 2026-05-22
feat(redis): wire RedisTLS config into RedisCache / pkg/redis

### LINKED ISSUE #36: feat: wire RedisTLS config into RedisCache / pkg/redis (follow-up to #35)
## Background

PR #35 added three new config fields to `DB`:

- `RedisTLS bool`
- `RedisTLSInsecureSkipVerify bool`
- `RedisTLSCAFile string`

However, these fields are **not currently read by any code path inside `octo-lib`**. The dominant Redis access pattern across downstream services (e.g. `octo-server` uses `ctx.GetRedisConn()` in ~188 call sites) goes through:

```
Context.GetRedisConn() → Context.NewRedisCache() → common.NewRedisCache(addr, password)
                                                  → pkg/redis.New(addr, password)
                                                  → redis.NewClient(&redis.Options{Addr, Password, MaxRetries})  // no TLSConfig
```

Neither `common.NewRedisCache` nor `pkg/redis.New` accepts TLS options, so adding `RedisTLS=true` to the config has **no effect** today. The dominant Redis traffic remains plaintext, which blocks migration to managed services that require in-transit encryption (AWS ElastiCache, Azure Cache for Redis, etc.).

## Proposed change

Wire the three TLS fields through to the underlying `redis.NewClient`. Suggested shape (back-compat preserved):

### 1. `pkg/redis/redis.go`

Add a new constructor that accepts a full `*rd.Options`, keep `New(addr, password)` unchanged:

```go
// NewWithOptions builds a Conn using a caller-supplied *redis.Options.
// Useful when callers need to configure TLSConfig, PoolSize, timeouts, etc.
func NewWithOptions(opts *rd.Options) *Conn {
    if opts.MaxRetries == 0 {
        opts.MaxRetries = 3
    }
    return &Conn{client: rd.NewClient(opts)}
}
```

Optionally add a small TLS helper (no `config` dependency, so it stays usable from `common`):

```go
// BuildTLSConfig constructs a *tls.Config suitable for go-redis from
// caller-supplied flags. Returns an error if the CA file cannot be loaded.
func BuildTLSConfig(insecureSkipVerify bool, caFile string) (*tls.Config, error) { /* ... */ }
```

### 2. `common/cache.go`

Add an additional constructor that accepts a `*redis.Conn` (no import cycle, no break for existing callers):

```go
// NewRedisCacheWithConn wraps a caller-built *redis.Conn into a RedisCache.
func NewRedisCacheWithConn(conn *redis.Conn) *RedisCache {
    return &RedisCache{conn: conn}
}
```

### 3. `config/context.go`

Inside `Context.NewRedisCache()`, build options from `cfg.DB.Redis*` (including TLS), then construct the cache via the new helpers:

```go
func (c *Context) NewRedisCache() *common.RedisCache {
    c.redisOnce.Do(func() {
        opts := &rd.Options{
            Addr:     c.cfg.DB.RedisAddr,
            Password: c.cfg.DB.RedisPass,
        }
        if c.cfg.DB.RedisTLS {
            tlsCfg, err := redis.BuildTLSConfig(
                c.cfg.DB.RedisTLSInsecureSkipVerify,
                c.cfg.DB.RedisTLSCAFile,
            )
            if err != nil {
                panic(fmt.Errorf("octo-lib: build redis tls: %w", err))
            }
            opts.TLSConfig = tlsCfg
        }
        c.redisCache = common.NewRedisCacheWithConn(redis.NewWithOptions(opts))
    })
    return c.redisCache
}
```

## Backward compatibility

- `pkg/redis.New(addr, password)` and `common.NewRedisCache(addr, password)` remain unchanged.
- Behavior is identical when `RedisTLS=false` (the default).
- Only opt-in (`RedisTLS: true` in config) triggers the new TLS code path.

## Out of scope

- No `go-redis` SDK upgrade. `v6.15.x` already supports `Options.TLSConfig`.
- `AsynctaskRedisAddr` does not get a separate TLS knob in this iteration — falls back to `RedisTLS` if/when it shares the same Redis instance. A future issue can split if needed.

## Test plan

- Unit test in `pkg/redis` for `BuildTLSConfig` (default values, CA load failure, valid CA).
- Integration test (optional) booting a local TLS Redis container and validating `PING` over TLS using `RedisTLS=true` + `RedisTLSInsecureSkipVerify=true`.

## Follow-up to

#33, #35

I can open the PR if a maintainer can confirm the proposed shape.

an9xyz:
## Summary

Closes #36. Follow-up to #33 / #35.

PR #35 added the TLS config fields but no code path inside `octo-lib` read them. The dominant Redis access pattern across downstream services (~188 call sites in `octo-server` use `ctx.GetRedisConn()`) routes through:

```
Context.GetRedisConn() → Context.NewRedisCache() → common.NewRedisCache(addr, password)
                                                  → pkg/redis.New(addr, password)
                                                  → redis.NewClient(&redis.Options{...}) // no TLSConfig
```

This PR wires the three TLS fields end-to-end so `db.redisTLS: true` actually produces a TLS connection — without changing any existing constructor signatures.

### `pkg/redis/redis.go`

- `BuildTLSConfig(insecureSkipVerify bool, caFile string) (*tls.Config, error)` — builds a `*tls.Config` from caller-supplied flags. Loads optional CA file (`os.ReadFile` + `AppendCertsFromPEM`), returns descriptive errors on missing/invalid PEM. Lives in `pkg/redis` to avoid a `common → config` import cycle.
- `NewWithOptions(opts *rd.Options) *Conn` — accepts a fully-populated `*rd.Options` and applies the same `MaxRetries=3` default as `New`.

### `common/cache.go`

- `NewRedisCacheWithConn(conn *redis.Conn) *RedisCache` — wraps a pre-built connection. Avoids forcing `common` to depend on `config` or to construct `*rd.Options` directly.

### `config/context.go`

`Context.NewRedisCache()` now builds `*rd.Options` from `cfg.DB.Redis*`. When `RedisTLS=true`, it calls `redis.BuildTLSConfig` and attaches the result to `opts.TLSConfig`. CA load failure panics with a wrapped error (matches existing `NewContext`/`getIntranetIP` panic style for misconfig).

## Backward compatibility

- `pkg/redis.New(addr, password)` — **unchanged**.
- `common.NewRedisCache(addr, password)` — **unchanged**.
- Default (`RedisTLS=false`) is byte-equivalent to the old hardcoded `*rd.Options`: same `Addr`, `Password`, `MaxRetries=3`, no TLS.
- Only `RedisTLS=true` opts into the new TLS path.

## Out of scope

- `AsynctaskRedisAddr` keeps its plaintext-only machinery construction (`config/asynctask.go`). The issue explicitly defers this; worth a separate ticket if/when consumers need TLS for async-task Redis.
- No `RedisTLSServerName` (SNI override) or `RedisTLSCertFile/KeyFile` (mTLS). Not needed for the target use case (AWS ElastiCache / Azure Cache / GCP Memorystore in-transit encryption); easy to add later behind the same `BuildTLSConfig` helper.
- No `go-redis` SDK upgrade required (`v6.15` already supports `Options.TLSConfig`).

## Test plan

- [x] `pkg/redis/tls_test.go` — 8 unit tests covering:
  - `BuildTLSConfig` with no CA, `InsecureSkipVerify`, missing CA file, invalid PEM, valid self-signed CA (generated at runtime via `ecdsa.GenerateKey` + `x509.CreateCertificate`, no fixture files needed).
  - `NewWithOptions` default `MaxRetries`, caller-supplied `MaxRetries`, nil `*rd.Options`.
- [x] `go test -race ./pkg/redis/ ./common/ ./config/` passes.
- [x] `go build ./...` passes.
- [x] The TLS-path itself (config field → `*tls.Config` → `redis.NewClient` over TLS) was already integration-verified in PR #35 against a local TLS Redis container (PONG over `rediss://`). This PR only adds the wiring; runtime path is unchanged.

## Follow-ups (separate issues recommended)

- `AsynctaskRedisAddr` TLS support.
- `RedisTLSServerName` for SNI override (needed when connecting to a TLS proxy whose cert SAN differs from the dial host).
- Optional `RedisTLSCertFile` / `RedisTLSKeyFile` for mTLS-protected self-hosted setups.

### DIFF
--- a/common/cache.go
+++ b/common/cache.go
@@ -19,6 +19,13 @@ func NewRedisCache(addr string, password string) *RedisCache {
 	return r
 }
 
+// NewRedisCacheWithConn wraps a caller-built *redis.Conn into a RedisCache.
+// Use this when the connection needs options beyond addr/password
+// (e.g. TLSConfig built via redis.BuildTLSConfig).
+func NewRedisCacheWithConn(conn *redis.Conn) *RedisCache {
+	return &RedisCache{conn: conn}
+}
+
 // Set Set
 func (r *RedisCache) Set(key string, value string) error {
 	return r.conn.Set(key, value)
--- a/common/cache.go
+++ b/common/cache.go
@@ -21,8 +21,12 @@ func NewRedisCache(addr string, password string) *RedisCache {
 
 // NewRedisCacheWithConn wraps a caller-built *redis.Conn into a RedisCache.
 // Use this when the connection needs options beyond addr/password
-// (e.g. TLSConfig built via redis.BuildTLSConfig).
+// (e.g. TLSConfig built via redis.BuildTLSConfig). Panics if conn is nil
+// to surface the misuse immediately rather than at first cache operation.
 func NewRedisCacheWithConn(conn *redis.Conn) *RedisCache {
+	if conn == nil {
+		panic("common.NewRedisCacheWithConn: nil *redis.Conn")
+	}
 	return &RedisCache{conn: conn}
 }
 
--- a/config/context.go
+++ b/config/context.go
@@ -1,10 +1,10 @@
 package config
 
 import (
+	"fmt"
 	"sync"
 	"time"
 
-	"github.com/RussellLuo/timingwheel"
 	"github.com/Mininglamp-OSS/octo-lib/common"
 	"github.com/Mininglamp-OSS/octo-lib/pkg/cache"
 	"github.com/Mininglamp-OSS/octo-lib/pkg/db"
@@ -13,7 +13,9 @@ import (
 	"github.com/Mininglamp-OSS/octo-lib/pkg/redis"
 	"github.com/Mininglamp-OSS/octo-lib/pkg/wkevent"
 	"github.com/Mininglamp-OSS/octo-lib/pkg/wkhttp"
+	"github.com/RussellLuo/timingwheel"
 	"github.com/bwmarrin/snowflake"
+	rd "github.com/go-redis/redis"
 	"github.com/gocraft/dbr/v2"
 	"github.com/olivere/elastic"
 	"github.com/opentracing/opentracing-go"
@@ -102,7 +104,21 @@ func (c *Context) DB() *dbr.Session {
 // NewRedisCache 创建一个redis缓存
 func (c *Context) NewRedisCache() *common.RedisCache {
 	c.redisOnce.Do(func() {
-		c.redisCache = common.NewRedisCache(c.cfg.DB.RedisAddr, c.cfg.DB.RedisPass)
+		opts := &rd.Options{
+			Addr:     c.cfg.DB.RedisAddr,
+			Password: c.cfg.DB.RedisPass,
+		}
+		if c.cfg.DB.RedisTLS {
+			tlsCfg, err := redis.BuildTLSConfig(
+				c.cfg.DB.RedisTLSInsecureSkipVerify,
+				c.cfg.DB.RedisTLSCAFile,
+			)
+			if err != nil {
+				panic(fmt.Errorf("octo-lib: build redis TLS config: %w", err))
+			}
+			opts.TLSConfig = tlsCfg
+		}
+		c.redisCache = common.NewRedisCacheWithConn(redis.NewWithOptions(opts))
 	})
 	return c.redisCache
 }
--- a/config/context_redis_test.go
+++ b/config/context_redis_test.go
@@ -1,0 +1,93 @@
+package config
+
+import (
+	"strings"
+	"sync"
+	"testing"
+
+	"github.com/spf13/viper"
+)
+
+func loadCfgWithDB(t *testing.T, yaml string) *Config {
+	t.Helper()
+	vp := viper.New()
+	vp.SetConfigType("yaml")
+	if err := vp.ReadConfig(strings.NewReader(yaml)); err != nil {
+		t.Fatalf("read config: %v", err)
+	}
+	cfg := New()
+	cfg.ConfigureWithViper(vp)
+	return cfg
+}
+
+// minimalContext builds a Context wired only for the bits NewRedisCache needs,
+// skipping the heavyweight NewContext (event pools, tracer, snowflake).
+func minimalContext(cfg *Config) *Context {
+	return &Context{
+		cfg:       cfg,
+		redisOnce: sync.Once{},
+	}
+}
+
+// Wiring path: RedisTLS=false produces a Conn whose Options carry no TLSConfig.
+func TestNewRedisCache_PlainConfig_NoTLSConfig(t *testing.T) {
+	cfg := loadCfgWithDB(t, `
+db:
+  redisAddr: 127.0.0.1:65535
+`)
+	c := minimalContext(cfg)
+	cache := c.NewRedisCache()
+	if cache == nil {
+		t.Fatal("expected non-nil cache")
+	}
+	if cache.GetRedisConn().Options().TLSConfig != nil {
+		t.Errorf("expected TLSConfig=nil when RedisTLS=false")
+	}
+}
+
+// Wiring path: RedisTLS=true attaches the TLSConfig produced by BuildTLSConfig.
+func TestNewRedisCache_TLSConfig_AttachesTLSConfig(t *testing.T) {
+	cfg := loadCfgWithDB(t, `
+db:
+  redisAddr: 127.0.0.1:65535
+  redisTLS: true
+  redisTLSInsecureSkipVerify: true
+`)
+	c := minimalContext(cfg)
+	cache := c.NewRedisCache()
+	opts := cache.GetRedisConn().Options()
+	if opts.TLSConfig == nil {
+		t.Fatal("expected non-nil TLSConfig when RedisTLS=true")
+	}
+	if !opts.TLSConfig.InsecureSkipVerify {
+		t.Errorf("expected InsecureSkipVerify=true propagated to TLSConfig")
+	}
+}
+
+// Wiring path: a bad CA file panics with a wrapped, contextful message.
+func TestNewRedisCache_TLSConfig_BadCAFile_Panics(t *testing.T) {
+	cfg := loadCfgWithDB(t, `
+db:
+  redisAddr: 127.0.0.1:65535
+  redisTLS: true
+  redisTLSCAFile: /definitely/does/not/exist.pem
+`)
+	c := minimalContext(cfg)
+	defer func() {
+		r := recover()
+		if r == nil {
+			t.Fatal("expected panic on missing CA file")
+		}
+		msg, ok := r.(error)
+		if !ok {
+			t.Fatalf("expected panic value to be error, got %T: %v", r, r)
+		}
+		if !strings.Contains(msg.Error(), "build redis TLS config") {
+			t.Errorf("panic missing wrapping context: %v", msg)
+		}
+		if !strings.Contains(msg.Error(), "does/not/exist.pem") {
+			t.Errorf("panic missing CA file path: %v", msg)
+		}
+	}()
+	_ = c.NewRedisCache()
+}
--- a/pkg/redis/redis.go
+++ b/pkg/redis/redis.go
@@ -32,22 +32,36 @@ func New(addr string, password string) *Conn {
 
 // NewWithOptions builds a Conn using caller-supplied *rd.Options. Useful when
 // callers need to configure TLSConfig, PoolSize, custom timeouts, etc.
-// MaxRetries defaults to 3 if unset, matching the behavior of New.
+// The caller's *rd.Options is not mutated — a shallow copy is taken before any
+// defaults are applied. MaxRetries==0 is treated as "unset" and replaced with
+// 3 to match the behavior of New; callers that need to disable retries should
+// use the underlying go-redis client directly.
 func NewWithOptions(opts *rd.Options) *Conn {
-	if opts == nil {
-		opts = &rd.Options{}
+	var local rd.Options
+	if opts != nil {
+		local = *opts
 	}
-	if opts.MaxRetries == 0 {
-		opts.MaxRetries = 3
+	if local.MaxRetries == 0 {
+		local.MaxRetries = 3
 	}
-	return &Conn{client: rd.NewClient(opts)}
+	return &Conn{client: rd.NewClient(&local)}
+}
+
+// Options returns the underlying go-redis client options. Intended for tests
+// and diagnostics; do not mutate the returned value.
+func (rc *Conn) Options() *rd.Options {
+	return rc.client.Options()
 }
 
 // BuildTLSConfig constructs a *tls.Config suitable for go-redis from caller
 // supplied flags. If caFile is empty, the system root pool is used. Returns
 // an error if the CA file cannot be read or contains no valid PEM certs.
+// MinVersion is pinned to TLS 1.2 to make the security posture explicit.
 func BuildTLSConfig(insecureSkipVerify bool, caFile string) (*tls.Config, error) {
-	cfg := &tls.Config{InsecureSkipVerify: insecureSkipVerify}
+	cfg := &tls.Config{
+		InsecureSkipVerify: insecureSkipVerify,
+		MinVersion:         tls.VersionTLS12,
+	}
 	if caFile == "" {
 		return cfg, nil
 	}
--- a/pkg/redis/redis.go
+++ b/pkg/redis/redis.go
@@ -1,7 +1,11 @@
 package redis
 
 import (
+	"crypto/tls"
+	"crypto/x509"
 	"errors"
+	"fmt"
+	"os"
 	"time"
 
 	rd "github.com/go-redis/redis"
@@ -24,6 +28,39 @@ func New(addr string, password string) *Conn {
 		Password:   password,
 	})
 	return c
+}
+
+// NewWithOptions builds a Conn using caller-supplied *rd.Options. Useful when
+// callers need to configure TLSConfig, PoolSize, custom timeouts, etc.
+// MaxRetries defaults to 3 if unset, matching the behavior of New.
+func NewWithOptions(opts *rd.Options) *Conn {
+	if opts == nil {
+		opts = &rd.Options{}
+	}
+	if opts.MaxRetries == 0 {
+		opts.MaxRetries = 3
+	}
+	return &Conn{client: rd.NewClient(opts)}
+}
+
+// BuildTLSConfig constructs a *tls.Config suitable for go-redis from caller
+// supplied flags. If caFile is empty, the system root pool is used. Returns
+// an error if the CA file cannot be read or contains no valid PEM certs.
+func BuildTLSConfig(insecureSkipVerify bool, caFile string) (*tls.Config, error) {
+	cfg := &tls.Config{InsecureSkipVerify: insecureSkipVerify}
+	if caFile == "" {
+		return cfg, nil
+	}
+	caPEM, err := os.ReadFile(caFile)
+	if err != nil {
+		return nil, fmt.Errorf("read redis CA file %q: %w", caFile, err)
+	}
+	pool := x509.NewCertPool()
+	if !pool.AppendCertsFromPEM(caPEM) {
+		return nil, fmt.Errorf("redis CA file %q contains no valid PEM certificates", caFile)
+	}
+	cfg.RootCAs = pool
+	return cfg, nil
 }
 
 // Close closes the underlying Redis client and releases resources.
--- a/pkg/redis/tls_test.go
+++ b/pkg/redis/tls_test.go
@@ -4,6 +4,7 @@ import (
 	"crypto/ecdsa"
 	"crypto/elliptic"
 	"crypto/rand"
+	"crypto/tls"
 	"crypto/x509"
 	"crypto/x509/pkix"
 	"encoding/pem"
@@ -62,6 +63,9 @@ func TestBuildTLSConfig_NoCA(t *testing.T) {
 	}
 	if cfg.InsecureSkipVerify {
 		t.Errorf("expected InsecureSkipVerify=false")
+	}
+	if cfg.MinVersion != tls.VersionTLS12 {
+		t.Errorf("MinVersion = %x, want TLS 1.2 (%x)", cfg.MinVersion, tls.VersionTLS12)
 	}
 }
 
@@ -125,5 +129,17 @@ func TestNewWithOptions_NilOpts(t *testing.T) {
 	conn := NewWithOptions(nil)
 	if conn == nil || conn.client == nil {
 		t.Fatal("expected non-nil conn and client even with nil opts")
+	}
+	if got := conn.Options().MaxRetries; got != 3 {
+		t.Errorf("MaxRetries = %d, want 3 (default applied on nil opts)", got)
+	}
+}
+
+// Regression: NewWithOptions must not mutate the caller's *rd.Options.
+func TestNewWithOptions_DoesNotMutateCaller(t *testing.T) {
+	caller := &rd.Options{Addr: "127.0.0.1:1"} // MaxRetries left at zero
+	_ = NewWithOptions(caller)
+	if caller.MaxRetries != 0 {
+		t.Errorf("caller's MaxRetries was mutated to %d, want 0", caller.MaxRetries)
 	}
 }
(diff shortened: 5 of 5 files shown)


Jerry-Xin (dismissed):
Summary: The PR is in scope for `octo-lib` and correctly wires the existing Redis TLS config into the primary `Context.GetRedisConn()` path without breaking existing constructors.

💬 Non-blocking
- 🟡 Warning: [pkg/redis/redis.go](/private/tmp/pr-review-QDu3To/repo/pkg/redis/redis.go:36) `NewWithOptions` mutates the caller’s `*rd.Options` and treats `MaxRetries == 0` as unset, so callers cannot explicitly disable retries through this public API. This matches current `New` behavior, but the comment says caller-supplied options, so documenting this more explicitly would avoid surprise.
- 🔵 Suggestion: [config/context.go](/private/tmp/pr-review-QDu3To/repo/config/context.go:105) The main wiring path is not directly unit-tested. The helper tests are good, but a small config-level test for TLS CA failure panic or successful TLS option construction would better protect the actual regression fixed by this PR.
- 🔵 Suggestion: [common/cache.go](/private/tmp/pr-review-QDu3To/repo/common/cache.go:25) Since `NewRedisCacheWithConn` is public, consider whether a nil guard is worthwhile. Passing nil currently creates a cache that will panic later on use.

✅ Highlights
- The relevance gate passes: Redis cache/config wiring is directly part of this repository’s functionality.
- Backward-compatible constructors remain unchanged.
- `BuildTLSConfig` returns useful errors for missing/invalid CA files.
- Local verification passed: `go test ./pkg/redis ./common ./config`, `go test -race ./pkg/redis ./common ./config`, `go build ./...`, and `git diff --check`.

yujiawei (dismissed):
# Code Review — PR #37 (octo-lib)

## Summary

Wires the Redis TLS config fields added in #35 through `Context.NewRedisCache()` so `db.redisTLS: true` actually produces a TLS-enabled `redis.Client`. Two small additive helpers are introduced — `pkg/redis.NewWithOptions` and `common.NewRedisCacheWithConn` — leaving the existing `redis.New(addr, password)` and `common.NewRedisCache(addr, password)` constructors untouched. Backward compatibility is preserved end-to-end: when `RedisTLS=false`, the resulting `*rd.Options` is byte-equivalent to the old path (same `Addr`, `Password`, `MaxRetries=3`, no `TLSConfig`).

## 1. Verification

| Check | Status | Evidence |
|---|---|---|
| Builds clean | ✅ | `go build ./...` — no output |
| `go vet ./...` | ✅ | clean |
| Tests on changed packages | ✅ | `go test ./pkg/redis/ ./common/ ./config/` — all pass |
| Backward compat: `pkg/redis.New` signature/behavior | ✅ | `pkg/redis/redis.go:23-31` unchanged |
| Backward compat: `common.NewRedisCache(addr, password)` | ✅ | `common/cache.go:16-20` unchanged |
| Default (`RedisTLS=false`) byte-equivalent to old path | ✅ | `config/context.go:107-110` builds `Addr+Password` only; `pkg/redis/redis.go:36-44` defaults `MaxRetries=3` — matches old `New` literal |
| TLS path actually attaches `*tls.Config` to client | ✅ | `config/context.go:111-120` → `pkg/redis/redis.go:43` (`rd.NewClient(opts)`) |
| Avoids `common → config` import cycle | ✅ | `BuildTLSConfig` lives in `pkg/redis`, `config` imports `pkg/redis`, `common` imports `pkg/redis`; no `common → config` edge |
| TLS unit tests cover happy + error paths | ✅ | `pkg/redis/tls_test.go` — 8 tests including missing CA, invalid PEM, valid self-signed CA, nil opts |

## 2. Findings

No P0 or P1 issues found. A few P2 / nit suggestions below — none block merge.

### P2 — `BuildTLSConfig` does not pin `MinVersion`

`pkg/redis/redis.go:50` constructs `&tls.Config{InsecureSkipVerify: insecureSkipVerify}` with no explicit `MinVersion`. Go 1.18+ defaults to `tls.VersionTLS12` on the client side, so this is currently safe in practice, but an explicit pin (`MinVersion: tls.VersionTLS12`) is more defensive against future Go runtime changes and makes the security posture obvious to readers.

Suggested:

```go
cfg := &tls.Config{
    InsecureSkipVerify: insecureSkipVerify,
    MinVersion:         tls.VersionTLS12,
}
```

### P2 — `NewWithOptions` mutates the caller's `*rd.Options` in place

`pkg/redis/redis.go:36-44` writes `opts.MaxRetries = 3` directly on the caller-supplied pointer. For the only current call site (`config/context.go:121`, where `opts` is locally built and immediately passed in) this is harmless, but it is a mild surprise for a function that takes `*rd.Options`. A shallow copy (`o := *opts; ...; rd.NewClient(&o)`) avoids the side effect; alternatively, document the in-place mutation. Not blocking — the current single call site is safe.

### P2 — Test gap: `TestNewWithOptions_NilOpts` doesn't assert default `MaxRetries`

`pkg/redis/tls_test.go:124-129` only checks `conn != nil && conn.client != nil`. The whole point of the nil-opts branch is that the same `MaxRetries=3` default still applies; asserting `conn.client.Options().MaxRetries == 3` would lock that behavior in.

### Nit — Panic message context

`config/context.go:117` panics with `fmt.Errorf("octo-lib: build redis TLS config: %w", err)`. Matches the codebase's existing panic-on-misconfig style (`NewContext`/`getIntranetIP`), so this is consistent. If an "early validation, no panic at first-use" model is ever adopted for this package, TLS config build is a natural candidate to migrate alongside the others — out of scope for this PR.

### Nit — `redis.NewWithOptions(nil)` constructs a client with `Addr=""`

go-redis falls back to `localhost:6379` in that case. The nil path is only exercised by the test today; the production call site always supplies a populated `*rd.Options`. Worth a one-line comment if you want to document the fallback, but not a defect.

## 3. Out-of-scope items, correctly deferred

The PR description and code consistently defer:

- `AsynctaskRedisAddr` TLS — `config/asynctask.go:38-39` still uses plaintext-only `redisbroker.NewGR` / `redisbackend.NewGR`. Issue #36 explicitly defers this.
- `RedisTLSServerName` (SNI override).
- `RedisTLSCertFile` / `RedisTLSKeyFile` for mTLS.

All three are reasonable follow-ups; the surface introduced in this PR (`BuildTLSConfig(insecureSkipVerify, caFile)`) is easy to extend without breaking callers.

## 4. Additional observations (not requested by the task)

- The `sync.Once` (`config/context.go:106`) means a CA-file read failure panics on the *first* `NewRedisCache` call only; subsequent callers receive a nil `redisCache` with no re-attempt. This is fine because the panic is fatal, but worth noting if `NewRedisCache` is ever changed to return an error rather than panic — `sync.Once` would silently mask the second call's intent.
- `common/cache.go:25-27` (`NewRedisCacheWithConn`) is a thin wrapper. Trivial and correct.
- `pkg/redis/tls_test.go:21-50` self-generates a self-signed CA at runtime via `ecdsa.GenerateKey` + `x509.CreateCertificate` — clean, no fixture files, no test-data drift.

## Verdict

The change is small, additive, and correctly threaded. Backward compatibility is preserved literally byte-for-byte on the default path, and the TLS opt-in path is covered by unit tests with realistic happy + failure cases. The P2 items above are improvement suggestions, not merge blockers.

Approving.

lml2468 (commented):
## Code Review — PR #37 (齐静春 📖)

### Summary

This PR wires the TLS config fields from #35 into the actual Redis client construction path (`Context.NewRedisCache()` → `redis.NewWithOptions()` with `TLSConfig`). This is exactly the P1 follow-up I flagged in my #34/#35 reviews. Clean implementation, no existing API breakage.

### Verification

- ✅ CI: Build ✅ Test ✅ Vet ✅ CodeQL ✅
- ✅ Old constructors untouched: `redis.New(addr, pass)` and `common.NewRedisCache(addr, pass)` unchanged
- ✅ `BuildTLSConfig` validates CA file existence and PEM validity before returning
- ✅ Panic on bad CA config is consistent with existing startup-time panics (`snowflake.NewNode`, `getIntranetIP`)
- ✅ Import cycle avoided: `BuildTLSConfig` lives in `pkg/redis`, `config/context.go` imports `pkg/redis` (no circular dep)
- ✅ Tests cover: no-CA, insecure-skip-verify, missing file, invalid PEM, valid CA, nil opts, default/custom MaxRetries
- ✅ `sync.Once` protection on `NewRedisCache()` preserved

### Findings

**🟡 P1 — `AsynctaskRedisAddr` path still not TLS-wired** (Non-blocking)

`NewAsyncTask()` in `config/asynctask.go:38` constructs Redis broker/backend via `redisbroker.NewGR(cnf, []string{cfg.DB.AsynctaskRedisAddr}, 0)` — no TLS. If `AsynctaskRedisAddr` resolves to the same managed Redis instance that requires TLS, async tasks will fail to connect.

Machinery v2 (`v2.0.10`) broker/backend constructors don't directly accept `*tls.Config`. Wiring TLS here likely requires either:
- (a) Using `rediss://` URL scheme in `AsynctaskRedisAddr` if machinery supports it, or
- (b) Checking if machinery v2 has a TLS-aware constructor/config option

This is a separate concern from the main Redis path and can be a follow-up. Worth tracking as an issue so it doesn't get missed when deploying to a TLS-only Redis environment.

**🟢 P2 — `NewWithOptions` mutates caller's `*rd.Options`** (Non-blocking, already flagged by Allen)

`MaxRetries == 0` treated as "unset" means callers can't explicitly disable retries. Current usage in `context.go` constructs `opts` locally so this is safe. Could be tightened with a sentinel or by using `opts.MaxRetries` only when the caller sets it, but not blocking for this PR.

### Verdict

**LGTM** — this closes the gap cleanly. Would APPROVE but cannot on same-account PR.

an9xyz:
Pushed a follow-up commit addressing the P2 feedback. Summary:

| Reviewer note | Resolution |
|---|---|
| `NewWithOptions` mutates caller's `*rd.Options` (Jerry-Xin, yujiawei, lml2468) | Shallow copy before applying defaults. `MaxRetries==0` sentinel documented explicitly. New test `TestNewWithOptions_DoesNotMutateCaller` locks the no-mutation contract. |
| `BuildTLSConfig` should pin `MinVersion` (yujiawei) | Pinned to `tls.VersionTLS12`. New assertion in `TestBuildTLSConfig_NoCA`. |
| `TestNewWithOptions_NilOpts` should assert default `MaxRetries` (yujiawei) | Done — asserts `MaxRetries == 3` on nil opts. |
| Wiring path in `config/context.go` lacks direct unit tests (Jerry-Xin) | Added `config/context_redis_test.go` with three cases: plain config → no TLSConfig; `RedisTLS=true` → TLSConfig attached; bad CA file → panic with wrapped + path-bearing error message. Requires a small `Options()` accessor on `redis.Conn` for diagnostics. |
| `NewRedisCacheWithConn` lacks nil guard (Jerry-Xin) | Panics immediately on nil to surface the misuse at construction time rather than at first cache op. |

`AsynctaskRedisAddr` TLS (P1, lml2468) remains a deliberate follow-up — Machinery v2's broker/backend constructors don't directly accept `*tls.Config`, so wiring it needs a separate investigation. I'll open a tracking issue.

All tests still green:
- `go test -race ./pkg/redis/ ./common/ ./config/` ✅
- `go build ./...` ✅

Jerry-Xin (approved):
Summary: The PR is in scope for `octo-lib` and correctly wires existing Redis TLS config through the dominant `Context.GetRedisConn()` path without breaking existing constructors.

🔴 Blocking: None.

💬 Non-blocking:
- 🔵 Suggestion: `pkg/redis/redis.go:50` exposes `Options()` as public API for tests/diagnostics, but it returns mutable go-redis options. The comment helps, and this is not blocking, but a narrower test-only approach would avoid expanding the public surface.

✅ Highlights:
- `config/context.go:105` cleanly preserves plaintext behavior while attaching `TLSConfig` only when `db.redisTLS` is enabled.
- `pkg/redis/redis.go:60` handles missing and invalid CA files with useful errors and pins TLS minimum version to 1.2.
- `common/cache.go:26` keeps existing constructor signatures intact while allowing callers to pass a prebuilt Redis connection.
- Tests cover config parsing, TLS wiring, CA handling, and `NewWithOptions` retry/default behavior.

Verification performed:
- `gh pr diff 37 --repo Mininglamp-OSS/octo-lib`
- `go test -race ./pkg/redis/ ./common/ ./config/`
- `go test ./...`
- `go build ./...`

yujiawei (approved):
# Code Review — PR #37 (octo-lib)

## Summary

This PR completes the Redis TLS wiring that PR #35 left dangling: the three `DB.RedisTLS*` config fields are now actually consumed by `Context.NewRedisCache()`, and downstream services using the dominant `ctx.GetRedisConn()` access pattern get TLS for free when `db.redisTLS: true`. The change is small, well-tested, and preserves binary-equivalent behavior on the plaintext default path.

## Verification

| Item | Result | Evidence |
|---|---|---|
| Build clean | ✅ | `go build ./...` passes |
| `go vet` clean | ✅ | no diagnostics |
| Unit tests pass under race | ✅ | `go test -race ./pkg/redis/ ./common/ ./config/` — all green; 12 new tests verified individually |
| Config fields exist | ✅ | `config/config.go:107-109` declare `RedisTLS`, `RedisTLSInsecureSkipVerify`, `RedisTLSCAFile`; loader at `config/config.go:583-585` reads them via viper |
| Backward-compat: `pkg/redis.New(addr, password)` unchanged | ✅ | `pkg/redis/redis.go:23-31` — signature + body untouched |
| Backward-compat: `common.NewRedisCache(addr, password)` unchanged | ✅ | `common/cache.go:16-20` — signature + body untouched |
| Default path (`RedisTLS=false`) byte-equivalent to old hardcoded options | ✅ | `config/context.go:107-110` builds `{Addr, Password}`; `NewWithOptions` defaults `MaxRetries=0 → 3` (`pkg/redis/redis.go:44-46`), matching the old `New` constant |
| Import cycle avoided | ✅ | `BuildTLSConfig` lives in `pkg/redis` (no `config` import); `common` keeps its existing `pkg/redis` dep only |
| Caller's `*rd.Options` not mutated | ✅ | `pkg/redis/redis.go:40-42` shallow-copies before mutation; regression test `TestNewWithOptions_DoesNotMutateCaller` (`pkg/redis/tls_test.go:139-145`) locks this in |
| TLS security defaults explicit | ✅ | `MinVersion: tls.VersionTLS12` pinned at `pkg/redis/redis.go:63` |
| CA load failure surfaces at startup, not at first op | ✅ | `config/context.go:116-118` wraps the error and panics inside `redisOnce.Do`, matching the existing `NewContext` panic style for misconfig |
| Nil-conn misuse fails loudly | ✅ | `common/cache.go:26-30` panics in `NewRedisCacheWithConn` rather than deferring NPE to first cache op |
| `AsynctaskRedisAddr` deferral honored | ✅ | `config/asynctask.go:38-39` still plaintext; PR body explicitly defers this and follow-up issues are recommended |

## Findings

### P0 / P1

None.

### P2 — non-blocking nits / suggestions

1. **`Conn.Options()` returns the live `*rd.Options` pointer** (`pkg/redis/redis.go:50-54`). The doc warns "do not mutate," but go-redis exposes the internal pointer (no clone), so a misbehaving caller could silently re-tune the live client. The tests rely on this pointer identity, which is fine. If you want stronger encapsulation later, consider returning a narrow read-only view or copying selected fields (`Addr`, `TLSConfig != nil`, `MaxRetries`). Not blocking.

2. **`MaxRetries == 0` is silently coerced to 3 inside `NewWithOptions`** (`pkg/redis/redis.go:44-46`). This is documented and matches `New`'s behavior, but it means a caller cannot disable retries through this constructor — they must drop down to `redis.NewClient` directly. If a future caller actually wants `MaxRetries=0`, the friendlier API would be a sentinel (e.g. a negative value, or a separate `DisableRetries bool`). Worth keeping in mind, not worth blocking on now.

3. **`BuildTLSConfig` is intentionally minimal** — no `ServerName` (SNI), no `MaxVersion`, no client cert (mTLS), no custom cipher suites. The PR body lists these as explicit follow-ups, which matches the target use case (AWS ElastiCache / Azure Cache / GCP Memorystore over public CA chains). If/when self-hosted or stunnel-fronted Redis enters the picture, the `RedisTLSServerName` follow-up will be necessary because `InsecureSkipVerify=false` + mismatched SAN will fail; please cut that ticket before the first non-managed deployment.

4. **Test goroutine hygiene** — `redis.NewClient` in go-redis v6.15 starts an idle-connection reaper goroutine immediately. The new tests (`config/context_redis_test.go` and `pkg/redis/tls_test.go::TestNewWithOptions_*`) create `*Conn` instances but never call `Close()`, so each test leaks one goroutine + its associated `connPool`. The leak is benign for the test process, and no race is observed, but `t.Cleanup(func(){ _ = conn.Close() })` would be cleaner and would also exercise the `Close` path.

5. **Three-arg `BuildTLSConfig(insecureSkipVerify, caFile)` ergonomics** — if/when more knobs land (SNI, mTLS), the signature will need to grow. Consider whether the next iteration should switch to an options struct (`BuildTLSConfig(opts TLSOptions)`) to avoid a parameter-explosion refactor across callers. Pure forward-looking nit.

### Additional observations (out of task scope)

- **Asynctask Redis path is still plaintext** (`config/asynctask.go:38-39`, using `cfg.DB.AsynctaskRedisAddr`). This is consistent with what the PR body and the original issue acknowledge. If `db.redisTLS: true` lands in production while machinery/asynctask continues to dial 6379 in cleartext to the same managed cluster, that connection will fail outright on TLS-only endpoints (e.g. ElastiCache with in-transit encryption required). Worth prioritizing the follow-up ticket sooner rather than later — otherwise enabling `RedisTLS=true` will half-break the workload.
- **`Context.NewRedisCache` panics on TLS misconfig** (`config/context.go:117`). This is the right call for a startup-time wiring error, but note that it fires on the *first* `c.NewRedisCache()` invocation rather than at `NewContext` construction time (since it's gated by `redisOnce.Do`). If any service code paths can boot without ever calling `NewRedisCache`, a CA-file typo would not be caught until traffic hits the cache. Probably fine in practice (almost every handler touches Redis), but a startup probe that calls `Cache().GetRedisConn().Ping()` after `NewContext` would surface it earlier.
- **`fmt.Errorf("...: %w", err)` inside a panic** (`config/context.go:117`) — the wrapping is harmless but `%w` semantics are mainly useful for `errors.Is/As` on a returned error. For a panic value the wrapping is purely cosmetic; either is fine.

## Verdict

LGTM. No correctness, security, or backward-compat blockers. Test coverage is appropriate for the wiring (happy path + insecure-skip + bad-CA-path on both layers). The nits above are suggestions for the next iteration, not blockers for this merge.

lml2468 (approved):
Re-review after 4a24a89 (incremental from a3d59bb).

## Previous findings status

1. **P1 — AsynctaskRedisAddr broker/backend not TLS-wired** (config/asynctask.go:38-39): Still present but **out of scope** for this PR. The asynctask path uses `machinery/v2` `redisbroker.NewGR` / `redisbackend.NewGR` which take plain address strings — a separate concern from the `Context.GetRedisConn()` → `pkg/redis` wiring this PR targets. Recommend tracking as a follow-up issue.

2. **P2 — NewWithOptions mutates caller opts** (pkg/redis/redis.go:36): **Resolved.** `NewWithOptions` now takes a shallow copy (`local = *opts`) before applying defaults. `TestNewWithOptions_DoesNotMutateCaller` explicitly guards against regression.

## Incremental diff review (a3d59bb → 4a24a89)

All changes are fixes / hardening responding to round-1 feedback:

- `pkg/redis/redis.go:35-44` — shallow copy + nil guard. Clean.
- `pkg/redis/redis.go:47-49` — `Options()` accessor for tests. Godoc warns against mutation. Acceptable surface expansion.
- `pkg/redis/redis.go:55-58` — `BuildTLSConfig` now pins `MinVersion: tls.VersionTLS12`. Good hardening.
- `common/cache.go:24-28` — `NewRedisCacheWithConn` panics on nil. Fail-fast, consistent with Go stdlib patterns.
- `config/context_redis_test.go` — 3 new wiring tests (plain, TLS, bad CA file). `minimalContext` avoids `NewContext` heavyweight setup. Tests are real: they exercise `ConfigureWithViper` → `NewRedisCache` → `Options()` end-to-end.
- `pkg/redis/tls_test.go` — MinVersion assertion, MaxRetries default on nil opts, caller mutation regression test. All targeted.

CI: Build ✅ Test ✅ Vet ✅ CodeQL ✅. No new issues.

Approving.

PR #39, opened 2026-05-23
feat(config): add S3Config for AWS S3 and S3-compatible storage backends

### LINKED ISSUE #38: feat(config): add S3Config for AWS S3 and S3-compatible storage backends
## Summary

Add a first-class `S3Config` struct (and corresponding `s3.*` Viper bindings) so that `octo-server` and other downstream consumers can target AWS S3 or any S3-compatible object storage (Cloudflare R2, Backblaze B2, Wasabi, MinIO with custom endpoints, etc.) through a dedicated, clearly-named configuration block.

This is the prerequisite for a follow-up `octo-server` PR that introduces an `awsS3` value for `fileService` and refactors the existing `tencentCOS` backend into a profile of a shared S3-compatible service implementation.

## Motivation

The current `MinioConfig` (`config/config.go:1029`) and `COSConfig` (`config/config.go:1048`) cover specific deployments but cannot cleanly express a generic S3 endpoint:

- `MinioConfig` has no `BucketName` or `Region` field — `octo-server` hard-codes the bucket via path-prefix routing (`modules/file/helpers.go`) and pins `Region: us-east-1` in `service_minio.go`. Pointing it at AWS S3 in a non-`us-east-1` region requires bucket-name gymnastics (the path prefixes `chat`, `moment`, etc. must each exist as a real bucket globally) and relies on `minio-go`'s implicit `GetBucketLocation` redirect for correct signing.
- `COSConfig` has `Bucket` + `Region` but is named after Tencent Cloud COS specifically; `octo-server`'s `service_cos.go` even hard-codes the endpoint template `cos.<region>.myqcloud.com`. Reusing it for AWS would require either (a) lying about the backend type in `fileService`, or (b) leaking a Tencent-shaped config into AWS deployments — both are confusing for operators and a maintenance liability.

A neutrally-named `S3Config` makes the configuration semantically accurate and lets us consolidate the SigV4 / presigned-URL implementation in `octo-server` (which is already 100% `minio-go`/SigV4 under the hood, for both MinIO and COS) into one well-tested code path.

### New struct

```go
// S3Config configures any S3-compatible object storage backend.
// Use this for AWS S3 or third-party providers that speak the S3 protocol
// (Cloudflare R2, Backblaze B2, Wasabi, etc.). Tencent Cloud COS and MinIO
// continue to be configured via their existing dedicated blocks for
// backward compatibility.
type S3Config struct {
    // Endpoint is the S3 API hostname, without scheme.
    // Examples:
    //   AWS S3:        s3.us-west-2.amazonaws.com
    //   Cloudflare R2: <accountid>.r2.cloudflarestorage.com
    //   MinIO:         minio.example.com:9000
    // REQUIRED — no default.
    Endpoint string

    // Region is the SigV4 signing region.
    // Examples: us-west-2, ap-southeast-1, auto (R2).
    // REQUIRED — must match the bucket's actual region for AWS; some
    // providers accept any value (e.g. R2 uses \"auto\").
    Region string

    // Bucket is the single bucket all uploads target. Object keys carry
    // their own logical prefix (chat/, moment/, ...), so a single bucket
    // is the recommended deployment topology.
    Bucket string

    // AccessKeyID / SecretAccessKey are the long-lived IAM credentials
    // for the bucket. See the security docs for the minimum IAM policy.
    AccessKeyID     string
    SecretAccessKey string

    // BucketURL, when set, overrides the URL returned by DownloadURL and
    // signed against by presigned GET/PUT. Use this for CDN-fronted
    // buckets or custom domains. Must be scheme://host[:port] with no
    // path component (same contract as MinioConfig.DownloadURL).
    BucketURL string

    // Prefix is an optional object-key prefix for multi-environment
    // isolation (e.g. \"staging/\" or \"prod/\"). Mirrors COSConfig.Prefix.
    Prefix string

    // UsePathStyle forces path-style addressing
    // (https://endpoint/bucket/key) instead of virtual-hosted style
    // (https://bucket.endpoint/key). Required by some S3 gateways and
    // by AWS for very old buckets or path-style-only endpoints; rarely
    // needed for AWS S3 in 2026, default false.
    UsePathStyle bool
}
```

### Wiring

Add to `Config`:
```go
S3 S3Config // S3 / S3-compatible storage configuration
```

In `ConfigureWithViper`, mirror the existing per-field Viper read pattern used for `MinioConfig` and `COSConfig`:
```go
c.S3.Endpoint        = c.getString(\"s3.endpoint\",        c.S3.Endpoint)
c.S3.Region          = c.getString(\"s3.region\",          c.S3.Region)
c.S3.Bucket          = c.getString(\"s3.bucket\",          c.S3.Bucket)
c.S3.AccessKeyID     = c.getString(\"s3.accessKeyID\",     c.S3.AccessKeyID)
c.S3.SecretAccessKey = c.getString(\"s3.secretAccessKey\", c.S3.SecretAccessKey)
c.S3.BucketURL       = c.getString(\"s3.bucketURL\",       c.S3.BucketURL)
c.S3.Prefix          = c.getString(\"s3.prefix\",          c.S3.Prefix)
c.S3.UsePathStyle    = c.getBool  (\"s3.usePathStyle\",    c.S3.UsePathStyle)
```

No defaults are populated for `Endpoint`, `Region`, or `Bucket` — downstream consumers are expected to fail-fast at startup when the operator selects an S3 backend but omits required fields. (`octo-server` will enforce this in the follow-up PR.)

### YAML shape (downstream usage example)

```yaml
fileService: \"awsS3\"
s3:
  endpoint: \"s3.us-west-2.amazonaws.com\"
  region: \"us-west-2\"
  bucket: \"my-bucket\"
  accessKeyID: \"AKIA...\"
  secretAccessKey: \"...\"
  bucketURL: \"https://my-bucket.s3.us-west-2.amazonaws.com\"  # optional
  prefix: \"prod/\"                                            # optional
  usePathStyle: false                                          # optional
```

## Backward compatibility

- `MinioConfig` and `COSConfig` are untouched. Existing deployments using `fileService: \"minio\"` or `fileService: \"tencentCOS\"` continue to work without any config migration.
- `S3Config` is an additive field on `Config`; the zero value is inert (no consumer in `octo-lib` itself reads it).
- The `s3.*` Viper key namespace does not currently collide with any other binding.

## Scope of this issue

This issue covers **only** the `octo-lib` change:
1. Add the `S3Config` struct.
2. Add the field on `Config`.
3. Wire the Viper bindings.
4. Document the struct fields (godoc) and add the YAML example to the relevant README section if one exists.

The follow-up work in `octo-server` (new `ServiceS3`, refactoring `ServiceCOS` into an S3 profile, the `awsS3` `fileService` dispatch, integration tests against MinIO, AWS IAM policy / CORS documentation) will land in a separate PR against `Mininglamp-OSS/octo-server` and depends on this one being merged first.

## Open questions for maintainers

1. **Field naming**: `BucketURL` mirrors `MinioConfig.DownloadURL` semantics but uses the COS spelling. Preference for `DownloadURL` to match Minio, or `BucketURL` to match COS, or a brand-new name (e.g. `PublicEndpoint`) to avoid further entrenching the legacy spellings?
2. **`UsePathStyle` default**: keep `false` (matches AWS S3 modern default, virtual-hosted style), or auto-detect from `Endpoint` shape? I lean toward explicit `false` default with operator opt-in for clarity.
3. **`Region` validation**: should `octo-lib` enforce a non-empty `Region` when `Endpoint` is set, or leave it to downstream? Current convention in `MinioConfig` / `COSConfig` is to do no validation here.
4. Any preference on whether this should go into a versioned config schema bump?

Happy to send the PR once direction is confirmed.

an9xyz:
## Summary

- Add `S3Config` struct for generic S3 / S3-compatible object storage (AWS S3, Cloudflare R2, Backblaze B2, Wasabi, MinIO with custom endpoints).
- Wire `s3.*` Viper bindings in `ConfigureWithViper` following the existing per-field pattern used for `MinioConfig` / `COSConfig`.
- Additive only — `MinioConfig` and `COSConfig` untouched; zero-value `S3` is inert.

Closes #38. This is the prerequisite for a follow-up `octo-server` PR that introduces an `awsS3` `fileService` and consolidates SigV4 / presigned-URL handling.

## Test plan

- [x] `go test -race ./config/ -run TestS3` passes (defaults + viper-driven population)
- [ ] Reviewer to confirm field naming choices flagged in the issue (`BucketURL` vs `DownloadURL`, default `UsePathStyle=false`, no `Region` validation here)

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -142,6 +142,7 @@ type Config struct {
 	Seaweed     SeaweedConfig // seaweedfs配置
 	Qiniu       QiniuConfig   // 七牛云配置
 	COS         COSConfig     // 腾讯云COS配置
+	S3          S3Config      // S3 / S3兼容存储配置（AWS S3、Cloudflare R2、Backblaze B2、Wasabi 等）
 
 	// ---------- 短信运营商 ----------
 	SMSCode                string // 模拟的短信验证码
@@ -648,6 +649,15 @@ func (c *Config) ConfigureWithViper(vp *viper.Viper) {
 	// 环境变量覆盖（安全：避免密钥硬编码在配置文件中）
 	StringEnv(&c.COS.SecretID, "TS_COS_SECRET_ID")
 	StringEnv(&c.COS.SecretKey, "TS_COS_SECRET_KEY")
+	// S3 / S3兼容存储
+	c.S3.Endpoint = c.getString("s3.endpoint", c.S3.Endpoint)
+	c.S3.Region = c.getString("s3.region", c.S3.Region)
+	c.S3.Bucket = c.getString("s3.bucket", c.S3.Bucket)
+	c.S3.AccessKeyID = c.getString("s3.accessKeyID", c.S3.AccessKeyID)
+	c.S3.SecretAccessKey = c.getString("s3.secretAccessKey", c.S3.SecretAccessKey)
+	c.S3.BucketURL = c.getString("s3.bucketURL", c.S3.BucketURL)
+	c.S3.Prefix = c.getString("s3.prefix", c.S3.Prefix)
+	c.S3.UsePathStyle = c.getBool("s3.usePathStyle", c.S3.UsePathStyle)
 
 	//#################### 短信服务 ####################
 	c.SMSCode = c.getString("smsCode", c.SMSCode)
@@ -1052,6 +1062,54 @@ type QiniuConfig struct {
 	BucketName string
 	AccessKey  string
 	SecretKey  string
+}
+
+// S3Config configures any S3-compatible object storage backend.
+// Use this for AWS S3 or third-party providers that speak the S3 protocol
+// (Cloudflare R2, Backblaze B2, Wasabi, etc.). Tencent Cloud COS and MinIO
+// continue to be configured via their existing dedicated blocks for
+// backward compatibility.
+type S3Config struct {
+	// Endpoint is the S3 API hostname, without scheme.
+	// Examples:
+	//   AWS S3:        s3.us-west-2.amazonaws.com
+	//   Cloudflare R2: <accountid>.r2.cloudflarestorage.com
+	//   MinIO:         minio.example.com:9000
+	// REQUIRED — no default.
+	Endpoint string
+
+	// Region is the SigV4 signing region.
+	// Examples: us-west-2, ap-southeast-1, auto (R2).
+	// REQUIRED — must match the bucket's actual region for AWS; some
+	// providers accept any value (e.g. R2 uses "auto").
+	Region string
+
+	// Bucket is the single bucket all uploads target. Object keys carry
+	// their own logical prefix (chat/, moment/, ...), so a single bucket
+	// is the recommended deployment topology.
+	Bucket string
+
+	// AccessKeyID / SecretAccessKey are the long-lived IAM credentials
+	// for the bucket. See the security docs for the minimum IAM policy.
+	AccessKeyID     string
+	SecretAccessKey string
+
+	// BucketURL, when set, overrides the URL returned by DownloadURL and
+	// signed against by presigned GET/PUT. Use this for CDN-fronted
+	// buckets or custom domains. Must be scheme://host[:port] with no
+	// path component (same contract as MinioConfig.DownloadURL).
+	BucketURL string
+
+	// Prefix is an optional object-key prefix for multi-environment
+	// isolation (e.g. "staging/" or "prod/"). Mirrors COSConfig.Prefix.
+	Prefix string
+
+	// UsePathStyle forces path-style addressing
+	// (https://endpoint/bucket/key) instead of virtual-hosted style
+	// (https://bucket.endpoint/key). Required by some S3 gateways and
+	// by AWS for very old buckets or path-style-only endpoints; rarely
+	// needed for AWS S3 in 2026, default false.
+	UsePathStyle bool
 }
 
 // COSConfig 腾讯云COS配置
--- a/config/config.go
+++ b/config/config.go
@@ -658,6 +658,12 @@ func (c *Config) ConfigureWithViper(vp *viper.Viper) {
 	c.S3.BucketURL = c.getString("s3.bucketURL", c.S3.BucketURL)
 	c.S3.Prefix = c.getString("s3.prefix", c.S3.Prefix)
 	c.S3.UsePathStyle = c.getBool("s3.usePathStyle", c.S3.UsePathStyle)
+	// 环境变量覆盖（安全：避免长期凭据硬编码在配置文件中）
+	// 先读取项目专属变量，再让 AWS 标准变量覆盖，便于 IAM role / IRSA 等 AWS 原生工作流
+	StringEnv(&c.S3.AccessKeyID, "TS_S3_ACCESS_KEY_ID")
+	StringEnv(&c.S3.SecretAccessKey, "TS_S3_SECRET_ACCESS_KEY")
+	StringEnv(&c.S3.AccessKeyID, "AWS_ACCESS_KEY_ID")
+	StringEnv(&c.S3.SecretAccessKey, "AWS_SECRET_ACCESS_KEY")
 
 	//#################### 短信服务 ####################
 	c.SMSCode = c.getString("smsCode", c.SMSCode)
--- a/config/config.go
+++ b/config/config.go
@@ -659,11 +659,13 @@ func (c *Config) ConfigureWithViper(vp *viper.Viper) {
 	c.S3.Prefix = c.getString("s3.prefix", c.S3.Prefix)
 	c.S3.UsePathStyle = c.getBool("s3.usePathStyle", c.S3.UsePathStyle)
 	// 环境变量覆盖（安全：避免长期凭据硬编码在配置文件中）
-	// 先读取项目专属变量，再让 AWS 标准变量覆盖，便于 IAM role / IRSA 等 AWS 原生工作流
-	StringEnv(&c.S3.AccessKeyID, "TS_S3_ACCESS_KEY_ID")
-	StringEnv(&c.S3.SecretAccessKey, "TS_S3_SECRET_ACCESS_KEY")
+	// 优先级遵循"越具体越优先"：YAML → AWS 通用变量 → TS_ 项目专属变量。
+	// AWS_* 提供 IRSA / IAM role 工作流的零配置兜底；TS_S3_* 表达对本应用的显式意图，
+	// 最后写入故胜出，避免宿主机上无关的 AWS 凭据静默覆盖到 R2 / B2 / MinIO 等非 AWS 部署。
 	StringEnv(&c.S3.AccessKeyID, "AWS_ACCESS_KEY_ID")
 	StringEnv(&c.S3.SecretAccessKey, "AWS_SECRET_ACCESS_KEY")
+	StringEnv(&c.S3.AccessKeyID, "TS_S3_ACCESS_KEY_ID")
+	StringEnv(&c.S3.SecretAccessKey, "TS_S3_SECRET_ACCESS_KEY")
 
 	//#################### 短信服务 ####################
 	c.SMSCode = c.getString("smsCode", c.SMSCode)
--- a/config/config_s3_test.go
+++ b/config/config_s3_test.go
@@ -95,19 +95,29 @@ s3:
 			wantSecret: "aws-secret",
 		},
 		{
-			name: "AWS standard env wins over TS_ prefixed env",
+			name: "TS_ prefixed env wins over AWS standard env",
 			envs: map[string]string{
 				"TS_S3_ACCESS_KEY_ID":     "ts-id",
 				"TS_S3_SECRET_ACCESS_KEY": "ts-secret",
 				"AWS_ACCESS_KEY_ID":       "aws-id",
 				"AWS_SECRET_ACCESS_KEY":   "aws-secret",
 			},
-			wantID:     "aws-id",
-			wantSecret: "aws-secret",
+			wantID:     "ts-id",
+			wantSecret: "ts-secret",
 		},
+	}
+	credEnvs := []string{
+		"TS_S3_ACCESS_KEY_ID",
+		"TS_S3_SECRET_ACCESS_KEY",
+		"AWS_ACCESS_KEY_ID",
+		"AWS_SECRET_ACCESS_KEY",
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
+			// 清空所有相关 env，避免宿主机/CI runner 上的真实凭据污染断言
+			for _, k := range credEnvs {
+				t.Setenv(k, "")
+			}
 			for k, v := range tt.envs {
 				t.Setenv(k, v)
 			}
--- a/config/config_s3_test.go
+++ b/config/config_s3_test.go
@@ -1,0 +1,71 @@
+package config
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/spf13/viper"
+)
+
+func TestS3Defaults(t *testing.T) {
+	cfg := New()
+	vp := viper.New()
+	cfg.ConfigureWithViper(vp)
+
+	if cfg.S3.Endpoint != "" {
+		t.Errorf("S3.Endpoint default should be empty, got %q", cfg.S3.Endpoint)
+	}
+	if cfg.S3.Region != "" {
+		t.Errorf("S3.Region default should be empty, got %q", cfg.S3.Region)
+	}
+	if cfg.S3.Bucket != "" {
+		t.Errorf("S3.Bucket default should be empty, got %q", cfg.S3.Bucket)
+	}
+	if cfg.S3.UsePathStyle {
+		t.Errorf("S3.UsePathStyle default should be false, got true")
+	}
+}
+
+func TestS3FromViper(t *testing.T) {
+	const yaml = `
+s3:
+  endpoint: s3.us-west-2.amazonaws.com
+  region: us-west-2
+  bucket: my-bucket
+  accessKeyID: AKIAEXAMPLE
+  secretAccessKey: secret
+  bucketURL: https://my-bucket.s3.us-west-2.amazonaws.com
+  prefix: prod/
+  usePathStyle: true
+`
+	vp := viper.New()
+	vp.SetConfigType("yaml")
+	if err := vp.ReadConfig(strings.NewReader(yaml)); err != nil {
+		t.Fatalf("read config: %v", err)
+	}
+
+	cfg := New()
+	cfg.ConfigureWithViper(vp)
+
+	tests := []struct {
+		name string
+		got  string
+		want string
+	}{
+		{"Endpoint", cfg.S3.Endpoint, "s3.us-west-2.amazonaws.com"},
+		{"Region", cfg.S3.Region, "us-west-2"},
+		{"Bucket", cfg.S3.Bucket, "my-bucket"},
+		{"AccessKeyID", cfg.S3.AccessKeyID, "AKIAEXAMPLE"},
+		{"SecretAccessKey", cfg.S3.SecretAccessKey, "secret"},
+		{"BucketURL", cfg.S3.BucketURL, "https://my-bucket.s3.us-west-2.amazonaws.com"},
+		{"Prefix", cfg.S3.Prefix, "prod/"},
+	}
+	for _, tt := range tests {
+		if tt.got != tt.want {
+			t.Errorf("S3.%s = %q, want %q", tt.name, tt.got, tt.want)
+		}
+	}
+	if !cfg.S3.UsePathStyle {
+		t.Errorf("expected S3.UsePathStyle=true")
+	}
+}
--- a/config/config_s3_test.go
+++ b/config/config_s3_test.go
@@ -69,3 +69,61 @@ s3:
 		t.Errorf("expected S3.UsePathStyle=true")
 	}
 }
+
+func TestS3CredentialsFromEnv(t *testing.T) {
+	const yaml = `
+s3:
+  accessKeyID: from-yaml-id
+  secretAccessKey: from-yaml-secret
+`
+	tests := []struct {
+		name       string
+		envs       map[string]string
+		wantID     string
+		wantSecret string
+	}{
+		{
+			name:       "TS_ prefixed env overrides YAML",
+			envs:       map[string]string{"TS_S3_ACCESS_KEY_ID": "ts-id", "TS_S3_SECRET_ACCESS_KEY": "ts-secret"},
+			wantID:     "ts-id",
+			wantSecret: "ts-secret",
+		},
+		{
+			name:       "AWS standard env overrides YAML",
+			envs:       map[string]string{"AWS_ACCESS_KEY_ID": "aws-id", "AWS_SECRET_ACCESS_KEY": "aws-secret"},
+			wantID:     "aws-id",
+			wantSecret: "aws-secret",
+		},
+		{
+			name: "AWS standard env wins over TS_ prefixed env",
+			envs: map[string]string{
+				"TS_S3_ACCESS_KEY_ID":     "ts-id",
+				"TS_S3_SECRET_ACCESS_KEY": "ts-secret",
+				"AWS_ACCESS_KEY_ID":       "aws-id",
+				"AWS_SECRET_ACCESS_KEY":   "aws-secret",
+			},
+			wantID:     "aws-id",
+			wantSecret: "aws-secret",
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			for k, v := range tt.envs {
+				t.Setenv(k, v)
+			}
+			vp := viper.New()
+			vp.SetConfigType("yaml")
+			if err := vp.ReadConfig(strings.NewReader(yaml)); err != nil {
+				t.Fatalf("read config: %v", err)
+			}
+			cfg := New()
+			cfg.ConfigureWithViper(vp)
+			if cfg.S3.AccessKeyID != tt.wantID {
+				t.Errorf("S3.AccessKeyID = %q, want %q", cfg.S3.AccessKeyID, tt.wantID)
+			}
+			if cfg.S3.SecretAccessKey != tt.wantSecret {
+				t.Errorf("S3.SecretAccessKey = %q, want %q", cfg.S3.SecretAccessKey, tt.wantSecret)
+			}
+		})
+	}
+}


Jerry-Xin (dismissed):
The PR is in scope for `octo-lib` and cleanly adds an inert S3 configuration block without disrupting existing storage config behavior.

💬 Non-blocking

🟡 Warning — [config/config.go](/private/tmp/pr-review-lxXao3/repo/config/config.go:656) and [config/config.go](/private/tmp/pr-review-lxXao3/repo/config/config.go:657): `S3.AccessKeyID` and `S3.SecretAccessKey` are only populated from Viper config keys. Since COS explicitly supports env overrides for secrets at [config/config.go](/private/tmp/pr-review-lxXao3/repo/config/config.go:650), consider adding equivalent overrides such as `TS_S3_ACCESS_KEY_ID` and `TS_S3_SECRET_ACCESS_KEY` so deployments are not nudged toward storing long-lived S3 credentials in config files.

🔵 Suggestion — [config/config.go](/private/tmp/pr-review-lxXao3/repo/config/config.go:1073): `Endpoint` is documented as “without scheme”, while MinIO-style local/S3-compatible deployments often need to distinguish `http` from `https`. If the follow-up implementation will assume HTTPS, this is fine, but it would be worth documenting that explicitly or adding a separate scheme/secure option later.

🔵 Suggestion — [config/config.go](/private/tmp/pr-review-lxXao3/repo/config/config.go:1097): The `BucketURL` comment says it is “signed against by presigned GET/PUT”. CDN/custom-domain download URLs are often not valid SigV4 signing endpoints unless they proxy the request exactly to S3 with matching host semantics. Consider narrowing the comment so future consumers do not accidentally sign uploads against a CDN-only URL.

🔵 Suggestion — [config/config.go](/private/tmp/pr-review-lxXao3/repo/config/config.go:29): If the planned downstream service selects `fileService: awsS3`, adding a `FileServiceAWSS3 FileService = "awsS3"` constant in this library would avoid downstream string literals. Not required for this config-only PR.

✅ Highlights

- The new `S3Config` is additive and zero-value inert, so existing `MinioConfig`, `COSConfig`, and default file-service behavior are unaffected.
- Viper bindings follow the existing per-field style used by the surrounding file-service configs.
- Tests cover default inert behavior and Viper-driven population of all new fields.
- Verified locally: `go test -race ./config/ -run TestS3` passes, and `go test ./...` passes.

lml2468 (commented):
@an9xyz Gate check failed, so I am stopping before code review.

Blocking items:
1. PR checks: `label / Label PR size and detect dependency changes` is failing. Please fix or rerun that check, then request another review.

Non-blocking notes:
- The branch is mergeable and the code CI jobs shown by `gh pr checks` passed, but the review spec requires stopping on any red gate check.

Highlights:
- Not reviewed yet: diff and source context were intentionally not read after the failed gate.

yujiawei (dismissed):
# Code Review — PR #39 (octo-lib)

## Summary

Clean, minimal, additive PR. It implements exactly what issue #38 specifies:
a new `S3Config` struct, the `s3.*` Viper bindings using the established
per-field `getString` / `getBool` pattern, and a focused test file that
covers both the zero-value default path and the viper-driven population
path. No existing types (`MinioConfig`, `COSConfig`, etc.) are touched,
and the zero-value `S3` field is inert for any consumer in this library,
so backward compatibility is intact.

Verified locally against head `7051ff1`:

- `go test -race ./config/ -run TestS3` — pass
- `go test ./config/` (full package) — pass
- `go vet ./config/` — clean
- Pattern parity with `MinioConfig` / `COSConfig` / `QiniuConfig` wiring at `config/config.go:612-648` — confirmed; the new S3 block at `config/config.go:652-660` mirrors the conventions exactly.
- Godoc on each field is unusually thorough — endpoint examples per provider, explicit `REQUIRED` markers, the `BucketURL` semantics, and the `UsePathStyle` rationale are all called out.

No P0 / P1 issues.

## Findings

### P0 / P1
None.

### P2 — suggestion: env-var override for S3 credentials (security parity)

`config/config.go:649-651` was added recently by PR #189 (`fix(config):
support env override for COS secrets`) precisely because long-lived
cloud credentials in YAML are a recurring footgun:

```go
// 环境变量覆盖（安全：避免密钥硬编码在配置文件中）
StringEnv(&c.COS.SecretID, "TS_COS_SECRET_ID")
StringEnv(&c.COS.SecretKey, "TS_COS_SECRET_KEY")
```

`S3Config` introduces a new pair of long-lived IAM credentials with
the same threat model but does not follow the same hardening pattern.
Consider adding, right after the `S3` block (around `config/config.go:660`):

```go
StringEnv(&c.S3.AccessKeyID, "TS_S3_ACCESS_KEY_ID")
StringEnv(&c.S3.SecretAccessKey, "TS_S3_SECRET_ACCESS_KEY")
```

Optional bonus: honor the AWS-standard `AWS_ACCESS_KEY_ID` /
`AWS_SECRET_ACCESS_KEY` env vars as well, since operators deploying
to AWS will reach for those by reflex and they compose cleanly with
IAM role / IRSA setups in the follow-up `octo-server` PR.

Not a merge blocker — this is a hardening follow-up and could
reasonably land alongside the `octo-server` consumer in PR-2 — but
it is the strongest concrete suggestion in this review and the cheapest
to include now while the file is open.

### P2 — answers to the three open questions in the PR body

1. **`BucketURL` vs `DownloadURL`.** Keep `BucketURL`. The godoc
   explicitly says the field is also signed against by presigned
   PUT, so `DownloadURL` would be actively misleading; `BucketURL`
   stays semantically accurate for both GET and PUT and avoids
   entrenching the legacy Minio spelling.

2. **`UsePathStyle` default `false`.** Correct for modern AWS S3
   and acceptable for R2 (virtual-hosted works there too); MinIO /
   path-style-only gateways require the explicit opt-in, which is
   clearly documented. No auto-detection needed — keep it explicit.

3. **`Region` validation in `octo-lib`.** Skip it here. Consistent
   with `MinioConfig` / `COSConfig` deferring validation to the
   downstream service implementation; the planned `octo-server`
   fail-fast on startup is the right enforcement point.

### P2 — nit: defaults test could pin the in-memory-default behavior

`TestS3Defaults` (`config/config_s3_test.go:10-27`) calls
`ConfigureWithViper` against an empty viper and asserts the result
is empty. That is correct, but it does not catch a regression
where someone rewrites a wiring line as
`c.S3.Bucket = c.getString("s3.bucket", "")` — i.e. accidentally
dropping the "respect any in-memory pre-set value" contract that
every other backend in this file relies on. Strengthening it would
mean pre-setting `cfg.S3.Bucket = "preset"` before
`ConfigureWithViper(vp)` with the empty viper and asserting the
value survives. Low priority — every other backend in this file
also lacks that test — but worth keeping in mind if you ever
generalize the defaults test.

### P2 — nit: `BucketURL` contract is documented but not enforced

The godoc on `BucketURL` (`config/config.go:1097-1100`) says
"Must be scheme://host[:port] with no path component". This is
not validated either here or in the test, which is consistent
with `MinioConfig.DownloadURL`'s analogous unchecked contract,
so it is fine to defer. Flagging only so the `octo-server`
follow-up doesn't forget to do a `url.Parse` sanity check before
building presigned URLs.

## Backward compatibility & blast radius

- `Config` gains one new field. Zero-value `S3` is inert for every
  consumer in `octo-lib` itself; only the follow-up `octo-server`
  PR will read it.
- `s3.*` Viper key namespace was checked against the rest of
  `ConfigureWithViper` — no collisions.
- No existing test was modified; only one new file added.

## Verdict

This PR is a textbook prerequisite landing: small, focused,
well-documented, exactly matching the issue spec, and additive in
both struct shape and Viper namespace. Tests are real (they
actually drive `ConfigureWithViper`, not just the struct). The
single hardening suggestion (env-var override for S3 credentials)
is a follow-up, not a blocker.

Approving.

lml2468 (approved):
[APPROVE] Config-only PR is well-structured, additive-safe, and ready to merge.

No blocking items found. The failing `label / Label PR size and detect dependency changes` check is a labeler bot workflow, not a code quality gate; Build, Test, Vet, and CodeQL are all green and the branch is mergeable.

💬 Non-blocking notes (not repeating points already raised by @Jerry-Xin):

- `Endpoint` is documented as scheme-free (e.g. `s3.us-west-2.amazonaws.com`) while `BucketURL` requires a full `scheme://host[:port]`. This is consistent with the `MinioConfig` convention but easy to confuse. A one-line note in the YAML config example or a follow-up doc update would help future adopters.

✅ Highlights

- Struct-level doc comments are thorough, with concrete examples for AWS S3, Cloudflare R2, and MinIO per field — unusually good for a config struct.
- Viper key naming (`s3.accessKeyID`, `s3.secretAccessKey`) is consistent with `oss.*` and `minio.*` patterns throughout `ConfigureWithViper`.
- `TestS3Defaults` and `TestS3FromViper` together cover all 8 new fields (including the `UsePathStyle` bool), leaving no untested surface.
- Zero-value inertness confirmed: an empty `S3Config` has no effect on existing file-service routing, Minio, or COS behavior.

lml2468 (dismissed):
Approved: this config-only PR is additive, tested, and does not change existing storage behavior or wire S3 into runtime client construction yet.

Blocking items:
1. None.

Non-blocking notes:
- Prior reviews already covered the S3 credential env-var parity and endpoint/BucketURL documentation concerns, so I am not repeating those as new findings.

Highlights:
- config/config.go:145 adds `Config.S3` as a zero-value inert field; `FileService` defaults and existing Minio/COS/OSS config paths are unchanged.
- config/config.go:653-660 binds the `s3.*` keys with the same per-field Viper pattern used by the surrounding storage configs.
- Head-branch trace confirmed `S3Config` is only present in the new config field, Viper bindings, tests, and type definition; there is no octo-lib S3 client construction to validate in this PR, matching the stated config-only scope.
- config/config_s3_test.go:10 and config/config_s3_test.go:29 cover empty defaults and Viper population for all new fields, including `UsePathStyle`.
- Verified against head 7051ff1d2f0262256a226e3fb68513022f0a3df8: `go test ./config/` and `go test -race ./config/ -run TestS3` both pass.

Jerry-Xin (approved):
Project relevance gate: passed; this PR extends the repository’s existing config package for file storage backends.

💬 Non-blocking
- 🟡 Warning: [config/config_s3_test.go:110](/private/tmp/pr-review-flXgEH/repo/config/config_s3_test.go:110) does not clear ambient `AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY` before each subtest. On a CI runner or developer machine with AWS credentials already set, the first case can unexpectedly read those variables because [config/config.go:665](/private/tmp/pr-review-flXgEH/repo/config/config.go:665) intentionally lets AWS env vars override `TS_` vars. Consider setting all four relevant env vars to `""` at the start of each subtest, then applying the case-specific values.
- 🟡 Warning: [config/config.go:663](/private/tmp/pr-review-flXgEH/repo/config/config.go:663) gives standard AWS env vars higher precedence than explicit `TS_S3_*` credentials and YAML config. That is documented and tested, but for generic S3-compatible storage it can surprise deployments that run with unrelated AWS credentials while targeting R2, B2, Wasabi, or MinIO. A safer precedence may be project-specific env first, AWS env only as fallback.
- 🔵 Suggestion: [config/config.go:32](/private/tmp/pr-review-flXgEH/repo/config/config.go:32) may also want a `FileServiceS3` / `FileServiceAWSS3` constant if the follow-up `octo-server` PR will use a stable `fileService` value. The raw string still works today, so this is not blocking.

✅ Highlights
- The new `S3Config` is additive and zero-value inert, matching the PR’s compatibility goal.
- Viper field wiring follows the existing per-field pattern used by `MinioConfig` and `COSConfig`.
- Local verification passed: `go test -race ./config/ -run TestS3` and `go test ./...`.

lml2468 (dismissed):
Re-review after 06084a3: the incremental commit adds `StringEnv` overrides for S3 credentials — exactly the hardening suggested in the first round.

Previous blocking items: none (was APPROVED).

Incremental diff review (7051ff1 → 06084a3):

1. `config/config.go:661-666` — adds `TS_S3_ACCESS_KEY_ID` / `TS_S3_SECRET_ACCESS_KEY` then `AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY`. Since `StringEnv` only writes when env is non-empty, the cascade is correct: YAML → TS_ → AWS_. AWS standard vars winning last is the right call for IAM role / IRSA workflows.

2. `config/config_s3_test.go:72-129` — `TestS3CredentialsFromEnv` covers all three precedence scenarios (TS-only, AWS-only, both-set). Uses `t.Setenv` for proper test isolation. All assertions check both key ID and secret.

No new issues. CI green (Build ✅ Test ✅ Vet ✅ CodeQL ✅).

Re-approving.

Jerry-Xin (dismissed):
Re-reviewed on 06084a3. Incremental change adds env-var override for S3 credentials (TS_S3_* + AWS_*), mirroring the existing COS hardening pattern. Precedence ordering is correct — TS_ first, then AWS overwrites — matching the stated intent for IRSA / IAM-role workflows. Test coverage is thorough (3 sub-tests covering each precedence layer). CI green. No blocking issues.

yujiawei (dismissed):
# Code Review — PR #39 (octo-lib)

## Summary

Clean, focused, additive prerequisite PR. Adds a neutrally-named `S3Config` struct plus `s3.*` Viper bindings that match the conventions already used for `MinioConfig`, `COSConfig`, and `QiniuConfig`. Existing types are untouched, the zero-value `S3` field is inert for every current consumer, and the incremental commit (06084a3) layered on the `StringEnv` credential overrides — including the AWS-standard `AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY` envs — that the earlier review round called out as the strongest hardening suggestion.

Verified at head `06084a3`:

- Wiring at `config/config.go:653-666` mirrors the COS / Minio per-field pattern at `config/config.go:612-651`.
- `StringEnv` (`config/config.go:996-1001`) only writes when the env value is non-empty, so the four-line cascade `TS_` → `AWS_` produces the documented precedence: **YAML → `TS_S3_*` → `AWS_*`** (AWS standard wins last).
- Tests (`config/config_s3_test.go`) drive `ConfigureWithViper` end-to-end and assert all three precedence layers.
- All CI checks green (Build / Test / Vet / CodeQL).
- No collisions in the new `s3.*` Viper namespace.

## Findings

### P0 / P1
None.

### P2 — nit: `TestS3CredentialsFromEnv` is not isolated against pre-set environment

`config/config_s3_test.go:72-129` uses `t.Setenv` only for the vars each subtest explicitly cares about. If the developer or CI shell already has e.g. `AWS_ACCESS_KEY_ID` exported (common on AWS-deployed CI runners or for engineers using `aws-vault`), the first subtest `"TS_ prefixed env overrides YAML"` will read that pre-existing value through the `StringEnv(&c.S3.AccessKeyID, "AWS_ACCESS_KEY_ID")` line and fail with an unexpected ID. Cheap fix: at the top of each subtest, clear the four relevant vars explicitly:

```go
for _, k := range []string{
    "TS_S3_ACCESS_KEY_ID", "TS_S3_SECRET_ACCESS_KEY",
    "AWS_ACCESS_KEY_ID", "AWS_SECRET_ACCESS_KEY",
} {
    t.Setenv(k, "")
}
```

`t.Setenv` with `""` is effectively an unset for `StringEnv`'s purposes (it only writes when `os.Getenv(...) != ""`), and the original cleanup behavior of `t.Setenv` still restores the prior value on test end. Not a merge blocker — most CI today doesn't pre-export AWS creds — but it removes a latent flake.

### P2 — nit: `BucketURL` and `Endpoint` contracts are documented but not enforced

The godoc on `Endpoint` (`config/config.go:1083`) says "without scheme" and on `BucketURL` (`config/config.go:1097-1100`) says "Must be scheme://host[:port] with no path component." Neither is validated here or in the test, which matches `MinioConfig.DownloadURL`'s analogous unchecked contract. Acceptable to defer — flagging only so the follow-up `octo-server` consumer remembers to do a `url.Parse` sanity check before building presigned URLs, ideally with a fail-fast on startup.

### P2 — nit: defaults test could pin the in-memory-default contract

`TestS3Defaults` (`config/config_s3_test.go:10-27`) only asserts that the zero-value path stays zero. It would not catch a regression where someone rewrites a wiring line as `c.S3.Bucket = c.getString("s3.bucket", "")` — i.e. drops the "respect any in-memory pre-set value" contract that the rest of `ConfigureWithViper` relies on. Strengthen by pre-setting e.g. `cfg.S3.Bucket = "preset"` before calling `ConfigureWithViper(vp)` with an empty viper and asserting it survives. Low priority — every other backend in this file also lacks this guard — but a worthwhile pattern if you ever generalize the defaults test.

### P2 — UX consideration on AWS-standard env precedence

The current order — AWS standard vars override `TS_S3_*` — is the right call for IAM-role / IRSA workflows where AWS credentials may be injected by the platform and should win. Worth keeping in mind: developers who have unrelated `AWS_ACCESS_KEY_ID` exported (for personal AWS work) will see those leak into the `octo-lib`-configured bucket if they also rely on `TS_S3_*` for staging. The chosen precedence is defensible; the situation is just worth documenting in operator-facing docs once the `octo-server` consumer lands.

## Backward compatibility & blast radius

- `Config` gains exactly one new field; `MinioConfig` / `COSConfig` / `QiniuConfig` are untouched.
- Zero-value `S3` is inert — no consumer in `octo-lib` itself reads it yet; the follow-up `octo-server` PR is the first user.
- `s3.*` Viper namespace verified clean (no collisions with existing keys).
- No existing tests modified.

## Verdict

Tight, well-documented, hardening already in place, two prior approvals on the same SHA, all CI green. The remaining items are nits / future-proofing for the downstream consumer, not blockers.

Approving.

lml2468 (approved):
Re-review after 2948789 (incremental from 06084a3).

## Incremental diff (06084a3 → 2948789)

Two changes, both improvements:

1. **Env-var precedence flipped** (config/config.go:662-667): Now `AWS_*` is read first, then `TS_S3_*` overwrites. Since `StringEnv` is last-writer-wins, `TS_S3_*` now takes priority over `AWS_*`. This is the correct design: project-specific intent (`TS_S3_*`) should beat ambient host credentials (`AWS_*`), especially for non-AWS S3-compatible deployments (R2, B2, MinIO). Comment updated to explain the rationale clearly.

2. **Test env cleanup** (config/config_s3_test.go:98-114): All credential env vars are now zeroed via `t.Setenv("", ...)` before each subtest, preventing host/CI runner credentials from polluting assertions. The "both set" test case correctly asserts `TS_*` wins. Good hardening.

No new issues. CI: Build ✅ Test ✅ Vet ✅.

Jerry-Xin (approved):
Re-reviewed on 2948789. Incremental fix addresses two issues:

1. **Precedence reversal** — TS_S3_* now wins over AWS_*, following "more-specific wins" semantics. This prevents ambient AWS credentials on developer machines or CI runners from silently bleeding into non-AWS deployments (R2/B2/MinIO). The new ordering YAML → AWS_* → TS_S3_* correctly mirrors the SDK specificity model.

2. **Test isolation** — all four credential env vars are explicitly cleared at the start of each subtest, eliminating flake potential on CI runners with ambient AWS_* set.

Both changes are well-motivated, minimal, and correctly implemented. CI green. No blocking issues.

yujiawei (approved):
# Code Review — PR #39 (octo-lib)

## Summary

Additive change introducing a generic `S3Config` struct (`config/config.go:1075-1121`) plus matching Viper bindings in `ConfigureWithViper` (`config/config.go:652-668`). `MinioConfig` and `COSConfig` are untouched; the zero-value `S3Config` is inert and no consumer in `octo-lib` itself reads it. Test coverage is solid (`config/config_s3_test.go`) and the package tests pass under `-race`.

Verdict: **APPROVED** — no correctness, security, or build-breaking issues. A couple of P2 nits below for the follow-up `octo-server` PR.

## 1. Verification

| Item | Result | Evidence |
|---|---|---|
| `S3Config` struct added with documented fields | ✅ | `config/config.go:1075-1121` |
| `Config.S3` field wired | ✅ | `config/config.go:145` |
| Viper bindings follow the existing per-field pattern (matches `COSConfig` block at `config/config.go:643-651`) | ✅ | `config/config.go:653-660` |
| Zero-value defaults remain empty | ✅ | `TestS3Defaults` at `config/config_s3_test.go:10-27` |
| YAML → struct round-trip | ✅ | `TestS3FromViper` at `config/config_s3_test.go:29-66` |
| Env-override precedence (YAML < AWS_* < TS_S3_*) | ✅ | `TestS3CredentialsFromEnv` at `config/config_s3_test.go:68-138` — relies on `StringEnv` ignoring empty values (`config/config.go:998-1003`) and last-write-wins ordering at `config/config.go:665-668` |
| `MinioConfig` / `COSConfig` untouched | ✅ | diff only adds; no removals |
| `go test -race ./config/` | ✅ | green locally, including the three new tests |
| `go build ./...` / `go vet ./...` | ✅ | clean |

## 2. Findings

No P0 / P1 issues.

### P2 — nit / follow-up

**P2-1. Silent override risk for non-AWS S3 backends on hosts with stray `AWS_*` env.** `config/config.go:665-666` reads `AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY` unconditionally, then `config/config.go:667-668` lets `TS_S3_*` win. The author's inline comment acknowledges the trade-off, and `TestS3CredentialsFromEnv` covers the precedence. The remaining footgun: an operator deploying R2 / B2 / MinIO who happens to be on a host with `AWS_ACCESS_KEY_ID` exported (CI runner with a global AWS profile, EC2 instance with a developer's `aws configure` left over, dev shell with `AWS_PROFILE` exported) gets those creds silently injected into `S3Config`. They cannot opt out by `unset`-ing or setting `TS_S3_ACCESS_KEY_ID=""`, because `StringEnv` ignores empty values (`config/config.go:998-1003`) — they must export a non-empty `TS_S3_ACCESS_KEY_ID`.

This is fine to land here (config-only stub, no consumer yet), but worth revisiting when `octo-server` actually wires `awsS3` as a `fileService`. Two cheap mitigations to consider in the follow-up PR:

- Only read `AWS_*` when `s3.endpoint` looks like an AWS endpoint (e.g. ends with `amazonaws.com`), or
- Drop the `AWS_*` read entirely and require operators to be explicit with `TS_S3_*`. The PR description's "IRSA / IAM role 零配置兜底" rationale doesn't actually apply to static `AWS_ACCESS_KEY_ID` — IRSA injects `AWS_WEB_IDENTITY_TOKEN_FILE` + `AWS_ROLE_ARN` and is consumed via the AWS SDK's default credential chain, not via static-AK env reads. So the "fallback" framing is slightly imprecise; in practice the `AWS_*` reads cover hand-rolled `aws configure`-style deployments, which is a smaller win than the comment implies.

Not blocking — just flagging so the follow-up PR doesn't carry forward an inaccurate mental model.

**P2-2. Doc nit — `BucketURL` contract.** The godoc says it "overrides the URL returned by DownloadURL and signed against by presigned GET/PUT" (`config/config.go:1105-1108`). Since `octo-lib` itself never reads `S3Config`, this contract is implicitly aspirational until `octo-server` lands. Worth re-reading the comment when the consumer PR opens to make sure the actual signing path matches the documented invariant ("scheme://host[:port] with no path component"). No action here.

**P2-3. Comment polish — `"越具体越优先"`.** The Chinese comment at `config/config.go:662-664` reads correctly but the phrase order ("YAML → AWS 通用变量 → TS_ 项目专属变量") describes the *write* sequence, not a precedence ladder; readers may parse "具体优先" as "first-write-wins". The behavior is correct (last write wins); a one-word tweak to "最后写入的最具体来源胜出" would make this less ambiguous. Pure cosmetics.

## 3. Recommendations

- Land as-is. The four open questions in the PR description (BucketURL vs DownloadURL, UsePathStyle default, Region validation, schema versioning) all have defensible answers in the current shape and don't block merge.
- In the follow-up `octo-server` PR: address P2-1 (gate `AWS_*` reads, or drop them), and add startup validation (`Endpoint`, `Region`, `Bucket` non-empty when `fileService=awsS3`).

## 4. Additional notes

- `getBool` (`config/config.go:840-846`) correctly distinguishes "key absent" from "key=false", so `s3.usePathStyle` can be omitted without flipping a previously-set default. Consistent with existing patterns.
- Viper's case-insensitive key handling means the YAML keys (`accessKeyID`, `usePathStyle`) match the lookup strings — verified by the green `TestS3FromViper`.
- `t.Setenv` provides automatic cleanup, and the test pre-clears all four credential env vars before each subtest, so this suite is safe to run on hosts that already export `AWS_ACCESS_KEY_ID`.

PR #43, opened 2026-05-25
feat(config): refine S3Config — DownloadURL rename + SessionToken

### LINKED ISSUE #42: feat(config): refine S3Config — rename BucketURL → DownloadURL, add SessionToken for STS/IRSA
## Summary

Two refinements to the freshly-landed `S3Config` (PR #39) surfaced while integrating it in [octo-server PR #147](https://github.com/Mininglamp-OSS/octo-server/pull/147) and reading reviewer feedback:

1. **Rename `BucketURL` → `DownloadURL`** so the field has a single, unambiguous responsibility (unsigned download URL only — never participates in SigV4 signing). The current name carries hidden two-role semantics that produced a critical SigV4 host-mismatch bug in the octo-server consumer.
2. **Add `SessionToken string`** so STS / IRSA / IMDSv2 / EKS Pod Identity workflows can pass temporary credentials in. The current static-only credential pair makes the IRSA story documented in the env-var override (`AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY`) only partially functional — the third leg (`AWS_SESSION_TOKEN`) is silently dropped today.

Both changes are safe to land before any production consumer exists (S3Config merged 2026-05-23; octo-server PR still open).

### Why the current name fails

`BucketURL` carries two responsibilities that turn out to conflict on AWS S3:

- **Browser-facing unsigned URL** — what `publicURL()` returns for `/v1/file/preview/*` redirects and for the `path` field in upload responses.
- **SigV4 signing host** — what `newPublicClient()` passes to minio-go for presigned PUT/GET URLs.

The downstream consumer (`octo-server/modules/file/service_s3.go`) tried to bridge the two roles by inspecting the BucketURL host shape (starts with `<bucket>.` → bucket-subdomain → DNS lookup; no `<bucket>.` prefix → CDN alias → path-style). Reviewers on PR #147 flagged this as broken for the documented CloudFront + Origin Access Control deployment recipe:

- Presigned PUT signed against `Host: d123.cloudfront.net` → AWS S3 rejects with `SignatureDoesNotMatch` (the canonical request expects the bucket-addressed S3 hostname).
- With OAC, CloudFront re-signs the origin request → the client's `Content-Length` (the MaxFileSize-bypass guard) is lost, security tradeoff is silently broken.
- Ambiguity when a bucket name happens to match a common subdomain word (`files`, `assets`, `media`).

A neutrally-named `DownloadURL` makes the contract enforceable: the field is consumed by `publicURL` / `DownloadURL` only, and **never** participates in SigV4 signing. Presigned URLs always sign against `Endpoint`. To deploy CloudFront / OAC, operators set `DownloadURL` to the CloudFront distribution URL and `Endpoint` to the canonical S3 API host; the browser hits CloudFront for unsigned reads and S3 directly for signed PUT/GET. This matches the actual AWS architecture and removes the host-shape detection magic entirely.

### Symmetry with MinioConfig

`MinioConfig` already has `DownloadURL` with the same role:

```go
type MinioConfig struct {
    URL         string // server-internal SDK target
    UploadURL   string // optional, server-side I/O override
    DownloadURL string // browser-facing host for presigned URLs
    ...
}
```

`S3Config.DownloadURL` would mirror this naming exactly. Operators switching backends between `minio` and `awsS3` keep the same mental model.

### Why a clean rename (no deprecation alias)

`S3Config` is 2 days old (commit `72ee7103`, 2026-05-23). No `octo-lib` consumer reads it yet — the only known downstream is octo-server PR #147, which is not merged and which I will sync to the new name. A clean rename now avoids the "deprecated alias maintained forever" tax.

### Field godoc (proposed)

```go
// DownloadURL is the browser-facing base URL used by publicURL / DownloadURL
// for unsigned object access (e.g. /v1/file/preview/* redirects, upload
// response `path` field). It does NOT participate in SigV4 signing —
// presigned PUT/GET URLs are always signed against `Endpoint`.
//
// Use cases:
//   - Empty: `DownloadURL` falls back to the canonical AWS hostname
//     constructed from `Endpoint` + `Bucket`.
//   - CloudFront / CDN front-door: set to the distribution URL, e.g.
//     `https://d123.cloudfront.net`. The browser fetches reads via
//     CloudFront (typically with OAC for private buckets); presigned
//     PUT/GET still flow directly to S3.
//   - Custom domain: set to your bucket-subdomain custom domain,
//     e.g. `https://my-bucket.cdn.example.com`.
//
// MUST be scheme://host[:port] with NO path component. Path prefixes
// are rejected at service construction.
DownloadURL string
```

### Viper key

`s3.downloadURL` (kebab-no-camel mirror of the field, matching the existing convention `s3.bucketURL`, `s3.accessKeyID`).

**Open question for maintainers**: keep a viper-level fallback for `s3.bucketURL` for one minor version, or fail loudly? My preference is to fail loudly given the field is days old.

### Why this matters

The current credential surface is:

```go
c.S3.AccessKeyID     // → minio-go credentials.NewStaticV4(id, secret, "")
c.S3.SecretAccessKey //                                              ^^
                     //                                              | session token always empty
```

The third arg to `NewStaticV4` is the STS session token. Without it, any rotating-credentials workflow is functionally disabled. Reviewers on PR #147 (Jerry-Xin / yujiawei / lml2468) all flagged that the env-var docs (`AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY` "for IRSA / IAM role workflows") overstate current capability — IRSA injects three vars, we only read two.

Adding the third closes the gap without a credential provider chain:

| Workflow | Mechanism | Today | After |
|---|---|---|---|
| Long-lived IAM user keys | Static `id` + `secret` | ✅ Works | ✅ Works |
| EKS IRSA (web identity) | Sidecar / webhook injects 3 env vars, refreshes before expiry | ❌ Token dropped | ✅ Works (deployment refreshes env, octo-server reads on startup / SIGHUP) |
| IMDSv2 (EC2 instance profile) | Pipeline reads IMDS, writes 3 env vars | ❌ Token dropped | ✅ Works |
| AWS SSO / `aws-vault` (developer laptops) | Tool exports 3 env vars | ❌ Token dropped | ✅ Works |

The deployment pipeline becomes responsible for refreshing the env vars before STS expiry (standard practice for IRSA webhooks, `aws-vault exec`, etc.). `octo-lib` itself does not need a credential provider chain — the env-var route is enough.

### Field godoc (proposed)

```go
// SessionToken is the AWS STS session token for temporary credentials.
// Empty for long-lived IAM user keys; populated by IRSA / IMDSv2 /
// EKS Pod Identity / AWS SSO workflows that issue rotating credentials.
//
// Lifecycle: ServiceS3 does NOT refresh the token itself — the
// deployment pipeline is responsible for refreshing `AWS_SESSION_TOKEN`
// (and the paired key/secret) before STS expiry. For static credentials
// leave this empty.
SessionToken string
```

### Viper + env wiring

```go
c.S3.SessionToken = c.getString("s3.sessionToken", c.S3.SessionToken)
StringEnv(&c.S3.SessionToken, "AWS_SESSION_TOKEN")
StringEnv(&c.S3.SessionToken, "TS_S3_SESSION_TOKEN")
```

Same env-var precedence pattern as `AccessKeyID` / `SecretAccessKey` (TS_-prefixed wins).

## What this issue is NOT proposing

Surveyed for completeness; explicitly out of scope here:

- `Insecure bool` / HTTP endpoint support — operators with HTTP needs (local MinIO emulator, dev with self-signed cert) should keep using the `minio` backend. S3 backend is HTTPS-only by design, consistent with its production framing.
- `RequestTimeout time.Duration` — belongs in the service layer (`context.WithTimeout`), not in config. ServiceS3 should ship a sensible default; tunable later if real workloads need it.
- `CABundle` / `InsecureSkipVerify` — corporate-CA edge cases. AWS SDK's standard `AWS_CA_BUNDLE` env var is already honored by the underlying HTTP transport. Add only if a concrete deployment needs it.
- AWS Transfer Acceleration — separate feature, separate PR.
- Multi-region / dual-stack endpoint pairs — single region is the deployment majority; advanced cases warrant separate design.

## Backward compatibility

- `MinioConfig` and `COSConfig` untouched.
- `S3Config.BucketURL` is removed (not renamed-with-alias) given the absence of `octo-lib` consumers in production. The single in-flight downstream consumer (octo-server PR #147) will be synced in the matching PR.
- `S3Config.SessionToken` is a new field; zero value (empty string) preserves existing static-credential behavior exactly.

## Scope of this issue

`octo-lib` only:

1. Rename `S3Config.BucketURL` field → `S3Config.DownloadURL`.
2. Rename viper key `s3.bucketURL` → `s3.downloadURL`.
3. Add `S3Config.SessionToken string` field.
4. Add viper key `s3.sessionToken` + `AWS_SESSION_TOKEN` / `TS_S3_SESSION_TOKEN` env bindings.
5. Update godoc for both fields with the single-responsibility / lifecycle wording above.
6. Update `config_s3_test.go` accordingly (rename the YAML key in `TestS3FromViper`, add a SessionToken case).

octo-server PR #147 will follow with the matching downstream changes (delete `publicEndpoint()` host-shape detection, route presigned URLs through canonical Endpoint always, pass SessionToken into `credentials.NewStaticV4`, update yaml docs, drop the three-option deployment recipe in favor of one clear `Endpoint` + `DownloadURL` story).

## Open questions for maintainers

1. **Viper key fallback for `s3.bucketURL`**: keep a one-minor-version warn-and-map fallback, or hard remove? I lean hard remove given the field's age.
2. **Field naming `DownloadURL` vs `PublicURL`**: `DownloadURL` mirrors `MinioConfig.DownloadURL` (precedent); `PublicURL` more honestly describes its role (it's also the host the upload response uses). Preference?
3. **`SessionToken` env-var name**: `AWS_SESSION_TOKEN` (standard) + `TS_S3_SESSION_TOKEN` (project override) — symmetric with the existing key/secret env layout. Confirm this is the right pattern?

Happy to send the PR once direction is confirmed.

an9xyz:
Closes #42.

## Summary

- Rename `S3Config.BucketURL` → `DownloadURL`; pin godoc contract to "never participates in SigV4 signing" so presigned PUT/GET always sign against `Endpoint`. Mirrors `MinioConfig.DownloadURL`.
- Add `S3Config.SessionToken` for STS / IRSA / IMDSv2 / EKS Pod Identity / AWS SSO. Wires both `AWS_SESSION_TOKEN` and `TS_S3_SESSION_TOKEN` with the same precedence as the existing key/secret pair (YAML → AWS_* → TS_S3_* wins).
- Viper keys: `s3.bucketURL` → `s3.downloadURL` (clean break, no alias); add `s3.sessionToken`.

## Why no deprecation alias

`S3Config` landed 2 days ago (`72ee710`, 2026-05-23). No `octo-lib` consumer reads it yet — the only known downstream is octo-server PR #147, which is not merged and will be synced in the matching PR. A clean break now avoids the "deprecated alias maintained forever" tax.

## Why the rename matters

The previous godoc explicitly said BucketURL is "signed against by presigned GET/PUT". Combined with its role as the unsigned browser URL, setting it to a CloudFront distribution produces `SignatureDoesNotMatch` on AWS S3 (canonical request expects the bucket-addressed S3 hostname). `DownloadURL` makes the single responsibility enforceable: it is only ever read by `publicURL` / `DownloadURL`-shaped helpers; presigned URLs always sign against `Endpoint`. To deploy CloudFront/OAC, operators set `DownloadURL` to the distribution URL and `Endpoint` to the canonical S3 API host.

## Out of scope (deferred per issue #42)

`Insecure`, `RequestTimeout`, `CABundle`, Transfer Acceleration, multi-region endpoint pairs.

## Test plan

- [x] `go test -race ./config/...` passes
- [x] `TestS3FromViper` updated for new YAML key + asserts SessionToken/DownloadURL
- [x] `TestS3CredentialsFromEnv` extended with SessionToken precedence cases (TS_ wins over AWS_*)
- [ ] octo-server PR #147 follow-up to consume new field names and pass `SessionToken` into `credentials.NewStaticV4`

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -655,7 +655,8 @@ func (c *Config) ConfigureWithViper(vp *viper.Viper) {
 	c.S3.Bucket = c.getString("s3.bucket", c.S3.Bucket)
 	c.S3.AccessKeyID = c.getString("s3.accessKeyID", c.S3.AccessKeyID)
 	c.S3.SecretAccessKey = c.getString("s3.secretAccessKey", c.S3.SecretAccessKey)
-	c.S3.BucketURL = c.getString("s3.bucketURL", c.S3.BucketURL)
+	c.S3.SessionToken = c.getString("s3.sessionToken", c.S3.SessionToken)
+	c.S3.DownloadURL = c.getString("s3.downloadURL", c.S3.DownloadURL)
 	c.S3.Prefix = c.getString("s3.prefix", c.S3.Prefix)
 	c.S3.UsePathStyle = c.getBool("s3.usePathStyle", c.S3.UsePathStyle)
 	// 环境变量覆盖（安全：避免长期凭据硬编码在配置文件中）
@@ -664,8 +665,10 @@ func (c *Config) ConfigureWithViper(vp *viper.Viper) {
 	// 最后写入故胜出，避免宿主机上无关的 AWS 凭据静默覆盖到 R2 / B2 / MinIO 等非 AWS 部署。
 	StringEnv(&c.S3.AccessKeyID, "AWS_ACCESS_KEY_ID")
 	StringEnv(&c.S3.SecretAccessKey, "AWS_SECRET_ACCESS_KEY")
+	StringEnv(&c.S3.SessionToken, "AWS_SESSION_TOKEN")
 	StringEnv(&c.S3.AccessKeyID, "TS_S3_ACCESS_KEY_ID")
 	StringEnv(&c.S3.SecretAccessKey, "TS_S3_SECRET_ACCESS_KEY")
+	StringEnv(&c.S3.SessionToken, "TS_S3_SESSION_TOKEN")
 
 	//#################### 短信服务 ####################
 	c.SMSCode = c.getString("smsCode", c.SMSCode)
@@ -1102,11 +1105,34 @@ type S3Config struct {
 	AccessKeyID     string
 	SecretAccessKey string
 
-	// BucketURL, when set, overrides the URL returned by DownloadURL and
-	// signed against by presigned GET/PUT. Use this for CDN-fronted
-	// buckets or custom domains. Must be scheme://host[:port] with no
-	// path component (same contract as MinioConfig.DownloadURL).
-	BucketURL string
+	// SessionToken is the AWS STS session token for temporary credentials.
+	// Empty for long-lived IAM user keys; populated by IRSA / IMDSv2 /
+	// EKS Pod Identity / AWS SSO workflows that issue rotating credentials.
+	//
+	// Lifecycle: the service layer does NOT refresh the token itself — the
+	// deployment pipeline is responsible for refreshing AWS_SESSION_TOKEN
+	// (and the paired key/secret) before STS expiry. Leave empty for
+	// static credentials.
+	SessionToken string
+
+	// DownloadURL is the browser-facing base URL used for unsigned object
+	// access (e.g. preview redirects, upload-response `path` field). It
+	// does NOT participate in SigV4 signing — presigned PUT/GET URLs are
+	// always signed against Endpoint.
+	//
+	// Use cases:
+	//   - Empty: callers fall back to the canonical hostname constructed
+	//     from Endpoint + Bucket.
+	//   - CloudFront / CDN front-door: set to the distribution URL, e.g.
+	//     https://d123.cloudfront.net. The browser fetches reads via
+	//     CloudFront (typically with OAC for private buckets); presigned
+	//     PUT/GET still flow directly to S3.
+	//   - Custom domain: set to a bucket-subdomain custom domain, e.g.
+	//     https://my-bucket.cdn.example.com.
+	//
+	// Must be scheme://host[:port] with no path component (same contract
+	// as MinioConfig.DownloadURL).
+	DownloadURL string
 
 	// Prefix is an optional object-key prefix for multi-environment
 	// isolation (e.g. "staging/" or "prod/"). Mirrors COSConfig.Prefix.
--- a/config/config.go
+++ b/config/config.go
@@ -1113,6 +1113,10 @@ type S3Config struct {
 	// deployment pipeline is responsible for refreshing AWS_SESSION_TOKEN
 	// (and the paired key/secret) before STS expiry. Leave empty for
 	// static credentials.
+	//
+	// Override precedence: YAML s3.sessionToken → AWS_SESSION_TOKEN →
+	// TS_S3_SESSION_TOKEN (project-prefixed wins). Same pattern as the
+	// AccessKeyID / SecretAccessKey pair.
 	SessionToken string
 
 	// DownloadURL is the browser-facing base URL used for unsigned object
--- a/config/config_s3_test.go
+++ b/config/config_s3_test.go
@@ -34,7 +34,8 @@ s3:
   bucket: my-bucket
   accessKeyID: AKIAEXAMPLE
   secretAccessKey: secret
-  bucketURL: https://my-bucket.s3.us-west-2.amazonaws.com
+  sessionToken: FwoGZXIvYXdzEXAMPLE
+  downloadURL: https://d123.cloudfront.net
   prefix: prod/
   usePathStyle: true
 `
@@ -57,7 +58,8 @@ s3:
 		{"Bucket", cfg.S3.Bucket, "my-bucket"},
 		{"AccessKeyID", cfg.S3.AccessKeyID, "AKIAEXAMPLE"},
 		{"SecretAccessKey", cfg.S3.SecretAccessKey, "secret"},
-		{"BucketURL", cfg.S3.BucketURL, "https://my-bucket.s3.us-west-2.amazonaws.com"},
+		{"SessionToken", cfg.S3.SessionToken, "FwoGZXIvYXdzEXAMPLE"},
+		{"DownloadURL", cfg.S3.DownloadURL, "https://d123.cloudfront.net"},
 		{"Prefix", cfg.S3.Prefix, "prod/"},
 	}
 	for _, tt := range tests {
@@ -75,42 +77,59 @@ func TestS3CredentialsFromEnv(t *testing.T) {
 s3:
   accessKeyID: from-yaml-id
   secretAccessKey: from-yaml-secret
+  sessionToken: from-yaml-token
 `
 	tests := []struct {
-		name       string
-		envs       map[string]string
-		wantID     string
-		wantSecret string
+		name      string
+		envs      map[string]string
+		wantID    string
+		wantSec   string
+		wantToken string
 	}{
 		{
-			name:       "TS_ prefixed env overrides YAML",
-			envs:       map[string]string{"TS_S3_ACCESS_KEY_ID": "ts-id", "TS_S3_SECRET_ACCESS_KEY": "ts-secret"},
-			wantID:     "ts-id",
-			wantSecret: "ts-secret",
+			name: "TS_ prefixed env overrides YAML",
+			envs: map[string]string{
+				"TS_S3_ACCESS_KEY_ID":     "ts-id",
+				"TS_S3_SECRET_ACCESS_KEY": "ts-secret",
+				"TS_S3_SESSION_TOKEN":     "ts-token",
+			},
+			wantID:    "ts-id",
+			wantSec:   "ts-secret",
+			wantToken: "ts-token",
 		},
 		{
-			name:       "AWS standard env overrides YAML",
-			envs:       map[string]string{"AWS_ACCESS_KEY_ID": "aws-id", "AWS_SECRET_ACCESS_KEY": "aws-secret"},
-			wantID:     "aws-id",
-			wantSecret: "aws-secret",
+			name: "AWS standard env overrides YAML",
+			envs: map[string]string{
+				"AWS_ACCESS_KEY_ID":     "aws-id",
+				"AWS_SECRET_ACCESS_KEY": "aws-secret",
+				"AWS_SESSION_TOKEN":     "aws-token",
+			},
+			wantID:    "aws-id",
+			wantSec:   "aws-secret",
+			wantToken: "aws-token",
 		},
 		{
 			name: "TS_ prefixed env wins over AWS standard env",
 			envs: map[string]string{
 				"TS_S3_ACCESS_KEY_ID":     "ts-id",
 				"TS_S3_SECRET_ACCESS_KEY": "ts-secret",
+				"TS_S3_SESSION_TOKEN":     "ts-token",
 				"AWS_ACCESS_KEY_ID":       "aws-id",
 				"AWS_SECRET_ACCESS_KEY":   "aws-secret",
+				"AWS_SESSION_TOKEN":       "aws-token",
 			},
-			wantID:     "ts-id",
-			wantSecret: "ts-secret",
+			wantID:    "ts-id",
+			wantSec:   "ts-secret",
+			wantToken: "ts-token",
 		},
 	}
 	credEnvs := []string{
 		"TS_S3_ACCESS_KEY_ID",
 		"TS_S3_SECRET_ACCESS_KEY",
+		"TS_S3_SESSION_TOKEN",
 		"AWS_ACCESS_KEY_ID",
 		"AWS_SECRET_ACCESS_KEY",
+		"AWS_SESSION_TOKEN",
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
@@ -131,8 +150,11 @@ s3:
 			if cfg.S3.AccessKeyID != tt.wantID {
 				t.Errorf("S3.AccessKeyID = %q, want %q", cfg.S3.AccessKeyID, tt.wantID)
 			}
-			if cfg.S3.SecretAccessKey != tt.wantSecret {
-				t.Errorf("S3.SecretAccessKey = %q, want %q", cfg.S3.SecretAccessKey, tt.wantSecret)
+			if cfg.S3.SecretAccessKey != tt.wantSec {
+				t.Errorf("S3.SecretAccessKey = %q, want %q", cfg.S3.SecretAccessKey, tt.wantSec)
+			}
+			if cfg.S3.SessionToken != tt.wantToken {
+				t.Errorf("S3.SessionToken = %q, want %q", cfg.S3.SessionToken, tt.wantToken)
 			}
 		})
 	}
--- a/config/config_s3_test.go
+++ b/config/config_s3_test.go
@@ -72,6 +72,27 @@ s3:
 	}
 }
 
+// TestS3LegacyBucketURLNotRead locks in the clean break from issue #42:
+// the deprecated `s3.bucketURL` viper key must not populate the renamed
+// `DownloadURL` field. If a future change reintroduces a fallback alias,
+// this test will fail and force an explicit decision.
+func TestS3LegacyBucketURLNotRead(t *testing.T) {
+	const yaml = `
+s3:
+  bucketURL: https://legacy.example.com
+`
+	vp := viper.New()
+	vp.SetConfigType("yaml")
+	if err := vp.ReadConfig(strings.NewReader(yaml)); err != nil {
+		t.Fatalf("read config: %v", err)
+	}
+	cfg := New()
+	cfg.ConfigureWithViper(vp)
+	if cfg.S3.DownloadURL != "" {
+		t.Errorf("legacy s3.bucketURL must not populate DownloadURL; got %q", cfg.S3.DownloadURL)
+	}
+}
+
 func TestS3CredentialsFromEnv(t *testing.T) {
 	const yaml = `
 s3:
@@ -86,6 +107,13 @@ s3:
 		wantSec   string
 		wantToken string
 	}{
+		{
+			name:      "no env vars — YAML wins",
+			envs:      nil,
+			wantID:    "from-yaml-id",
+			wantSec:   "from-yaml-secret",
+			wantToken: "from-yaml-token",
+		},
 		{
 			name: "TS_ prefixed env overrides YAML",
 			envs: map[string]string{


lml2468 (dismissed):
[APPROVE]

Clean rename + addition. Reviewed both files line-by-line:

**config.go**
- `BucketURL` → `DownloadURL` rename is semantically correct — the field never participates in SigV4 signing; the new name matches `MinioConfig.DownloadURL` and eliminates the previous misleading godoc ("signed against by presigned GET/PUT").
- `SessionToken` field + env wiring follows the exact same precedence pattern as the existing key/secret pair (`AWS_*` → `TS_S3_*`, last-write-wins). Correct for STS / IRSA / Pod Identity use cases.
- Godoc on both new fields is thorough and accurate.

**config_s3_test.go**
- `TestS3FromViper` covers YAML-only path for both new fields.
- `TestS3CredentialsFromEnv` extends all three env-override scenarios (TS_ only, AWS_ only, TS_ wins over AWS_) with SessionToken — mirrors existing key/secret coverage.
- Env cleanup list updated to include the new token vars.

No deprecation alias needed — `S3Config` landed 2 days ago with zero downstream consumers merged.

LGTM.

Jerry-Xin (dismissed):
This PR is in scope for `octo-lib` and the S3 config changes are consistent with the existing config package patterns.

💬 Non-blocking
- 🔵 Suggestion `config/config_s3_test.go:29`: Consider adding one regression test that `s3.bucketURL` no longer populates `S3.DownloadURL`. Since the clean break is intentional, a negative test would lock that behavior down.

✅ Highlights
- `config/config.go:658-671` correctly wires `s3.sessionToken`, `s3.downloadURL`, and the intended YAML → AWS_* → TS_S3_* precedence.
- `config/config.go:1118-1135` clearly documents that `DownloadURL` is browser-facing only and does not participate in SigV4 signing.
- `config/config_s3_test.go:75-159` covers `SessionToken` loading and env precedence, including TS_* winning over AWS_*.

Verification: `go test -race ./config/...` passes.

yujiawei (dismissed):
# Code Review — PR #43 (octo-lib)

## Verification

| Check | Result | Evidence |
|---|---|---|
| Field rename `S3Config.BucketURL` → `DownloadURL` is complete | ✅ | `config/config.go:1135`; `grep S3.BucketURL` across the repo returns 0 hits. The other `BucketURL` survivors (`OSSConfig` 1054, `COSConfig` 1155) are unrelated fields on different config blocks. |
| Viper key `s3.bucketURL` → `s3.downloadURL` rebound | ✅ | `config/config.go:659`. No fallback alias, matching the "no production consumer yet" justification in the PR body. |
| `S3Config.SessionToken` field added with godoc | ✅ | `config/config.go:1108-1116` (lifecycle expectation explicitly disclaimed). |
| Viper key `s3.sessionToken` wired | ✅ | `config/config.go:658`. |
| Env vars `AWS_SESSION_TOKEN` / `TS_S3_SESSION_TOKEN` wired with the same precedence as the existing key/secret pair | ✅ | `config/config.go:668,671` — sandwich is `YAML → AWS_* → TS_S3_*` (last writer wins), identical pattern to `AccessKeyID`/`SecretAccessKey` at lines 666-670. |
| Tests updated for new YAML key + SessionToken precedence cases | ✅ | `config/config_s3_test.go:37,38,61,62` (FromViper); `:80,94-95,103-105,116-119` (env precedence); `credEnvs` slice extended at `:129,132`. |
| `go test -race ./config/...` passes | ✅ | `ok github.com/Mininglamp-OSS/octo-lib/config 3.656s` against PR head `3562b6b`. |
| `go vet ./...` / `go build ./...` clean | ✅ | No diagnostics. |

## Findings

No P0 / P1 findings. The change is small, surgical, and consistent with the existing `MinioConfig.DownloadURL` precedent. Items below are P2/nit — none are blocking.

### P2 — Mismatched-credential-triple risk (pre-existing, worsened by this PR)

`config/config.go:666-671` independently overrides `AccessKeyID`, `SecretAccessKey`, and now `SessionToken` per env-var tier. If env vars are partially set across tiers, the resulting triple can be incoherent. Concrete example:

- YAML: empty
- AWS_ACCESS_KEY_ID + AWS_SECRET_ACCESS_KEY + AWS_SESSION_TOKEN all set (e.g. EKS Pod Identity webhook)
- TS_S3_ACCESS_KEY_ID set to a long-lived test key, TS_S3_SECRET_ACCESS_KEY / TS_S3_SESSION_TOKEN unset

End state: `AccessKeyID = ts-id`, `SecretAccessKey = aws-secret`, `SessionToken = aws-token` — three credentials that were never issued together. SigV4 will fail with `SignatureDoesNotMatch` or `InvalidToken` at request time, not at config load.

This pattern existed before this PR, but adding the third leg of the triple multiplies the failure modes. Two ways to harden, neither blocking for this PR:

1. **Triple-atomic override**: only override the triple if at least one of the three is set at a tier, and snapshot the other two from the same tier. This treats `(id, secret, token)` as a single credential object rather than three independent strings.
2. **Document the gotcha**: add a one-line comment near `config/config.go:663-665` explicitly warning that operators must set the three vars at the same tier (e.g. all three under `TS_S3_*` or all three under `AWS_*`), otherwise an incoherent credential will be assembled silently.

Recommendation: at minimum, take option 2 in this PR or a follow-up. Option 1 is the more correct fix but has a behavior-change footprint that warrants its own PR.

### P2 — `TS_S3_SESSION_TOKEN` is undocumented in the godoc

`config/config.go:1108-1116` mentions the lifecycle contract (deployment pipeline refreshes `AWS_SESSION_TOKEN`) but never names the project-prefixed `TS_S3_SESSION_TOKEN` override that lines 671 wires in. The existing `AccessKeyID`/`SecretAccessKey` godoc at `:1103-1106` has the same gap, so this is a consistency comment, not a regression. Worth a one-line addition in this PR while the area is hot:

```
// Override precedence: YAML s3.sessionToken → AWS_SESSION_TOKEN →
// TS_S3_SESSION_TOKEN (project-prefixed wins). Same pattern as the
// AccessKeyID / SecretAccessKey pair.
```

### Nit — Test coverage gap: YAML-only baseline for `SessionToken`

`TestS3CredentialsFromEnv` (`config/config_s3_test.go:75`) clears every env var and re-applies tier-specific cases, but no case asserts that `from-yaml-token` survives when no env vars are set. Symmetric with the existing gap for `from-yaml-id` / `from-yaml-secret`, so it's consistent. A short fourth case ("no env vars — YAML wins") would close the gap for all three credential fields at once.

### Nit — Clean-break decision

The "no `s3.bucketURL` alias" decision is well-justified in the PR body: `S3Config` landed 2 days ago, the only known consumer is the still-open octo-server #147. I confirmed there are no other S3Config readers in this repo (`grep S3Config` returns only the type definition and one reference at `config/config.go:145`). Reasonable call.

## Security-sensitive review

This PR is on the `needs-human-review` track, so calling out items a human auditor should verify manually:

- **No secret in logs / error strings.** `config/config.go` does not log any of `AccessKeyID`, `SecretAccessKey`, or the new `SessionToken` field. The change does not introduce any new logging path. ✅
- **Env-var precedence semantics.** The implemented pattern (`YAML → AWS_* → TS_S3_*`, last-writer-wins) is symmetric with what already shipped for `AccessKeyID`/`SecretAccessKey` and is what reviewers asked for in issue #42. ✅
- **No new attack surface.** The added field is a passive string carried through `S3Config`. It is not parsed, not hashed, not used to make outbound requests in this PR. It will be passed into the AWS SDK's `credentials.NewStaticV4` call site that lives in the downstream consumer (octo-server PR #147), which is out of scope for this review.
- **No credential rotation logic introduced.** The godoc explicitly disclaims this responsibility (`config/config.go:1112-1115`); refresh remains the deployment pipeline's job. ✅
- **Worth a human's eye:** the mismatched-triple risk above. It is not a vulnerability per se, but a misconfiguration mode that surfaces at runtime as a confusing AWS error rather than a config-load failure. Consider whether to harden this in a follow-up before broader rollout.

## Cross-repo coordination note (informational, not a review blocker)

Verified that the downstream consumer `octo-server` PR #147 still references `cfg.S3.BucketURL` (e.g. `modules/file/service_s3.go` and `api_unit_test.go` in that PR's diff). Per the "no deprecation alias" decision, that PR cannot pass `go build` against an octo-lib bumped to this commit until it is synced to `cfg.S3.DownloadURL`. The PR body acknowledges this. Maintainers may want to land this PR and the downstream sync close together to keep `octo-server`'s `main` healthy if it also re-pins octo-lib in the meantime — but neither is a precondition for merging *this* PR.

## Verdict

The implementation matches the design in issue #42 exactly, the rename is mechanically clean, the new field is wired symmetrically with the existing credential pattern, and the tests cover the precedence ladder. The remaining items are improvement suggestions, not changes the author must make before merge.

lml2468 (dismissed):
Changes requested because the PR currently fails the review gate.

I stopped at Phase 1 as requested: `check-sprint / check-sprint` is completed with `FAILURE`, and several checks are still pending. Please get CI green, then I can re-run the incremental review from `3562b6b53d45` to `bf1f4fa0ccb9`.

yujiawei (approved):
# Code Review — PR #43 (octo-lib)

## Summary

Config-only refinement to the freshly-landed `S3Config`:

1. `BucketURL` → `DownloadURL` (clean rename, no alias)
2. New `SessionToken` field + `AWS_SESSION_TOKEN` / `TS_S3_SESSION_TOKEN` env bindings
3. Viper key `s3.bucketURL` → `s3.downloadURL`

Scope matches issue #42 exactly; downstream `credentials.NewStaticV4` wiring is correctly deferred to octo-server PR #147.

## 1. Verification

| Item | Status | Evidence |
|---|---|---|
| Field rename applied at the struct | ✅ | `config/config.go:1139` (`DownloadURL string`) replaces the old `BucketURL` field at the same struct position |
| Viper key updated, no alias kept | ✅ | `config/config.go:659` reads `s3.downloadURL`; legacy `s3.bucketURL` is gone — and `TestS3LegacyBucketURLNotRead` (`config/config_s3_test.go:79`) locks the decision in |
| `SessionToken` field added with thorough godoc | ✅ | `config/config.go:1108-1120` |
| Viper key for session token | ✅ | `config/config.go:658` |
| Env bindings, AWS-then-TS order so TS wins | ✅ | `config/config.go:668,671` — `StringEnv` last-write-wins is the same precedence pattern as the existing key/secret pair |
| `StringEnv` empty-string semantics preserve precedence | ✅ | `config/config.go:1001-1006` — only overwrites when env value is non-empty, so an unset TS var does not blank out an AWS-injected token |
| `TestS3FromViper` asserts new keys | ✅ | `config/config_s3_test.go:61-62` |
| `TestS3CredentialsFromEnv` extended for all four cases (no-env, TS-only, AWS-only, TS-over-AWS) | ✅ | `config/config_s3_test.go:91-160` — including the new "no env vars — YAML wins" case which was missing before |
| Env cleanup list includes the three new vars | ✅ | `config/config_s3_test.go:162-167` |
| `go test -race ./config/...` passes | ✅ | green locally |
| No callers of `BucketURL` left in the repo | ✅ | only remaining hits for `BucketURL` are unrelated `OSSConfig` / `COSConfig` fields |
| No credential leakage via logs | ✅ | grep over `config/` finds no log/print of `s3` credentials |

## 2. Findings

None at P0 or P1 — no correctness, security, data-loss, or build-break issues.

## 3. Notes (non-blocking, P2 / nit)

### N1. Mixed-credential hazard via partial env override (pre-existing pattern)

`StringEnv` overrides per-variable, not per-credential-triple. Concretely, if an operator sets the full `AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY` / `AWS_SESSION_TOKEN` IRSA triple and also exports only `TS_S3_ACCESS_KEY_ID` (e.g. leftover from a different deployment), the resulting in-memory credentials are `TS-id` + `AWS-secret` + `AWS-token`. The IAM layer will reject this with a confusing `SignatureDoesNotMatch`.

This pattern predates this PR — the id/secret pair has had the same gap since `72ee710`. Adding `SessionToken` widens the mismatch surface from 2 to 3 variables. Not a blocker for this PR, but worth a doc note (or a future "if you set any `TS_S3_*` cred var, all three must be set together") and ideally a unit test that locks the contract once decided. File it as a follow-up under #42 if you don't want to scope-creep this PR.

### N2. `DownloadURL` shape is documented but not enforced here

The godoc at `config/config.go:1137-1138` says "Must be scheme://host[:port] with no path component", but `ConfigureWithViper` doesn't validate it — the original issue #42 noted "Path prefixes are rejected at service construction", which sites the check in octo-server. That's fine for `octo-lib` (config struct is intentionally dumb), but please make sure octo-server PR #147 actually performs that rejection at service construction, otherwise the godoc promise drifts from reality.

### N3. Godoc references `AWS_SESSION_TOKEN` env var name in `SessionToken` precedence text

`config/config.go:1117-1119` reads:

> Override precedence: YAML s3.sessionToken → AWS_SESSION_TOKEN → TS_S3_SESSION_TOKEN (project-prefixed wins).

This is accurate, but for symmetry consider linking to a single canonical "env-var precedence" doc block near `ConfigureWithViper` (lines 663-665 already explain the rationale once for the credential pair). Right now the rationale is duplicated in two places. Pure polish.

## 4. Additional Observations

- The decision to skip a deprecation alias is well-justified (2-day-old field, single in-flight consumer in octo-server PR #147). `TestS3LegacyBucketURLNotRead` is a thoughtful safety net.
- Godoc mentions the operational lifecycle — "the service layer does NOT refresh the token itself" — which is exactly the kind of contract that prevents future "why doesn't IRSA work after 1 hour?" issues. Good call.
- The renamed field is placed in the struct between the credential block and `Prefix`, which mirrors the visual flow of the godoc. Reads naturally.

## 5. Human-review attention points (security-sensitive PR)

Items worth a manual check before merging:

1. **Cross-repo coupling**: This PR is a breaking API change with no alias. Confirm `octo-server` PR #147 is genuinely the only consumer; if any private fork or other repo imports `octo-lib/config` and references `S3Config.BucketURL`, it will fail to compile after this lands. A quick `grep` across known internal consumers is cheap insurance.
2. **Release / version coordination**: If `octo-lib` is consumed via a Go module pin, the synced octo-server PR #147 must bump its `go.mod` to a `octo-lib` commit at or after this merge. Worth confirming the merge order is planned (octo-lib first, then octo-server).
3. **Audit trail for credential surface expansion**: `SessionToken` increases the credential material the process holds in memory. The PR does not introduce logging of it, which is correct, but please confirm any downstream observability / debug-dump paths in octo-server PR #147 also redact this field.

## Verdict

No P0 / P1 blockers. The PR is small, focused, well-tested, and the godoc educates operators on the SigV4 contract that the rename was designed to enforce. Approving.

lml2468 (approved):
# Code Review — PR #43 Re-review (bf1f4fa0ccb9)

**Incremental from:** `3562b6b5` → `bf1f4fa0`

## Summary

The new commit addresses Allen's previous 🟡 observation: adds godoc documenting the `TS_S3_SESSION_TOKEN` env var override precedence on the `SessionToken` field, and adds two new test cases (`TestS3LegacyBucketURLNotRead` ensuring clean break from deprecated `s3.bucketURL`, and a "no env vars — YAML wins" baseline case).

## Verification

**Full PR scope** (2 files, +72/-24):
- `BucketURL` → `DownloadURL` rename: clean, no remaining `BucketURL` references in S3 config (OSS/COS `BucketURL` is a separate concern, out of scope).
- `SessionToken` field added with symmetric env override chain: YAML → `AWS_SESSION_TOKEN` → `TS_S3_SESSION_TOKEN` (TS_ wins). Same pattern as AccessKeyID/SecretAccessKey. ✅
- Godoc on `SessionToken` now documents the override precedence and lifecycle expectations. ✅
- `TestS3LegacyBucketURLNotRead` locks in the clean break — prevents accidental fallback alias. ✅
- All 4 credential env test cases cover: YAML-only, TS_-only, AWS-only, TS_-over-AWS. SessionToken asserted in each. ✅

**CI:** Build ✅ | Test ✅ | Vet ✅ | CodeQL ✅ (`check-sprint` failure is a project management workflow, not a quality gate)

## Findings

No P0/P1 issues. Allen's 🟡 about per-field credential override producing inconsistent combos is a valid design observation but inherent to the existing env-override pattern (AccessKeyID/SecretAccessKey already had this) — not a regression from this PR.

## Verdict

**APPROVED** — rename is clean, SessionToken wiring is symmetric with existing credential fields, test coverage is thorough, and the incremental commit properly addresses the documentation gap.

lml2468 (approved):
[APPROVE] Re-review after new commits (3562b6b → bf1f4fa).

Delta from previous review:

1. **`SessionToken` godoc** now documents override precedence inline — clearer for consumers.
2. **`TestS3LegacyBucketURLNotRead`** — excellent regression guard: locks in the clean break so a future accidental viper alias for `s3.bucketURL` will fail loudly.
3. **"no env vars — YAML wins"** test case added — fills the baseline gap from the previous revision.

All changes are strictly additive improvements to docs and test coverage. Core logic unchanged. Still LGTM.

PR #45, opened 2026-05-27
fix: move permissions to workflow top level for pr-contributor-welcome

lml2468:
## 问题

`.github/workflows/pr-contributor-welcome.yml` 存在权限配置 bug：

- 顶层 `permissions: {}` 清空了所有权限
- job 级别的 `permissions: issues: write` 对 reusable workflow 调用（`uses:`）无效
- 导致 `GITHUB_TOKEN` 没有写权限，发欢迎评论时出现 403 错误

## 修复

将权限配置移至 workflow 顶层，并移除 job 级别的 `permissions` 块：

```yaml
# 修复前（错误写法）
permissions: {}

jobs:
  welcome:
    uses: Mininglamp-OSS/.github/.github/workflows/reusable-pr-contributor-welcome.yml@main
    permissions:
      issues: write

# 修复后（正确写法）
permissions:
  issues: write

jobs:
  welcome:
    uses: Mininglamp-OSS/.github/.github/workflows/reusable-pr-contributor-welcome.yml@main
```

## 参考

GitHub 文档：[reusable workflow 中的权限继承](https://docs.github.com/en/actions/using-workflows/reusing-workflows#supported-keywords-for-jobs-that-call-a-reusable-workflow) — job 级别的 `permissions` 对 `uses:` 类型的 job 不适用，必须在 workflow 顶层设置。

### DIFF
--- a/.github/workflows/pr-contributor-welcome.yml
+++ b/.github/workflows/pr-contributor-welcome.yml
@@ -10,11 +10,9 @@ on:
   pull_request_target: # zizmor: ignore[dangerous-triggers] metadata-only automation; no PR code executed
     types: [opened]
 
-permissions: {}
+permissions:
+  issues: write
 
 jobs:
   welcome:
     uses: Mininglamp-OSS/.github/.github/workflows/reusable-pr-contributor-welcome.yml@main
-    permissions:
-      issues: write
-      pull-requests: write


lml2468 (commented):
[APPROVE]

**Verdict: APPROVED** ✅ _(submitted as COMMENT because `gh` is authed as PR author)_

Clean, correct fix. The diff does exactly what the PR description says.

**Analysis:**

1. **Root cause confirmed** — the `welcome` CI job on this very PR fails with `403 Resource not accessible by integration` when calling `POST /issues/comments`, proving the bug is real.

2. **Fix is correct** — GitHub docs explicitly state that job-level `permissions` is [not supported for reusable workflow calls](https://docs.github.com/en/actions/using-workflows/reusing-workflows#supported-keywords-for-jobs-that-call-a-reusable-workflow) (`uses:`). Moving `permissions` to the workflow top level is the right approach.

3. **Dropping `pull-requests: write` is fine** — the `x-accepted-github-permissions` response header from the failed run shows `issues=write; pull_requests=write` (either suffices). Since the reusable workflow only posts a welcome comment via the Issues API, `issues: write` alone is sufficient. Principle of least privilege applied correctly.

4. **No security concerns** — this workflow uses `pull_request_target` (already noted with a zizmor ignore comment) and only posts a greeting; no PR code is checked out or executed.

LGTM. Ship it 🚢

yujiawei (approved):
# Code Review — PR #45 (octo-lib)

## Summary

Small, focused change to `.github/workflows/pr-contributor-welcome.yml` (+2 / -4 in one file). The end state is correct and tighter than what it replaces, so this is good to merge. One note on the PR description's technical framing is included below as a non-blocking suggestion.

## What the PR actually does

Before:
```yaml
permissions: {}

jobs:
  welcome:
    uses: Mininglamp-OSS/.github/.github/workflows/reusable-pr-contributor-welcome.yml@main
    permissions:
      issues: write
      pull-requests: write
```

After:
```yaml
permissions:
  issues: write

jobs:
  welcome:
    uses: Mininglamp-OSS/.github/.github/workflows/reusable-pr-contributor-welcome.yml@main
```

Two effective changes:
1. Hoist `issues: write` from job-level to workflow top-level.
2. Drop `pull-requests: write` entirely.

## Verification

✅ **End state is minimal and correct.** The reusable workflow `Mininglamp-OSS/.github/.github/workflows/reusable-pr-contributor-welcome.yml` only calls `github.rest.issues.createComment(...)` (line ~117 of that file) — PR comments use the issues REST API, so `issues: write` is sufficient. `pull-requests: write` was unused; removing it is a proper least-privilege cleanup.

✅ **Reusable workflow's own job-level `permissions: issues: write` is respected.** The reusable workflow declares `permissions: {}` at workflow level and `permissions: issues: write` on its job. The caller's GITHUB_TOKEN must grant at least what the callee asks for; top-level `issues: write` here satisfies that.

✅ **`pull_request_target` trigger is still safe.** Reusable workflow does not `actions/checkout` PR code; only reads PR metadata via API. The existing `zizmor: ignore[dangerous-triggers]` annotation remains accurate.

✅ **`actions/github-script` is SHA-pinned** (`@3a2844b7e9c422d3c10d287c895573f7108da1b3 # v9.0.0`) in the reusable workflow — supply-chain hygiene preserved.

✅ **No regression in workflow surface.** Only one job exists in this file, so promoting permissions from job-scope to workflow-scope does not widen exposure to any other job.

## Findings

### P0 / P1
None.

### P2 / Nits (non-blocking)

**N1 — PR description's technical premise is slightly off.** The description states that "job 级别的 `permissions: issues: write` 对 reusable workflow 调用（`uses:`）无效". This isn't quite right: per the GitHub Actions docs on [supported keywords for jobs that call a reusable workflow](https://docs.github.com/en/actions/sharing-automations/reusing-workflows#supported-keywords-for-jobs-that-call-a-reusable-workflow), `jobs.<job_id>.permissions` is explicitly listed as a supported keyword, and job-level permissions on a `uses:` job *do* propagate to the called workflow's GITHUB_TOKEN.

The most likely real cause of the 403 was a different interaction (e.g. the repository/org "Workflow permissions" default setting clamping the token before the job-level grant took effect, or a startup race against a freshly-changed org policy). Either way, the fix in this PR — hoisting to workflow-level and dropping the unused `pull-requests` scope — is still the right outcome; it's just worth noting that the framing "job-level permissions are invalid on `uses:`" is a misdiagnosis of the mechanism. Consider rewording the PR description (and any follow-up commit message / runbook) as "simplify and tighten permissions to match the reusable workflow's actual needs" rather than "job-level permissions don't work." Pure documentation polish — does not affect the diff.

**N2 — Optional future-proofing.** Because `permissions:` is now at workflow scope, any future job added to this file inherits `issues: write` by default. Today there's only one job so this is fine, but if a second job is ever added that doesn't need write, prefer re-introducing the job-level scope on the new job (e.g. `permissions: contents: read`) rather than widening the top-level grant.

## Recommendation

The change is small, correct, and strictly reduces privilege. Approving.

Jerry-Xin (approved):
Summary: This PR is in scope for `octo-lib` and the workflow change is safe to merge.

💬 Non-blocking
- 🟡 Warning: `.github/workflows/pr-contributor-welcome.yml:13` now grants `issues: write` at workflow scope. This is acceptable because the workflow has only one job, but the PR description’s claim that job-level `permissions` are invalid for reusable workflow caller jobs appears inaccurate under current GitHub Actions syntax. The effective fix here is still compatible with the called reusable workflow, which only needs `issues: write` to create PR comments.
- 🔵 Suggestion: `.github/workflows/pr-contributor-welcome.yml:18` still references the reusable workflow by mutable `@main`. That is an existing pattern in this repo, not introduced here, but pinning to a SHA would improve supply-chain stability if the organization does not require live central workflow updates.

✅ Highlights
- The change keeps the token narrowly scoped to `issues: write`, which is sufficient for issue/PR comments.
- Removing `pull-requests: write` reduces unnecessary token permission for this welcome-comment workflow.
- The `pull_request_target` usage remains limited to metadata-only reusable automation and does not check out PR code.

PR #47, opened 2026-05-27
feat(wkhttp): add ErrorRenderer/TokenParser hooks + move ratelimit/secure_cors (#46)

### LINKED ISSUE #46: [i18n] Add ErrorRenderer hook + UserInfo/TokenParser primitives + migrate AuthMiddleware/RateLimit hardcoded strings
## Context

octo-server is rolling out cross-service i18n (see internal design v7.2). The shared HTTP layer in `pkg/wkhttp` is the only place where authentication and rate-limit errors are produced, but it currently hardcodes Chinese strings and provides no extension point for downstream renderers. This issue tracks the minimal changes octo-lib needs so downstream services can plug in a localizer without forking the package.

The proposal keeps backward compatibility: every new primitive ships with a default fallback so projects that do not opt in keep current behavior.

### 1. Move-in from octo-server (consolidate dual-source drift)

octo-server currently maintains a local copy of `pkg/wkhttp/{ratelimit.go, ratelimit_helper.go, secure_cors.go}` plus tests. Three rate-limit code paths there hardcode Chinese (`请求过于频繁，请稍后再试`) which the i18n initiative needs to migrate. Two-source maintenance is the root cause:

- Move `ratelimit.go`, `ratelimit_helper.go`, `secure_cors.go`, and their tests into `octo-lib/pkg/wkhttp/`.
- octo-server will then drop its local copy in a follow-up PR.

### 2. New primitives (no behavior change without opt-in)

New files under `pkg/wkhttp/`:

**`error_renderer.go`**
- `type ErrorSpec struct { Code, DefaultMessage string; TransportStatus, SemanticStatus int; Params, Details map[string]any; Internal bool }`
- `type ErrorRenderer interface { Render(c *Context, spec ErrorSpec) }`
- Default fallback renderer that writes `{msg: spec.DefaultMessage, status: spec.TransportStatus}` so library users without an injected renderer keep current behavior.
- Method on the existing `WKHttp` struct: `func (w *WKHttp) SetErrorRenderer(r ErrorRenderer)` — instance method, not a package-global (avoids cross-test contamination).
- Method on `Context`: `func (c *Context) RenderError(spec ErrorSpec)` that looks up the renderer through the owning `WKHttp`.

**`userinfo.go`**
- `type UserInfo struct { UID, Name, Role, Language string }`
- `func WithUser(ctx context.Context, info UserInfo) context.Context`
- `func UserFromCtx(ctx context.Context) (UserInfo, bool)`

**`token_parser.go`**
- `type TokenParser interface { Parse(ctx context.Context, token string) (UserInfo, error) }`
- `func (w *WKHttp) SetTokenParser(p TokenParser)` — instance method.
- AuthMiddleware uses the injected parser when present, otherwise falls back to the current ``strings.SplitN(value, \"@\", 3)`` behavior to preserve existing semantics.

### 3. Modifications to existing files

**`pkg/wkhttp/http.go` — AuthMiddleware (three hardcoded sites)**

Replace these three abort calls with `c.RenderError(ErrorSpec{...})`. Verified by direct inspection of the current `http.go`:

- `:274` `token不能为空，请先登录！` (401)
- `:281` `请先登录！` (401)
- `:288` `token有误！` (401, split-failure branch)

Suggested codes (final naming will be agreed with downstream):
- `err.shared.auth.token_missing`
- `err.shared.auth.required`
- `err.shared.auth.token_invalid`

AuthMiddleware should additionally write `WithUser(ctx, info)` alongside the existing `c.Set(\"uid\", ...)` / `c.Set(\"name\", ...)` / `c.Set(\"role\", ...)` so consumers can migrate gradually.

**`pkg/wkhttp/ratelimit.go` — three rate-limit sites (after move-in)**

All three currently emit `请求过于频繁，请稍后再试` (HTTP 429). Replace with `c.RenderError(ErrorSpec{Code: \"err.shared.rate.limited\", ..., Details: map[string]any{\"retry_after\": ...}})`. The three call sites are:

- Global per-IP limiter
- Strict per-IP limiter (login / SMS / search endpoints)
- Per-UID limiter

**`pkg/wkhttp/http.go` — CORS (`CORSMiddleware`, around line 340-356)**

- `AllowHeaders`: add `X-Octo-Error-Envelope`, `X-Octo-Lang`, `Accept-Language`.
- `ExposeHeaders`: **new field** — current `CORSMiddleware` does not set ExposeHeaders at all. Add `Content-Language`, `Vary`.

### 4. Open question: `CheckLoginRole` / `CheckLoginRoleIsSuperAdmin`

`pkg/wkhttp/http.go:161-164` and `:173-176` return Chinese errors through `errors.New`:

- `errors.New(\"登录用户角色错误\")`
- `errors.New(\"该用户无权执行此操作\")`

These cross the boundary into business code as `error` values and get rendered downstream. Two options:

1. Convert these methods to return `(ErrorSpec, bool)` (or similar) so the renderer can localize them — couples role checks to the new primitive.
2. Leave them as-is and require business code to translate at the call site.

We lean toward option 2 (smaller blast radius), but flagging for octo-lib maintainers to decide.

## Backward compatibility

- All new primitives are additive. Existing consumers see no behavior change unless they call `SetErrorRenderer` / `SetTokenParser`.
- The default fallback renderer emits the same envelope shape (`{msg, status}`) the current hardcoded paths produce.
- The default token parser preserves the current `uid@name@role` split — downstream can opt into a versioned JSON parser later.

## Out of scope (lives in downstream services)

- Translation files / locales
- Error code registry (`pkg/i18n/codes`)
- AST extractor / CI lint rules
- Email / push template i18n

## Suggested release plan

- Single minor version bumping with all changes above. The renderer hook + middleware sites must ship together so downstream can validate the full path in one upgrade.

## Testing

octo-lib side:
- Default fallback renderer keeps existing envelope under unit test.
- Concurrent `SetErrorRenderer` calls do not interleave state (instance-scoped guarantee).
- AuthMiddleware with default parser still passes existing tests; with an injected parser, returns `UserInfo` through `UserFromCtx`.
- RateLimit middleware renders `err.shared.rate.limited` via injected renderer.
- CORS middleware advertises the new Allow/Expose headers.

an9xyz:
Closes #46.

## Summary

Adds the minimal `pkg/wkhttp` primitives needed for downstream services (octo-server first) to roll out cross-service i18n, and consolidates the dual-source rate-limit/CORS files that were previously maintained in both octo-lib and octo-server.

All additions are backward compatible: services that do not call `SetErrorRenderer` / `SetTokenParser` keep working without changes. The default fallback renderer emits a `{msg, status}` envelope (matching the shape produced by every existing `Context.ResponseError*` helper), and the legacy `uid@name@role` token parser is used when no custom parser is injected.

**Minor wire-format note for AuthMiddleware aborts.** The three pre-existing `AuthMiddleware` abort paths previously emitted `{msg}` only — they did not include a `status` field, unlike the rest of the `ResponseError*` helpers. After this PR they go through the default renderer, which emits `{msg, status}`. The HTTP status code, `msg` value, and `Content-Type` are unchanged; only the JSON body gains a `status` field on these three branches. Clients that consume only `msg` are unaffected; clients that strictly schema-validate `{msg}` only would need to permit the extra field. This brings the auth aborts into line with the rest of the envelope shape and is generally desirable, but worth calling out explicitly.

## What's in

**New primitives**
- `error_renderer.go` — `ErrorSpec`, `ErrorRenderer` interface, `defaultErrorRenderer` fallback, `WKHttp.SetErrorRenderer`, `Context.RenderError`.
- `userinfo.go` — `UserInfo` struct + `WithUser` / `UserFromCtx` context helpers (D20).
- `token_parser.go` — `TokenParser` interface, `WKHttp.SetTokenParser`, `legacyTokenParser`, and exported sentinels `ErrTokenMissing` / `ErrTokenNotFound` / `ErrTokenInvalid`. Custom parsers should return these via `errors.Is` to inherit the same i18n key mapping.

**Migrations in `http.go`**
- `AuthMiddleware`: the three hardcoded Chinese abort strings now route through `c.RenderError` with codes `err.shared.auth.token_missing` / `required` / `token_invalid`. On success the middleware populates both `c.Set(\"uid\"/\"name\"/\"role\", ...)` (legacy) and `WithUser(ctx, info)` so callers can migrate gradually.
- `CORSMiddleware`: `Access-Control-Allow-Headers` extended with `X-Octo-Error-Envelope`, `X-Octo-Lang`, `Accept-Language`. New `Access-Control-Expose-Headers: Content-Language, Vary` — the original middleware did not emit `ExposeHeaders` at all, so this is a new header rather than an edit.

**Move-in from octo-server**
- `ratelimit.go`, `secure_cors.go`, pure env-parser portions of `ratelimit_helper.go`, plus tests. octo-server's local copies should be deleted in a follow-up PR.
- All three rate-limit abort paths now emit `err.shared.rate.limited` with `retry_after` carried in `ErrorSpec.Details`.

**Concurrency**
- Both `errorRenderer` and `tokenParser` use `atomic.Pointer[box]` rather than `atomic.Value`. `atomic.Value` requires the dynamic type of every stored value to match the first one, which would panic when different concrete renderer implementations are injected at different times (tests, multi-tenant wiring). `TestSetErrorRendererSwitchConcreteTypes` is the regression test.

### `ratelimit_helper.go` only partially moved

`SharedUIDRateLimiter` from octo-server's helper file stays in octo-server. It depends on `github.com/Mininglamp-OSS/octo-server/pkg/redis` and `octo-lib/config.Context`, so moving it would create a reverse dependency from lib to app. Only the pure helpers (`ParseRPSFromEnv`, `ParseBurstFromEnv`) moved.

Downstream impact: octo-server will keep its existing `SharedUIDRateLimiter` wrapper but switch its inner call to the new `*WKHttp` method (see signature change below).

### `RateLimitMiddleware` signature changed

The historical `RateLimitMiddleware(ctx, client, rps, burst, exclude...) gin.HandlerFunc` (package function) is now `(*WKHttp).RateLimitMiddleware(...) HandlerFunc` (instance method). Same for `StrictIPRateLimitMiddleware` and `UIDRateLimitMiddleware`. This is required because rate-limit aborts must call `c.RenderError`, which routes through the `*WKHttp`-scoped injected renderer (instance-scoped by design, not package-global, to avoid cross-test contamination).

Downstream impact (octo-server, ~5 call sites):
```go
// Before
r.UseGin(wkhttp.RateLimitMiddleware(ctx, rds, rps, burst, excludes...))
// After
r.Use(r.RateLimitMiddleware(ctx, rds, rps, burst, excludes...))
```

### `CheckLoginRole` / `CheckLoginRoleIsSuperAdmin` deliberately out of scope

`pkg/wkhttp/http.go:158-179` keeps returning hardcoded Chinese `errors.New(\"登录用户角色错误\")` and `errors.New(\"该用户无权执行此操作\")`. **Decision: option 2 from issue #46 §4 — translation is the responsibility of the calling service, not octo-lib.**

Downstream guidance (octo-server 0.7 thread / login / space adoption):
- Do not forward the raw error string from `c.CheckLoginRole()` to the client.
- Identify the error with `errors.Is`, then render it via the service's i18n-aware `ResponseErrorL` using business-side codes such as `err.shared.auth.role_missing` (\"登录用户角色错误\") and `err.shared.auth.role_forbidden` (\"该用户无权执行此操作\").
- These codes register in octo-server's i18n code registry, not in octo-lib.

Rationale:
1. These methods return `error` to the business layer, not directly to a response writer. Changing the signature to `(ErrorSpec, bool)` would touch every role-check call site.
2. octo-server's `ResponseError` path is already where translation gets centralised; role-check errors flow through it naturally.
3. Role-check failures are a sparse 403 path, not the high-frequency 401 path the middleware migration targets.

If multiple services later need to reuse the same role-check translations, we can open a follow-up to push the methods up to lib and ship a minor.

### `SecureCORSOverrideMiddleware` preflight ordering — known limitation, not fixed here

The middleware rewrites `Access-Control-Allow-Origin` per whitelist, but the existing `CORSMiddleware` aborts `OPTIONS` requests with status 204 before downstream middleware runs. That means credentialed preflights to whitelisted origins still see `Allow-Origin: *` from `CORSMiddleware` and are rejected by browsers — the whitelist hook works for non-preflight requests but not the preflight handshake.

This pre-exists in octo-server (the source file's docstring explicitly flagged it) and is not introduced by this PR. Fixing it requires merging `CORSMiddleware` and `SecureCORSOverrideMiddleware` into a single whitelist-aware middleware that handles `OPTIONS` last — a separable design change outside issue #46's scope. Tracking issue will follow.

In practice, production deployments terminate CORS at nginx / CLB, so the application-layer middleware is a fallback. The new `X-Octo-Lang` / `Accept-Language` headers in `Allow-Headers` are still advertised on the non-preflight path.

## Test plan

- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `go test -race -count=1 ./pkg/wkhttp/` passes (coverage 66.1%)
- [x] `TestDefaultRendererPreservesLegacyEnvelope` — fallback emits `{msg, status}` identical to prior hardcoded abort
- [x] `TestSetErrorRendererInstanceScoped` — two `WKHttp` instances do not leak renderer state across each other
- [x] `TestSetErrorRendererSwitchConcreteTypes` — regression test for the atomic.Value → atomic.Pointer[box] fix
- [x] `TestSetErrorRendererNilResetsToFallback` — `SetErrorRenderer(nil)` restores default; re-injection works
- [x] `TestSetErrorRendererConcurrent` — passes under `-race`
- [x] `TestAuthMiddlewareLegacyFallback` — three abort branches keep producing the historical Chinese messages with status 401
- [x] `TestAuthMiddlewareRendererRouted` — three abort branches surface the correct i18n codes via injected renderer
- [x] `TestAuthMiddlewareSuccessPopulatesContextAndUser` — both `c.Set` and `WithUser(ctx, ...)` populated
- [x] `TestAuthMiddlewareUsesInjectedTokenParser` / `TestAuthMiddlewareInjectedParserErrorFallsThrough` — custom parser path
- [x] `TestCORSMiddlewareAdvertisesI18nHeaders` — `Allow-Headers` includes the three new headers; `Expose-Headers` includes `Content-Language, Vary`
- [x] `TestRateLimitErrorSpec` — common ErrorSpec shape and `retry_after` details
- [x] `TestRateLimitMiddlewareDeniesViaRenderer` / `TestRateLimitMiddlewareDefaultFallback` — gated on testenv Redis, skip otherwise (consistent with octo-server's convention)

### DIFF
--- a/pkg/wkhttp/auth_middleware_test.go
+++ b/pkg/wkhttp/auth_middleware_test.go
@@ -1,0 +1,204 @@
+package wkhttp
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+)
+
+// fakeCache 是 cache.Cache 的内存实现，仅用于本包测试。
+type fakeCache struct {
+	data map[string]string
+}
+
+func newFakeCache(seed map[string]string) *fakeCache {
+	cp := make(map[string]string, len(seed))
+	for k, v := range seed {
+		cp[k] = v
+	}
+	return &fakeCache{data: cp}
+}
+
+func (c *fakeCache) Set(k, v string) error                                   { c.data[k] = v; return nil }
+func (c *fakeCache) Delete(k string) error                                   { delete(c.data, k); return nil }
+func (c *fakeCache) SetAndExpire(k, v string, _ time.Duration) error         { c.data[k] = v; return nil }
+func (c *fakeCache) Get(k string) (string, error)                            { return c.data[k], nil }
+
+// TestAuthMiddlewareLegacyFallback 验证三处历史错误分支映射到正确的 i18n key，
+// 同时未注入 renderer 时 envelope 与历史完全一致。
+func TestAuthMiddlewareLegacyFallback(t *testing.T) {
+	cache := newFakeCache(map[string]string{
+		"token:goodtoken": "uid1@alice@admin",
+		"token:bad":       "not-an-at-separated-value",
+	})
+
+	cases := []struct {
+		name       string
+		token      string
+		wantStatus int
+		wantMsg    string
+	}{
+		{"missing", "", http.StatusUnauthorized, "token不能为空，请先登录！"},
+		{"not_found", "unknown", http.StatusUnauthorized, "请先登录！"},
+		{"invalid_split", "bad", http.StatusUnauthorized, "token有误！"},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			wk := New()
+			mw := wk.AuthMiddleware(cache, "token:")
+
+			w := httptest.NewRecorder()
+			req := httptest.NewRequest(http.MethodGet, "/", nil)
+			if tc.token != "" {
+				req.Header.Set("token", tc.token)
+			}
+			runMiddleware(wk, mw, w, req)
+
+			if w.Code != tc.wantStatus {
+				t.Fatalf("status=%d want=%d body=%s", w.Code, tc.wantStatus, w.Body.String())
+			}
+			var body map[string]any
+			if err := json.Unmarshal(w.Body.Bytes(), &body); err != nil {
+				t.Fatalf("body not json: %v", err)
+			}
+			if body["msg"] != tc.wantMsg {
+				t.Errorf("msg=%v want=%v", body["msg"], tc.wantMsg)
+			}
+		})
+	}
+}
+
+// TestAuthMiddlewareRendererRouted 注入 renderer 后三处错误带各自的 i18n Code。
+func TestAuthMiddlewareRendererRouted(t *testing.T) {
+	cache := newFakeCache(map[string]string{
+		"token:bad": "not-an-at-separated-value",
+	})
+
+	cases := []struct {
+		name     string
+		token    string
+		wantCode string
+	}{
+		{"missing", "", "err.shared.auth.token_missing"},
+		{"not_found", "unknown", "err.shared.auth.required"},
+		{"invalid_split", "bad", "err.shared.auth.token_invalid"},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			wk := New()
+			stub := &stubRenderer{}
+			wk.SetErrorRenderer(stub)
+			mw := wk.AuthMiddleware(cache, "token:")
+
+			w := httptest.NewRecorder()
+			req := httptest.NewRequest(http.MethodGet, "/", nil)
+			if tc.token != "" {
+				req.Header.Set("token", tc.token)
+			}
+			runMiddleware(wk, mw, w, req)
+
+			if stub.lastSpec().Code != tc.wantCode {
+				t.Errorf("Code=%q want=%q", stub.lastSpec().Code, tc.wantCode)
+			}
+		})
+	}
+}
+
+// TestAuthMiddlewareSuccessPopulatesContextAndUser 验证成功路径同时写
+// c.Set(...) 与 WithUser(ctx, ...)——保证向后兼容 + 新 context 通道可用。
+func TestAuthMiddlewareSuccessPopulatesContextAndUser(t *testing.T) {
+	cache := newFakeCache(map[string]string{
+		"token:ok": "uid1@alice@admin",
+	})
+
+	wk := New()
+	var seenUID, seenName, seenRole string
+	var seenUser UserInfo
+	var seenOK bool
+
+	auth := wk.AuthMiddleware(cache, "token:")
+	assertHandler := HandlerFunc(func(c *Context) {
+		if v, ok := c.Get("uid"); ok {
+			seenUID, _ = v.(string)
+		}
+		if v, ok := c.Get("name"); ok {
+			seenName, _ = v.(string)
+		}
+		if v, ok := c.Get("role"); ok {
+			seenRole, _ = v.(string)
+		}
+		seenUser, seenOK = UserFromCtx(c.Request.Context())
+	})
+
+	w := httptest.NewRecorder()
+	req := httptest.NewRequest(http.MethodGet, "/", nil)
+	req.Header.Set("token", "ok")
+	runChain(wk, []HandlerFunc{auth, assertHandler}, w, req)
+
+	if w.Code != http.StatusOK && w.Code != 0 {
+		// httptest.NewRecorder 默认 200；中间件未 abort 时不会显式写 status。
+		t.Logf("status code = %d (acceptable if no body written)", w.Code)
+	}
+	if seenUID != "uid1" || seenName != "alice" || seenRole != "admin" {
+		t.Errorf("c.Set values: uid=%q name=%q role=%q", seenUID, seenName, seenRole)
+	}
+	if !seenOK {
+		t.Fatal("UserFromCtx returned ok=false")
+	}
+	if seenUser.UID != "uid1" || seenUser.Name != "alice" || seenUser.Role != "admin" {
+		t.Errorf("UserInfo = %+v", seenUser)
+	}
+}
+
+// stubTokenParser 让测试控制 Parse 的返回值，验证自定义 parser 优先级与
+// sentinel 映射。
+type stubTokenParser struct {
+	info UserInfo
+	err  error
+}
+
+func (s stubTokenParser) Parse(_ context.Context, token string) (UserInfo, error) {
+	return s.info, s.err
+}
+
+// TestAuthMiddlewareUsesInjectedTokenParser 验证 SetTokenParser 之后优先走自定义 parser。
+func TestAuthMiddlewareUsesInjectedTokenParser(t *testing.T) {
+	wk := New()
+	wk.SetTokenParser(stubTokenParser{info: UserInfo{UID: "custom", Name: "n", Role: "r", Language: "en"}})
+	mw := wk.AuthMiddleware(newFakeCache(nil), "ignored:")
+
+	w := httptest.NewRecorder()
+	req := httptest.NewRequest(http.MethodGet, "/", nil)
+	req.Header.Set("token", "any")
+	var seenUser UserInfo
+	runChain(wk, []HandlerFunc{mw, func(c *Context) {
+		seenUser, _ = UserFromCtx(c.Request.Context())
+	}}, w, req)
+
+	if seenUser.UID != "custom" || seenUser.Language != "en" {
+		t.Errorf("UserInfo = %+v", seenUser)
+	}
+}
+
+// TestAuthMiddlewareInjectedParserErrorFallsThrough 自定义 parser 返回未识别 error 时，
+// 默认归类为 err.shared.auth.required（避免泄露内部错误细节）。
+func TestAuthMiddlewareInjectedParserErrorFallsThrough(t *testing.T) {
+	wk := New()
+	stub := &stubRenderer{}
+	wk.SetErrorRenderer(stub)
+	wk.SetTokenParser(stubTokenParser{err: errors.New("custom parser failure")})
+	mw := wk.AuthMiddleware(newFakeCache(nil), "x:")
+
+	w := httptest.NewRecorder()
+	req := httptest.NewRequest(http.MethodGet, "/", nil)
+	req.Header.Set("token", "any")
+	runMiddleware(wk, mw, w, req)
+
+	if stub.lastSpec().Code != "err.shared.auth.required" {
+		t.Errorf("Code=%q want err.shared.auth.required", stub.lastSpec().Code)
+	}
+}
--- a/pkg/wkhttp/cors_test.go
+++ b/pkg/wkhttp/cors_test.go
@@ -1,0 +1,52 @@
+package wkhttp
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+)
+
+// TestCORSMiddlewareAdvertisesI18nHeaders 验证 i18n 协议要求的三个请求头
+// 出现在 Access-Control-Allow-Headers，两个响应头出现在 Access-Control-Expose-Headers。
+// 这是 issue 46 CORS 校准的核心验收。
+func TestCORSMiddlewareAdvertisesI18nHeaders(t *testing.T) {
+	wk := New()
+	wk.Use(CORSMiddleware())
+	wk.GET("/", noopHandler)
+
+	w := httptest.NewRecorder()
+	req := httptest.NewRequest(http.MethodGet, "/", nil)
+	wk.ServeHTTP(w, req)
+
+	allow := w.Header().Get("Access-Control-Allow-Headers")
+	for _, h := range []string{"X-Octo-Error-Envelope", "X-Octo-Lang", "Accept-Language"} {
+		if !strings.Contains(allow, h) {
+			t.Errorf("AllowHeaders missing %q: %s", h, allow)
+		}
+	}
+
+	expose := w.Header().Get("Access-Control-Expose-Headers")
+	if expose == "" {
+		t.Fatalf("Access-Control-Expose-Headers not set; want Content-Language, Vary")
+	}
+	for _, h := range []string{"Content-Language", "Vary"} {
+		if !strings.Contains(expose, h) {
+			t.Errorf("ExposeHeaders missing %q: %s", h, expose)
+		}
+	}
+}
+
+func TestCORSMiddlewareOptionsShortCircuits(t *testing.T) {
+	wk := New()
+	wk.Use(CORSMiddleware())
+	wk.Any("/", noopHandler)
+
+	w := httptest.NewRecorder()
+	req := httptest.NewRequest(http.MethodOptions, "/", nil)
+	wk.ServeHTTP(w, req)
+
+	if w.Code != http.StatusNoContent {
+		t.Errorf("OPTIONS status = %d, want 204", w.Code)
+	}
+}
--- a/pkg/wkhttp/error_renderer.go
+++ b/pkg/wkhttp/error_renderer.go
@@ -1,0 +1,83 @@
+package wkhttp
+
+import (
+	"net/http"
+
+	"github.com/gin-gonic/gin"
+)
+
+// ErrorSpec 描述一次错误响应的 primitive，供 ErrorRenderer 翻译/渲染。
+//
+// 设计要点：
+//   - Code 是稳定的 i18n key（如 "err.shared.auth.required"），由调用方保证全局唯一。
+//   - DefaultMessage 是未注入 renderer 时的兜底文案（与历史硬编码文案一致），
+//     保证 octo-lib 单独使用时行为不变。
+//   - TransportStatus 是 HTTP 状态码（实际写到响应头）。
+//   - SemanticStatus 是业务侧期望的状态码：上游可能因 envelope 协议把 TransportStatus
+//     固定为 200，但仍需保留语义 401/403/429——这两个字段允许分离。未来扩展用。
+//   - Params 用于翻译模板插值（如 {"retry_seconds": 30}），由 renderer 消费。
+//   - Details 携带不需要翻译的结构化字段，会被 renderer 透传到响应体或响应头
+//     （例如 rate-limit 的 retry_after）。
+//   - Internal=true 时 renderer 应避免把敏感内部细节暴露给客户端。
+type ErrorSpec struct {
+	Code            string
+	DefaultMessage  string
+	TransportStatus int
+	SemanticStatus  int
+	Params          map[string]any
+	Details         map[string]any
+	Internal        bool
+}
+
+// ErrorRenderer 由下游服务实现并通过 WKHttp.SetErrorRenderer 注入。
+// 同一个 *WKHttp 实例内只有一个 renderer 生效；不同 WKHttp 实例彼此隔离，
+// 因此并发测试中可以为每个实例独立注入而互不污染。
+type ErrorRenderer interface {
+	Render(c *Context, spec ErrorSpec)
+}
+
+// errorRendererBox 包装 ErrorRenderer 以便 atomic.Pointer 持有具体类型。
+// 直接存接口需要用 atomic.Value，但 atomic.Value 要求每次 Store 的动态类型
+// 一致——注入不同具体实现会 panic，因此采用 boxed 模式。
+type errorRendererBox struct {
+	r ErrorRenderer
+}
+
+// defaultErrorRenderer 是未显式注入 renderer 时的兜底实现，
+// 输出与历史 c.AbortWithStatusJSON 形态一致的 envelope：{msg, status}。
+// 这样未启用 i18n 的项目升级到本版本后行为完全不变。
+type defaultErrorRenderer struct{}
+
+// Render 输出 {msg: spec.DefaultMessage, status: spec.TransportStatus}。
+// 若 TransportStatus 为 0（调用方遗漏），回退到 400 以避免下游误判为成功。
+func (defaultErrorRenderer) Render(c *Context, spec ErrorSpec) {
+	status := spec.TransportStatus
+	if status == 0 {
+		status = http.StatusBadRequest
+	}
+	c.JSON(status, gin.H{
+		"msg":    spec.DefaultMessage,
+		"status": status,
+	})
+}
+
+// RenderError 通过所属 WKHttp 的 renderer 渲染错误。
+// 调用方仍需自行 c.Abort()（与 gin 中间件惯例一致），以确保后续 handler 不再执行。
+func (c *Context) RenderError(spec ErrorSpec) {
+	r := c.renderer()
+	r.Render(c, spec)
+}
+
+// renderer 返回当前 Context 所属 WKHttp 的 ErrorRenderer。
+// 兜底逻辑：若 Context 未关联 WKHttp（裸调 allocateContext 的单测）或 box 为 nil
+// （从未注入 / 注入 nil 重置），返回 defaultErrorRenderer，保证 RenderError 总是可用。
+func (c *Context) renderer() ErrorRenderer {
+	if c == nil || c.wk == nil {
+		return defaultErrorRenderer{}
+	}
+	box := c.wk.errorRenderer.Load()
+	if box == nil || box.r == nil {
+		return defaultErrorRenderer{}
+	}
+	return box.r
+}
--- a/pkg/wkhttp/error_renderer_test.go
+++ b/pkg/wkhttp/error_renderer_test.go
@@ -1,0 +1,222 @@
+package wkhttp
+
+import (
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"sync"
+	"sync/atomic"
+	"testing"
+
+	"github.com/gin-gonic/gin"
+)
+
+// stubRenderer 把每次渲染的 spec 暂存供断言，并写一个固定 envelope。
+// lastSpec 通过 atomic.Pointer 保护，避免并发测试触发 -race 误报。
+type stubRenderer struct {
+	lastSpecPtr atomic.Pointer[ErrorSpec]
+	called      atomic.Int32
+}
+
+func (s *stubRenderer) Render(c *Context, spec ErrorSpec) {
+	cp := spec
+	s.lastSpecPtr.Store(&cp)
+	s.called.Add(1)
+	c.JSON(spec.TransportStatus, gin.H{
+		"code": spec.Code,
+		"msg":  spec.DefaultMessage,
+	})
+}
+
+func (s *stubRenderer) lastSpec() ErrorSpec {
+	if p := s.lastSpecPtr.Load(); p != nil {
+		return *p
+	}
+	return ErrorSpec{}
+}
+
+// TestDefaultRendererPreservesLegacyEnvelope 验证未注入 renderer 时
+// 输出仍是历史 envelope {msg, status}——这是 issue 46 的向后兼容承诺。
+func TestDefaultRendererPreservesLegacyEnvelope(t *testing.T) {
+	w := httptest.NewRecorder()
+	ginCtx, _ := gin.CreateTestContext(w)
+	ctx := &Context{Context: ginCtx} // wk == nil 触发 fallback 路径
+	ctx.RenderError(ErrorSpec{
+		Code:            "err.shared.auth.required",
+		DefaultMessage:  "请先登录！",
+		TransportStatus: http.StatusUnauthorized,
+	})
+
+	if w.Code != http.StatusUnauthorized {
+		t.Fatalf("status = %d, want 401", w.Code)
+	}
+	var body map[string]any
+	if err := json.Unmarshal(w.Body.Bytes(), &body); err != nil {
+		t.Fatalf("body not json: %v", err)
+	}
+	if body["msg"] != "请先登录！" {
+		t.Errorf("msg = %v, want 请先登录！", body["msg"])
+	}
+	if int(body["status"].(float64)) != http.StatusUnauthorized {
+		t.Errorf("status field = %v, want 401", body["status"])
+	}
+}
+
+// TestSetErrorRendererRouted 验证注入 renderer 后 RenderError 走自定义路径。
+func TestSetErrorRendererRouted(t *testing.T) {
+	wk := New()
+	stub := &stubRenderer{}
+	wk.SetErrorRenderer(stub)
+
+	w := httptest.NewRecorder()
+	ginCtx, _ := gin.CreateTestContext(w)
+	ctx := &Context{Context: ginCtx, wk: wk}
+	ctx.RenderError(ErrorSpec{
+		Code:            "err.test.foo",
+		DefaultMessage:  "fallback",
+		TransportStatus: http.StatusTeapot,
+	})
+
+	if stub.called.Load() != 1 {
+		t.Fatalf("renderer not invoked")
+	}
+	if stub.lastSpec().Code != "err.test.foo" {
+		t.Errorf("spec.Code = %q", stub.lastSpec().Code)
+	}
+	if w.Code != http.StatusTeapot {
+		t.Errorf("status = %d, want 418", w.Code)
+	}
+}
+
+// TestSetErrorRendererInstanceScoped 验证两个 WKHttp 实例的 renderer 互不污染——
+// 这是 issue 46 强调的"实例方法而非全局"的核心保证，回归测试。
+func TestSetErrorRendererInstanceScoped(t *testing.T) {
+	a, b := New(), New()
+	stubA, stubB := &stubRenderer{}, &stubRenderer{}
+	a.SetErrorRenderer(stubA)
+	b.SetErrorRenderer(stubB)
+
+	wA := httptest.NewRecorder()
+	gA, _ := gin.CreateTestContext(wA)
+	(&Context{Context: gA, wk: a}).RenderError(ErrorSpec{Code: "x", TransportStatus: 400})
+
+	if stubA.called.Load() != 1 || stubB.called.Load() != 0 {
+		t.Fatalf("cross-instance leak: a=%d b=%d", stubA.called.Load(), stubB.called.Load())
+	}
+}
+
+// TestSetErrorRendererConcurrent 并发设置 renderer 与渲染不应触发数据竞争
+// （go test -race 时验证）。
+func TestSetErrorRendererConcurrent(t *testing.T) {
+	wk := New()
+	wk.SetErrorRenderer(&stubRenderer{})
+
+	var wg sync.WaitGroup
+	for i := 0; i < 20; i++ {
+		wg.Add(2)
+		go func() {
+			defer wg.Done()
+			wk.SetErrorRenderer(&stubRenderer{})
+		}()
+		go func() {
+			defer wg.Done()
+			w := httptest.NewRecorder()
+			g, _ := gin.CreateTestContext(w)
+			(&Context{Context: g, wk: wk}).RenderError(ErrorSpec{Code: "x", TransportStatus: 400})
+		}()
+	}
+	wg.Wait()
+}
+
+// TestDefaultRendererZeroTransportStatus 触发 TransportStatus=0 的回退分支——
+// 防止调用方遗漏字段时下游误判为 200 成功。
+func TestDefaultRendererZeroTransportStatus(t *testing.T) {
+	w := httptest.NewRecorder()
+	g, _ := gin.CreateTestContext(w)
+	(&Context{Context: g}).RenderError(ErrorSpec{Code: "x", DefaultMessage: "m"})
+	if w.Code != http.StatusBadRequest {
+		t.Errorf("status = %d, want 400 fallback", w.Code)
+	}
+}
+
+// altRenderer 是另一个 ErrorRenderer 具体类型；与 stubRenderer 同时存在用于验证
+// 注入不同具体类型不会触发 atomic 类型不变约束（这是把 atomic.Value 换成 boxed
+// atomic.Pointer 的核心动机）。
+type altRenderer struct{ tag string }
+
+func (a *altRenderer) Render(c *Context, spec ErrorSpec) {
+	c.JSON(spec.TransportStatus, gin.H{"renderer": a.tag, "code": spec.Code})
+}
+
+// TestSetErrorRendererSwitchConcreteTypes 验证 SetErrorRenderer 可以反复
+// 注入不同具体类型的 renderer 而不会 panic——atomic.Value 会在此处崩溃，
+// 这是 P0 修复 atomic.Value → atomic.Pointer[errorRendererBox] 的回归测试。
+func TestSetErrorRendererSwitchConcreteTypes(t *testing.T) {
+	wk := New()
+
+	// 第一次：stubRenderer
+	wk.SetErrorRenderer(&stubRenderer{})
+	w1 := httptest.NewRecorder()
+	g1, _ := gin.CreateTestContext(w1)
+	(&Context{Context: g1, wk: wk}).RenderError(ErrorSpec{Code: "a", TransportStatus: 400})
+
+	// 第二次：altRenderer（不同具体类型）—— atomic.Value 在此处 panic。
+	wk.SetErrorRenderer(&altRenderer{tag: "alt"})
+	w2 := httptest.NewRecorder()
+	g2, _ := gin.CreateTestContext(w2)
+	(&Context{Context: g2, wk: wk}).RenderError(ErrorSpec{Code: "b", TransportStatus: 400})
+
+	if !strings.Contains(w2.Body.String(), `"renderer":"alt"`) {
+		t.Errorf("second renderer not used: %s", w2.Body.String())
+	}
+
+	// 第三次：又换回 stubRenderer
+	stub3 := &stubRenderer{}
+	wk.SetErrorRenderer(stub3)
+	w3 := httptest.NewRecorder()
+	g3, _ := gin.CreateTestContext(w3)
+	(&Context{Context: g3, wk: wk}).RenderError(ErrorSpec{Code: "c", TransportStatus: 400})
+	if stub3.lastSpec().Code != "c" {
+		t.Errorf("third renderer not invoked")
+	}
+}
+
+// TestSetErrorRendererNilResetsToFallback 验证 SetErrorRenderer(nil) 后
+// RenderError 回退到 default fallback envelope，且后续可以再次注入。
+func TestSetErrorRendererNilResetsToFallback(t *testing.T) {
+	wk := New()
+	stub := &stubRenderer{}
+	wk.SetErrorRenderer(stub)
+
+	// 重置
+	wk.SetErrorRenderer(nil)
+
+	w := httptest.NewRecorder()
+	g, _ := gin.CreateTestContext(w)
+	(&Context{Context: g, wk: wk}).RenderError(ErrorSpec{
+		Code:            "err.x",
+		DefaultMessage:  "fallback msg",
+		TransportStatus: http.StatusUnauthorized,
+	})
+
+	if stub.called.Load() != 0 {
+		t.Errorf("stub renderer invoked after reset (called=%d)", stub.called.Load())
+	}
+	if w.Code != http.StatusUnauthorized {
+		t.Errorf("status=%d want 401", w.Code)
+	}
+	if !strings.Contains(w.Body.String(), "fallback msg") {
+		t.Errorf("body=%q, want fallback DefaultMessage", w.Body.String())
+	}
+
+	// 再注入：能恢复
+	stub2 := &stubRenderer{}
+	wk.SetErrorRenderer(stub2)
+	w2 := httptest.NewRecorder()
+	g2, _ := gin.CreateTestContext(w2)
+	(&Context{Context: g2, wk: wk}).RenderError(ErrorSpec{Code: "z", TransportStatus: 400})
+	if stub2.called.Load() != 1 {
+		t.Errorf("re-injected renderer not invoked")
+	}
+}
--- a/pkg/wkhttp/http.go
+++ b/pkg/wkhttp/http.go
@@ -4,8 +4,8 @@ import (
 	"errors"
 	"net/http"
 	"strconv"
-	"strings"
 	"sync"
+	"sync/atomic"
 
 	"github.com/Mininglamp-OSS/octo-lib/pkg/cache"
 	"github.com/Mininglamp-OSS/octo-lib/pkg/log"
@@ -28,6 +28,14 @@ const (
 type WKHttp struct {
 	r    *gin.Engine
 	pool sync.Pool
+
+	// errorRenderer 存 *errorRendererBox（包装 ErrorRenderer 接口）。
+	// 用 atomic.Pointer 而非 atomic.Value 的原因：atomic.Value 要求每次 Store 的
+	// 动态类型一致，注入不同具体类型的 renderer 会 panic；测试与多模块场景
+	// 下都需要支持任意实现切换，因此与 TokenParser 一致采用 boxed atomic.Pointer。
+	errorRenderer atomic.Pointer[errorRendererBox]
+	// tokenParser 存 *tokenParserBox（包装 TokenParser 接口）。同上。
+	tokenParser atomic.Pointer[tokenParserBox]
 }
 
 // New New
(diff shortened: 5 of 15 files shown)


lml2468 (approved):
## ✅ APPROVED

Thorough review of all 15 files (+1650 -21). This is a well-designed framework-layer PR.

### Architecture Assessment

**Design decisions are sound:**

1. **Instance-scoped hooks** (`SetErrorRenderer` / `SetTokenParser` on `*WKHttp`, not package-global) — correct for test isolation and multi-instance scenarios. The regression test `TestSetErrorRendererInstanceScoped` validates this.

2. **`atomic.Pointer[box]` pattern** — correctly avoids `atomic.Value`'s same-type constraint. The test `TestSetErrorRendererSwitchConcreteTypes` is the P0 regression guard. Box indirection is the right idiom here.

3. **Backward compatibility via fallback defaults** — `defaultErrorRenderer` emits the exact historical `{msg, status}` envelope; `legacyTokenParser` preserves cache-based "uid@name@role" parsing. Upgrading downstream services to this lib version is zero-risk if they don't inject anything.

4. **`RateLimitMiddleware` → instance method** — necessary for accessing the instance-scoped `ErrorRenderer`. The signature break is documented and has a clear migration path (~5 call sites in octo-server).

5. **Lua token bucket script** — well-designed: skip-write on deny path avoids DDoS write amplification, first-access exception handles key initialization, TTL is `fill_time * 2` for natural GC.

6. **fail-open + throttled degradation warnings** — `logDegrade` with `atomic.Int64` + CAS is a clean alternative to `sync.Once` that doesn't permanently silence after first occurrence.

7. **`SecureCORSOverrideMiddleware` returns `gin.HandlerFunc` not `HandlerFunc`** — intentional: it manipulates headers only, doesn't need `RenderError`, and must interleave with other gin-level middleware. Correct choice.

### Test coverage

20+ tests including:
- Race condition (`-race`) coverage
- Concurrent renderer swap
- nil-reset + re-injection
- Legacy envelope preservation
- Custom parser sentinel mapping + fallthrough
- Redis-gated rate-limit E2E (skip when unavailable)

### Non-blocking observations

1. **`ratelimit_test.go` reimplements `strings.Contains`** as a local `contains()` helper — trivial, but `strings` is already used in other test files in the same package.

2. **`CheckLoginRole` / `CheckLoginRoleIsSuperAdmin` left un-migrated** — the rationale in the PR description is solid (returns error to business layer, sparse 403 path, would touch every call site). Correct deferral.

3. **CORS preflight ordering limitation** (documented pre-existing) — the plan to merge `CORSMiddleware` + `SecureCORSOverrideMiddleware` into a single whitelist-aware handler is the right follow-up. Not this PR's scope.

4. **CI `check-sprint` failure** is administrative (sprint label not set), not a code issue.

No blocking issues found.

lml2468 (dismissed):
[APPROVE]

**Verdict: APPROVED** ✅

Thorough, well-architected PR. +1650 lines but the diff reads cleanly: new primitives are sharply scoped, backward compatibility is preserved with rigorous fallback paths, and test coverage is comprehensive (including race, concurrency, and instance-isolation tests). The PR body itself is exemplary documentation.

---

## What I verified

1. **Backward compatibility** — `defaultErrorRenderer.Render` emits the exact historical `{msg, status}` envelope; `legacyTokenParser` reproduces the old `cache.Get → uid@name[@role] → split` flow; `CORSMiddleware` adds headers without removing any existing ones. All three abort branches in `AuthMiddleware` produce identical output when no renderer is injected (`TestAuthMiddlewareLegacyFallback`).

2. **`atomic.Pointer[box]` over `atomic.Value`** — correct choice. `atomic.Value` requires type-consistent stores; injecting different concrete `ErrorRenderer`/`TokenParser` implementations across tests or multi-tenant wiring would panic. `TestSetErrorRendererSwitchConcreteTypes` is the right regression test for this.

3. **Rate-limit Lua script** — token-bucket logic is correct: refill calculation, TTL = `ceil(fill_time * 2)`, deny-path write-skip to avoid DDoS write amplification, defensive `rate <= 0` guard. The `state[2] == false` first-access edge case is handled properly.

4. **Fail-open design** — Redis unavailability → allow + throttled warning. This is the right call for an IM system where Redis outage already means degraded service; rate-limiting adding a 503 on top would amplify the blast radius. `logDegrade` with CAS-based throttling prevents log storms.

5. **CORS changes** — three new `Allow-Headers` (`X-Octo-Error-Envelope`, `X-Octo-Lang`, `Accept-Language`) and new `Expose-Headers` (`Content-Language`, `Vary`) are additive. Tests verify both.

6. **`SecureCORSOverrideMiddleware` preflight limitation** — pre-existing, documented, not introduced by this PR. Tracking issue promised.

7. **CI** — Build ✅ Test ✅ Vet ✅. `check-sprint` failure is a project-board label issue (Sprint W22 not set), unrelated to code.

---

## Suggestions (non-blocking)

### S1: `SetTokenParser(nil)` style inconsistency

```go
// SetTokenParser stores typed nil:
l.tokenParser.Store((*tokenParserBox)(nil))

// SetErrorRenderer stores untyped nil:
l.errorRenderer.Store(nil)
```

Both are functionally equivalent for `atomic.Pointer[T]`, but the inconsistent style could confuse future readers. Consider aligning on one form.

### S2: `contains()` in `ratelimit_test.go` reimplements `strings.Contains`

```go
func contains(s, sub string) bool { ... }
```

`strings.Contains` does exactly this and is already used elsewhere in the package.

### S3: `SecureCORSOverrideMiddleware` returns `gin.HandlerFunc` vs `CORSMiddleware` returns `HandlerFunc`

This is an intentional design choice (documented in the PR body), but a one-line code comment at the function signature noting the reason (must compose with raw gin middleware chains) would help future readers.

---

LGTM. Solid engineering work 🚢

Jerry-Xin (dismissed):
The PR is in scope for `octo-lib` and the `pkg/wkhttp` changes are relevant to the library.

💬 Non-blocking

🟡 Warning: `pkg/wkhttp/ratelimit.go:180` trusts `X-Real-Ip` and `X-Forwarded-For` directly. If a service is ever reachable without a trusted proxy sanitizing those headers, clients can spoof IPs to bypass per-IP rate limits. Consider documenting that requirement clearly or deriving client IP through trusted-proxy configuration.

🟡 Warning: `pkg/wkhttp/error_renderer.go:58` adds a `status` field to the default auth error body. The previous hardcoded auth aborts returned only `msg`, so this is not byte-for-byte wire compatible despite the compatibility wording. Likely acceptable, but the PR description/tests should avoid claiming exact identity.

🔵 Suggestion: `pkg/wkhttp/http.go:307` short-circuits empty `token` headers before invoking a custom `TokenParser`. That means injected parsers cannot support alternate token sources such as `Authorization` or cookies. If this is intentional, document that the hook parses only the legacy `token` header value.

🔵 Suggestion: `pkg/wkhttp/secure_cors_test.go:82` does not assert the denied-origin header removal in the realistic middleware chain. Add an integration test with `CORSMiddleware()` followed by `SecureCORSOverrideMiddleware(...)` for non-preflight allowed and denied origins.

✅ Highlights

The renderer and token parser injection are instance-scoped and race-safe, with good regression coverage around concrete type switching.

Auth success now preserves legacy `c.Set` values while adding context-based `UserInfo`, which gives downstream services a gradual migration path.

Verified locally: `go test ./pkg/wkhttp`, `go test ./...`, `go vet ./...`, and `go test -race -count=1 ./pkg/wkhttp/` all pass.

yujiawei (dismissed):
# Code Review — PR #47 (octo-lib)

## Summary

The PR introduces `ErrorRenderer` / `TokenParser` injection hooks in `pkg/wkhttp`, migrates `AuthMiddleware` / `CORSMiddleware` to use them for i18n, and consolidates `ratelimit.go` / `secure_cors.go` from octo-server into this library. Backward compatibility is preserved by a `defaultErrorRenderer` and `legacyTokenParser` that mirror the historical behavior when no renderer/parser is injected.

I built the package, ran `go vet`, `go test -count=1` and `go test -race -count=1` on the head SHA — all clean (Redis-gated rate-limit tests skip locally as documented).

## Verdict

No P0/P1 issues. The design is deliberate, the docstrings explain non-obvious choices (`atomic.Pointer[box]` vs `atomic.Value`, fail-open vs fail-closed semantics, deployment-level CORS preflight ordering), and the test matrix covers the migrated code paths plus the regression that motivated the boxed-pointer choice. Items below are nits / suggestions / things a human deployer should double-check given the security-sensitive label — none block merge.

## Detailed findings

### P2 — `defaultErrorRenderer` envelope is a superset of the prior auth abort body (not strict equality)

The PR description states the default fallback "emits the same `{msg, status}` envelope today's hardcoded paths produce". For `AuthMiddleware`, that is not literally true: the prior implementation in `pkg/wkhttp/http.go` (main) called

```go
c.AbortWithStatusJSON(http.StatusUnauthorized, gin.H{"msg": "..."})
```

with **no `status` field** in the body. The new `defaultErrorRenderer.Render` in `pkg/wkhttp/error_renderer.go:53-62` always writes `{"msg": ..., "status": ...}`. So the new fallback is a superset (extra `status` field) rather than byte-equal to the old one. Other paths that already used `ResponseError` did emit `status`, so the inconsistency is now resolved — this is just worth flagging because the PR description overstates the equivalence and downstream clients that key off exact body shape (rare, but possible) will see a new field.

Suggestion: tighten the wording in the PR description, or note in the `defaultErrorRenderer` doc comment that the auth-abort path historically omitted `status` and now includes it.

### P2 — `getClientIP` trusts `X-Real-Ip` first; correctness is a deployment contract

`pkg/wkhttp/ratelimit.go:180-194` reads `X-Real-Ip` before walking `X-Forwarded-For`. The docstring is honest that this assumes single-layer Tencent CLB with `pass-to-target`. In that topology CLB does set `X-Real-IP` and the assumption holds. But if any future deployment puts an extra hop in front (CDN, Kubernetes ingress, nginx proxy chain), or if the ingress controller is reconfigured to forward the client-supplied `X-Real-Ip` instead of overwriting it, an attacker can spoof the limiter key per request — the Redis bucket is then partitioned by attacker-controlled value and the strict per-IP limiters become bypassable.

This is a `needs-human-review` security item rather than a code defect: the code matches the documented topology, the comment correctly flags the assumption, and there is no evidence the topology has changed. I'd recommend that whoever merges this PR confirms (a) Tencent CLB still terminates HTTP at L7 and overwrites `X-Real-Ip`, and (b) there is no CDN or extra reverse-proxy layer in front of the relevant services. If either assumption later changes, this function becomes a real CVE.

### P2 — `tokenBucketScript` early-return on `rate <= 0` is fail-closed; the rest of the layer is fail-open

`pkg/wkhttp/ratelimit.go:37` returns `{0, 0, 1}` (deny + retry-after 1s) when `rate <= 0`. That is fail-CLOSED for misconfiguration, which contradicts the function-level fail-open contract documented above (Redis errors fail open, by design). In practice this only fires if a caller passes `rps <= 0`, which would be a config bug; but if `ParseRPSFromEnv` ever returns 0 because of an env-var rename, every request to that route is rejected with a 429.

Suggestion: either (a) validate `rps > 0` in `newKeyedLimiter` and panic at construction (loud, fast), or (b) make the Lua early-return fail-open with a logged warning. Today's behavior is acceptable but inconsistent.

### P2 — `RateLimitMiddleware` discards the `degraded` flag from `kl.allow`

`pkg/wkhttp/ratelimit.go:243` and the two sister middlewares ignore the fourth return value (`degraded`). When Redis is unreachable the call returns `allowed=true, remaining=burst, retryAfter=0, degraded=true`, and `setRateLimitHeaders` then writes `X-RateLimit-Remaining=<burst>` to every response while the limiter is silently disabled. That's misleading from an observability standpoint — clients that key off these headers (or operators reading them in logs) cannot tell the difference between "you have full quota" and "the limiter is degraded".

Suggestion: when `degraded=true`, either skip emitting the rate-limit headers entirely, or emit a distinct `X-RateLimit-Scope` value (e.g. `ip:degraded`) so dashboards can flag the state. Not blocking — `logDegrade` already records the failure server-side.

### P2 — `SetTokenParser` and `SetErrorRenderer` use slightly different nil-store idioms

`SetErrorRenderer` does `l.errorRenderer.Store(nil)` (untyped nil) at `http.go:58`, while `SetTokenParser` does `l.tokenParser.Store((*tokenParserBox)(nil))` at `token_parser.go:22`. Both are functionally equivalent for `atomic.Pointer[T]` — the Load returns a nil pointer either way and the consumer's `if box == nil` check handles it. Nit: pick one style for symmetry.

### Nit — `legacyTokenParser.Parse` ErrTokenMissing branch is unreachable through `AuthMiddleware`

`pkg/wkhttp/token_parser.go:70-72` returns `ErrTokenMissing` when `strings.TrimSpace(token) == ""`, but `AuthMiddleware` (`http.go:307-317`) already short-circuits on `token == ""` before invoking the parser. The branch survives only as defense-in-depth for callers who use `legacyTokenParser` directly; harmless, but worth a one-line comment to that effect, otherwise readers may go looking for the call path that triggers it.

### Nit — `info.Name` may be the empty string when cache value is `"uid@"`

`legacyTokenParser.Parse` accepts `len(parts) >= 2` even when `parts[1] == ""`, so a malformed cache entry like `"uid@"` parses as `UserInfo{UID: "uid", Name: ""}` and the middleware will still `c.Set("name", "")`. The previous implementation had identical behavior, so this is preserved-by-design. Just calling it out so it is not surprising under audit.

### Acknowledged limitations from PR description (not findings)

- `SecureCORSOverrideMiddleware` cannot fix preflight ordering because `CORSMiddleware` aborts `OPTIONS` upstream with status 204 — pre-existing in octo-server, deployments terminate CORS at CLB/nginx so the application-layer middleware is a fallback. Reasonable to defer a fix to a separate PR.
- `SharedUIDRateLimiter` not moved (would create a reverse dependency). The follow-up rewiring in octo-server is documented in the PR body.
- `RateLimitMiddleware` signature changed from a package function to a `*WKHttp` method. Required for `c.RenderError` plumbing; documented; downstream octo-server adoption is explicit (~5 call sites).
- `CheckLoginRole` / `CheckLoginRoleIsSuperAdmin` deliberately left with hardcoded Chinese strings — translation is the calling service's responsibility. Reasonable per issue #46 §4 option 2.

## Test coverage observations

- `TestSetErrorRendererSwitchConcreteTypes` is the right regression test for the `atomic.Value → atomic.Pointer[errorRendererBox]` fix and is well-targeted.
- `TestAuthMiddlewareLegacyFallback` asserts `body["msg"]` and `w.Code`, but does not pin the body to exclude the new `status` field. If strict wire-format equality is a goal, consider adding an explicit assertion (or removing the claim from the PR description, see P2 above).
- `TestRateLimitMiddleware*` tests are gated on a local Redis at `127.0.0.1:6399` and `t.Skip` otherwise. CI must run the testenv Redis container or these branches are not exercised. Worth confirming the CI workflow has Redis in scope before relying on these for regression detection — mention this to whoever owns CI.
- The fail-open path in `keyedLimiter.allow` (Redis unreachable → `allowed=true, degraded=true`) has no unit test. Adding one would let us catch a future refactor that accidentally changes the fail-mode.

## Security-sensitive checklist (per `needs-human-review` label)

A human reviewer should confirm before merge:

1. Tencent CLB still single-layer in front of all services using `RateLimitMiddleware` / `StrictIPRateLimitMiddleware`, and still overwrites `X-Real-Ip` (see P2 above).
2. No environment configuration exists that could pass `rps=0` to `newKeyedLimiter` (see P2 on Lua early-return).
3. The downstream octo-server PR that deletes its local copies of `ratelimit.go` / `secure_cors.go` lands together with this one, so there is one source of truth and no drift between the two trees.
4. The `err.shared.*` codes used here (`auth.token_missing`, `auth.required`, `auth.token_invalid`, `rate.limited`) are registered in the consumer's i18n catalog before the consumer injects a non-default renderer in production.

Otherwise, LGTM — the new primitives are minimal and the consolidation is a clear win.

lml2468 (approved):
## ✅ Re-APPROVED (6ed4b8633a05)

Delta from 2e5470b → 6ed4b86 is a single chore commit addressing collective review feedback. No logic changes:

1. `token_parser.go` — `SetTokenParser(nil)` now stores untyped `nil`, aligned with `SetErrorRenderer` style ✓
2. `ratelimit_test.go` — custom `contains()` replaced with `strings.Contains` ✓
3. `secure_cors.go` — added godoc explaining `gin.HandlerFunc` return type intent + preflight follow-up note ✓
4. `ratelimit.go` — documented `getClientIP` trusted-proxy assumption (deployment-must-guarantee) ✓

All four address non-blocking feedback from first round. Architecture assessment unchanged.

Jerry-Xin (approved):
The PR is in scope for `octo-lib` and the `wkhttp` changes are cohesive. I found no merge-blocking correctness or security issues.

💬 Non-blocking

🟡 Warning: `SecureCORSOverrideMiddleware` only works reliably if no later handler/middleware rewrites CORS headers. It mutates headers before `c.Next()` at `pkg/wkhttp/secure_cors.go:123`, so downstream code can still reintroduce `Access-Control-Allow-Origin: *` or credentials headers. The tests at `pkg/wkhttp/secure_cors_test.go:61` and `pkg/wkhttp/secure_cors_test.go:86` explicitly avoid asserting deletion in those cases. Consider tightening the documented ordering contract or adding a combined whitelist-aware CORS middleware later.

🟡 Warning: `CORSMiddleware` still emits `Access-Control-Allow-Origin: *` with `Access-Control-Allow-Credentials: true` and short-circuits `OPTIONS` before the secure override can run at `pkg/wkhttp/http.go:430` and `pkg/wkhttp/http.go:436`. The PR calls this out as pre-existing, so I would not block on it, but it remains important for services relying on application-layer credentialed CORS.

🔵 Suggestion: The new public rate-limit methods accept invalid `rps`/`burst` values without Go-side validation at `pkg/wkhttp/ratelimit.go:224`, `pkg/wkhttp/ratelimit.go:280`, and `pkg/wkhttp/ratelimit.go:320`. The Lua script avoids crashing for `rate <= 0`, but misconfiguration can become “deny all” behavior. A defensive clamp or constructor error would make this library API harder to misuse.

✅ Highlights

The renderer and parser hooks are instance-scoped and race-safe, and the `atomic.Pointer` box approach avoids the concrete-type panic risk that `atomic.Value` would introduce.

Auth middleware preserves legacy `c.Set(...)` behavior while adding `UserInfo` on `context.Context`, which gives downstream services a clean migration path.

Verification run locally: `go test ./...`, `go vet ./...`, and `go test -race -count=1 ./pkg/wkhttp` all pass.

lml2468 (dismissed):
## ✅ Re-APPROVED (71de1cce1658)

Force-push rebase — same content as previously reviewed 6ed4b86. No new code changes. Architecture assessment unchanged. Re-approving to maintain review on latest HEAD.

Jerry-Xin (dismissed):
Summary: The PR is in scope for `octo-lib` and the `pkg/wkhttp` changes are cohesive; I found no blocking correctness or security issues.

**💬 Non-blocking**
- 🟡 Warning: [pkg/wkhttp/ratelimit.go:107](/Users/mlamp/.openclaw/tmp/pr-review-g4dalno6/repo/pkg/wkhttp/ratelimit.go:107), [pkg/wkhttp/ratelimit.go:224](/Users/mlamp/.openclaw/tmp/pr-review-g4dalno6/repo/pkg/wkhttp/ratelimit.go:224)  
  The public rate-limit constructors assume positive `rps` and `burst`, but do not enforce it. If a caller bypasses `ParseRPSFromEnv` / `ParseBurstFromEnv` or passes a bad default, `rps <= 0` or `burst <= 0` can effectively deny traffic. Consider validating in `newKeyedLimiter` or the middleware constructors.

- 🔵 Suggestion: [pkg/wkhttp/secure_cors_test.go:61](/Users/mlamp/.openclaw/tmp/pr-review-g4dalno6/repo/pkg/wkhttp/secure_cors_test.go:61), [pkg/wkhttp/secure_cors.go:123](/Users/mlamp/.openclaw/tmp/pr-review-g4dalno6/repo/pkg/wkhttp/secure_cors.go:123)  
  The secure CORS tests do not assert header removal for the realistic “upstream CORS writes first, override middleware rewrites next” ordering. Adding one test with a preceding middleware that sets CORS headers would better lock down the security-sensitive behavior.

**✅ Highlights**
- `ErrorRenderer` and `TokenParser` are instance-scoped and race-safe, with good regression coverage for renderer replacement.
- Auth middleware preserves legacy context keys while adding structured `UserInfo` propagation.
- The rate-limit implementation is atomic at Redis level and fail-open behavior is documented.
- Local verification passed: `go test ./pkg/wkhttp`, `go test ./...`, and `go vet ./...`.

yujiawei (dismissed):
# Code Review — PR #47 (octo-lib)

## Summary

This PR introduces `pkg/wkhttp` primitives (`ErrorRenderer`, `TokenParser`, `UserInfo`) for cross-service i18n, migrates rate limiting and secure CORS override from octo-server, and refactors `AuthMiddleware` to route the three legacy hardcoded Chinese errors through the new renderer. The implementation is careful, well-commented, and backward compatible. Tests cover the regression-sensitive paths (renderer hot-swap, instance isolation, legacy fallback envelope, sentinel-based error mapping).

I have reviewed every changed file at HEAD `71de1cce`. No P0/P1 blockers found. A handful of P2 observations are listed below; none of them gate this PR.

## What works well

- **`atomic.Pointer[box]` instead of `atomic.Value`.** The design choice is correct — `atomic.Value` panics when the dynamic type changes between `Store` calls, which would surface as a runtime crash the first time a service swaps to a different concrete renderer (e.g. test fixture vs production wiring). The boxed pointer pattern + `TestSetErrorRendererSwitchConcreteTypes` regression test is the right call. Same reasoning applied consistently to `tokenParser`.
- **Instance-scoped renderer/parser** (`atomic.Pointer` lives on `*WKHttp`, not as a package global). `TestSetErrorRendererInstanceScoped` codifies the cross-test isolation guarantee, which matters for parallel test execution.
- **Backward-compatible default envelope.** `defaultErrorRenderer` preserves the existing `{msg, status}` shape so services that do not opt into `SetErrorRenderer` see no behavioral change. The `TransportStatus == 0` fallback to 400 is a sensible defensive guard.
- **Sentinel errors (`ErrTokenMissing` / `ErrTokenNotFound` / `ErrTokenInvalid`) exported with `errors.Is` semantics.** Custom parsers can opt into the same i18n key mapping without re-stringifying error messages. The default fallback to `err.shared.auth.required` for unrecognized errors is good — it avoids leaking internal error detail through the response body.
- **Token-bucket Lua script.** Atomic, correct math (`filled = min(burst, tokens + delta * rate)`), TTL self-managed by setting `EXPIRE` to `2 * fill_time`. The deny-path write-skip optimization is sound: `(filled, tokens, ts)` is a pure function of stored state and `now`, so skipping the writeback under sustained denial is correctness-preserving and DDoS-friendly. The first-touch exception (`state[2] == false`) correctly initializes the key + TTL even when the very first request is denied (`burst < 1`).
- **`logDegrade` throttle** uses CAS on `lastDegradeWarn` rather than `sync.Once` — failures are not silently dropped after the first warning, which is the right tradeoff for an infra-availability signal.
- **`fail-closed` (unknown IP) vs `fail-open` (Redis down).** The split is clearly documented and tested by the IP fallback path. UID-rate-limit fail-open on missing UID is an explicit (and correctly documented) tradeoff: it is a defense-in-depth layer that *must* be paired with global per-IP limiting and proper middleware ordering. The docstring is unusually clear about this — good.
- **`SecureCORSOverrideMiddleware` design choices.** Using `gin.HandlerFunc` rather than `wkhttp.HandlerFunc` is correct since this middleware needs to mix with non-wkhttp upstream CORS middlewares (e.g. dmwork-lib). `addVaryOrigin` is correctly idempotent and tolerant of both single-header-multi-value and multi-header value formats.
- **Comments are dense and explain *why*, not *what*.** Multiple non-obvious decisions (atomic.Pointer over atomic.Value, Lua deny-path write-skip, fail-open semantics, `SecureCORSOverrideMiddleware` ordering requirement) are documented inline. Future maintainers will not have to re-derive these.

## Findings

### P2 — Trust assumption around `X-Real-Ip` / `X-Forwarded-For` header (security-sensitive, documentation-only follow-up suggested)

`getClientIP` (`pkg/wkhttp/ratelimit.go:186`) trusts `X-Real-Ip` first, then rightmost `X-Forwarded-For`, then `RemoteAddr`. The docstring already calls out the assumption: the deployment must terminate behind a trusted reverse proxy that strips client-supplied `X-Real-Ip` / overwrites `X-Forwarded-For`. Tencent CLB in `pass-to-target` mode usually does this, but it is configurable, and a misconfigured deployment would let a client trivially bypass per-IP rate limits and per-IP strict limits by spoofing `X-Real-Ip`.

This is acknowledged in the code comment; not a blocker. Suggestion for a follow-up:

- Optionally accept an injected `TrustedProxyChecker` (or a list of trusted CIDRs) and only honor `X-Real-Ip` / `X-Forwarded-For` when `RemoteAddr` falls within that set. octo-server / downstream services can pass their own checker; default could be "trust anything" preserving today's behavior. This would let misconfigured environments (e.g. accidental host-network exposure during dev) fail loudly rather than silently.

Not actionable in this PR. Worth tracking as an issue.

### P2 — Auth middleware envelope wire-format change (already disclosed in PR body)

The three pre-existing `AuthMiddleware` abort branches previously emitted only `{msg}`. After this PR they go through the default renderer which emits `{msg, status}`. The PR body documents this explicitly. Behaviorally compatible for any client reading only the `msg` field; would break only a client that strict-schema-validates `{msg}` exclusively. Acceptable, but worth flagging in release notes.

### P2 — Test gaps

- `TestRateLimitMiddlewareDeniesViaRenderer` / `TestRateLimitMiddlewareDefaultFallback` are gated on `127.0.0.1:6399` Redis being available (`t.Skipf` otherwise). The fail-open path — i.e. Redis goes down → `kl.allow` returns `(true, burst, 0, true)` and the request is unconditionally allowed with a throttled warn log — is not covered by any test. Suggest adding a unit test that injects a stub `*rd.Client` whose `Script.Run` returns an error, asserting:
    1. Request is allowed.
    2. `X-RateLimit-*` headers are written sensibly (or omitted, depending on the documented contract — currently they *are* written, which is fine).
    3. `logDegrade` throttle behaves as documented across multiple consecutive failures.

  Not a blocker (the path is short and well-commented), but the fail-open behavior is the most consequential at runtime and currently relies on review reading rather than CI assertion.

- `TestSecureCORSOverrideMiddleware "no origin"` and `"origin not allowed"` subtests do not actually verify that the override middleware deletes `Access-Control-Allow-Origin` / `Access-Control-Allow-Credentials`. The test fixture has the handler write `*` after the middleware runs, so the assertion is effectively limited to "Vary contains Origin". The comments in the test acknowledge this, but consider replacing the fixture handler with a no-op to make the actual override behavior assertable. Otherwise the test does not protect against a regression where the middleware silently stops deleting the headers.

### P2 — Minor: `Vary: Origin` not added when `Origin` header is missing

`SecureCORSOverrideMiddleware` only calls `addVaryOrigin` when `origin != ""`. Strictly, the response *does* vary by `Origin` (because the headers it sets/deletes depend on `Origin`), so `Vary: Origin` would be useful for downstream caches even on no-origin requests. Cosmetic — extremely unlikely to matter in production where there is no shared cache between an authenticated origin and an anonymous one for the same URL, but the RFC-correct version is to always advertise `Vary`.

### P2 — `CheckLoginRole` / `CheckLoginRoleIsSuperAdmin` deferred (already disclosed)

These two methods continue to return hardcoded Chinese `errors.New(...)` strings. PR body documents the rationale (option 2 from #46 §4: translation responsibility lives in the service). Reasonable scope decision; downstream services need to make sure they map these via `errors.Is` rather than `err.Error()` string comparison so the eventual lib-side migration can rename them safely.

### P2 — `RateLimitMiddleware` signature break (already disclosed)

Package functions → instance methods on `*WKHttp`. The PR body lists the call-site change required in octo-server. Worth confirming during the octo-server cleanup PR that all call sites (the body claims ~5) are converted, ideally with a `golangci-lint` rule or grep before merge.

### P2 — Maintainability nits (non-blocking)

- `errorRendererBox` and `tokenParserBox` are nearly identical structurally (`r ErrorRenderer` / `p TokenParser`). Not worth a generic `box[T]` (Go generics + atomic.Pointer interaction is awkward and the two box types only exist twice). Leave as is.
- `ErrorSpec.Internal` field is declared but unused in this PR. Either drop it for now or document a concrete planned use case in the comment so it doesn't become dead-state. Minor.
- `defaultErrorRenderer` writes `c.JSON(...)` but does not call `c.Abort()`. The contract is documented (callers `c.Abort()` themselves), and every call site in this PR follows it. Worth making the contract more discoverable in the `ErrorRenderer` interface comment, since custom `Render` implementations that *do* call `Abort` themselves would double-abort harmlessly but inconsistently. Documentation-only.

### Cross-file dependency check

I explicitly verified the following cross-file invariants, since reasoning about middleware ordering is the kind of place review-by-glance misses things:

1. `Context.wk` is set in `WKHttpHandler` *before* `handlerFunc(hc)` runs (`http.go:212`), and reset to `nil` in `Context.reset()` before `pool.Put` (`http.go:213` defer). `RenderError` correctly handles `wk == nil` by falling back to `defaultErrorRenderer` — both `TestDefaultRendererPreservesLegacyEnvelope` (constructs `&Context{Context: ginCtx}` with `wk == nil`) and the production pool path are covered. Pool reuse + concurrent requests are safe because the renderer is read from the `*WKHttp`-scoped atomic pointer, not from per-`Context` state.
2. `legacyTokenParser` reuses the existing cache contract: `cache.Get` of a missing key returns `("", nil)` for `fakeCache`, mapped to `ErrTokenNotFound`. Production cache implementations should be re-verified for the same semantics (i.e. cache-miss should not return a non-nil error other than a sentinel "not found"); if any production cache returns `redis.Nil` or similar, the legacy parser would map it to `ErrTokenNotFound` via the `err != nil` branch — same `err.shared.auth.required` outcome — but the `||` short-circuit hides that path from logs. Fine for behavior, worth a code comment if the cache.Cache contract is not crystal clear.
3. `SecureCORSOverrideMiddleware` ordering: registered AFTER upstream `CORSMiddleware`, so in the gin chain `corsMW → secureCORSMW → handler` the override's header writes happen *after* the upstream's writes (last-write-wins on `c.Writer.Header()` before any response is flushed). This matches the docstring. The pre-existing `OPTIONS` short-circuit at `CORSMiddleware` happens *before* `secureCORSMW` runs, which is the limitation the PR body explicitly defers. Behavior is correct given the documented constraint.

## Verdict

No P0/P1 blockers. The design choices around atomic.Pointer, fail-open rate limiting, instance-scoped renderer/parser, and the legacy envelope preservation are all defensible and well-tested. The trust assumption around `X-Real-Ip` / `X-Forwarded-For` is the most security-sensitive item but is explicitly documented in code; suggested follow-up is non-blocking. Recommend approving and merging; track the test-gap and trusted-proxy hardening as follow-ups.

lml2468 (approved):
## ✅ Re-APPROVED (ff5535903470)

Incremental commit `ff55359` addresses Jerry-Xin's non-blocking feedback:

1. **`newKeyedLimiter` panics on `rps<=0` or `burst<=0`** — correct choice. Rate-limit middleware is created once at startup; invalid config would cause 100% 429 in production. Fail-loud at boot is far better than silent runtime failure. The Lua script's `if rate <= 0` remains as a last-ditch defense for defense-in-depth.

2. **`token_parser.go`** — comment clarification on `legacyTokenParser`.

3. **`ratelimit_test.go`** — test coverage for the new validation.

No logic regressions. Architecture assessment unchanged.

yujiawei (approved):
# Code Review — PR #47 (octo-lib)

## Summary

Reviewed the full diff (15 files, +1692/-21) and ran `go build ./...`, `go vet ./pkg/wkhttp`, and `go test -race -count=1 ./pkg/wkhttp/` locally on `ff55359`. All clean; non-Redis tests pass under the race detector (local coverage 53.0% — the gap to the body's claimed 66.1% is the gated `TestRateLimitMiddlewareDeniesViaRenderer` / `…DefaultFallback`, which correctly `t.Skipf` without testenv Redis).

The PR cleanly accomplishes three goals:

1. Introduces `ErrorRenderer` / `TokenParser` injection points on `*WKHttp` so downstream services can layer i18n on top of `wkhttp` without forking.
2. Migrates the three hardcoded Chinese abort branches in `AuthMiddleware` and the rate-limit / CORS abort paths to the new `RenderError` channel, with a default fallback that preserves the historical `{msg, status}` envelope verbatim.
3. Consolidates `ratelimit.go` / `secure_cors.go` (and the pure helpers from `ratelimit_helper.go`) into octo-lib so octo-server stops carrying drift-prone copies.

The design is correct, the trade-offs are documented loudly (PR body, godoc, and inline comments), and the tests directly cover the load-bearing invariants — including the concurrency regression for the `atomic.Value → atomic.Pointer[box]` swap and the cross-instance isolation guarantee. No P0 or P1 issues found.

## 1. Correctness review

### `AuthMiddleware` — three abort branches

- `pkg/wkhttp/http.go:303-339` cleanly mirrors the historical decision tree:
  - empty header → `ErrTokenMissing` / `err.shared.auth.token_missing`
  - cache miss / empty value → `ErrTokenNotFound` / `err.shared.auth.required`
  - fewer than two `@`-separated parts → `ErrTokenInvalid` / `err.shared.auth.token_invalid`
- `authErrorSpec` (`http.go:344-376`) maps via `errors.Is`, with the unknown-error default collapsing to `err.shared.auth.required` so a custom parser cannot leak internal failure detail through the response. Good defensive choice.
- Success path correctly populates both `c.Set("uid"|"name"|"role", …)` (legacy callers) and `WithUser(c.Request.Context(), info)` (new context-channel callers); existing call sites such as `GetLoginUID()` / `GetLoginRole()` continue to work without change.
- Default fallback (`error_renderer.go:53-62`) emits `{msg, status}` with a 400 floor when `TransportStatus == 0`. The PR body already calls out the new `status` field on the three auth abort branches; the change is intentional alignment with the rest of `ResponseError*`.

### `legacyTokenParser` parity

- `token_parser.go:69-89` reproduces the prior `strings.Split` semantics (`< 2` parts → invalid, third element optional). One subtle but harmless behavioural shift: with `info.Role == ""` the new code skips `c.Set("role", "")`, while the old code set the empty string when a third part was present (e.g. `uid@name@`). Observable through `c.GetString("role")` it is identical (both return `""`); only `_, ok := c.Get("role")` would now report `ok=false`. No call site in this repo or in octo-server's `pkg/wkhttp` consumers relies on key presence rather than value, so this is a non-issue in practice. Worth keeping in mind for downstream review.
- `Parse` accepts `ctx` purely for forward compatibility with a `context`-aware `cache.Cache`; the `_` parameter is honest about today's reality.

### Rate limit (`pkg/wkhttp/ratelimit.go`)

- The Lua token-bucket script is atomic-by-construction (single `EVAL`), and the deny-path write skip is correctly gated by `state[2] == false` so brand-new keys still get a TTL when the very first request is denied (`burst<1` cannot occur — see panic in `newKeyedLimiter`, but the script remains independently safe).
- `newKeyedLimiter` panics on `rps<=0` / `burst<=0`. Loud-fast at construction is the right call: the alternative (silent fail-closed inside Lua) would manifest as 100 % 429s in production with no breadcrumb. `TestNewKeyedLimiterRejectsInvalidConfig` covers all four invalid permutations.
- `keyedLimiter.allow` correctly handles malformed Redis responses (length / type mismatch) by failing open with `degraded=true` and logging through `logDegrade`. `logDegrade` uses `atomic.Int64.CompareAndSwap` to throttle warnings to one per 30 s without a permanent `sync.Once` mute — appropriate for a long-running degradation.
- The three middleware variants (`RateLimitMiddleware`, `StrictIPRateLimitMiddleware`, `UIDRateLimitMiddleware`) consistently call `setRateLimitHeaders(...)` *before* `c.RenderError`, so the `X-RateLimit-*` and `Retry-After` headers ship even when a custom renderer is injected.
- Receiver migration to `(*WKHttp).RateLimitMiddleware` etc. is necessary so aborts can route through the instance-scoped renderer — correctly surfaced as a "Downstream impact" in the PR body. octo-server adopters should double-check the `r.Use(r.RateLimitMiddleware(...))` rename when they remove the local copies.

### `getClientIP` — trust assumption

- `ratelimit.go:196-210` documents the trust contract prominently: `X-Real-Ip` and `X-Forwarded-For` are believed verbatim, so any deployment that exposes the service without a header-stripping reverse proxy can have per-IP limits bypassed. This is unchanged behaviour from the octo-server original; the documentation upgrade is the right move for a shared library.
- Rightmost-XFF parsing (`parts[len(parts)-1]`) is the correct choice for the documented single-trusted-proxy CLB topology. Multi-hop deployments would need a count-from-right policy — out of scope.
- Fallback to `unknownIPKey` is technically a *shared-bucket* policy, not strictly "fail-closed" as the inline comment claims. The shared bucket still drains rapidly under load, so the practical effect ("everyone with no IP shares one quota") is the right defensive default. Not a behavioural bug — would suggest tightening the comment wording in a follow-up to avoid misleading future readers.

### `SecureCORSOverrideMiddleware`

- Wildcard matching in `IsOriginAllowed` is correctly anchored: stripping the leading `*` to a `.suffix` and requiring `len(origin) > len(suffix)` rejects bare-host shadowing (`a.com` vs `*.a.com`) and the empty `.a.com` edge case. Spot-checked candidate evasions:
  - `evil.com.a.com` — matches (legitimate sub-subdomain match).
  - `xa.com` — rejected (does not end in `.a.com`).
  - `b.a.com.evil.com` — rejected (ends in `.evil.com`, not `.a.com`).
- `addVaryOrigin` is idempotent and tolerates both single-header-multi-field and multi-header forms. Good.
- `ParseAllowedOrigins` explicitly drops literal `"*"` and warns — important, since `"*" + Allow-Credentials: true` is a spec violation that would otherwise propagate through this layer.
- The known preflight ordering limitation (CORSMiddleware aborts `OPTIONS` 204 before SecureCORSOverrideMiddleware can rewrite headers, so credentialed preflights still see `Allow-Origin: *`) is correctly inherited from the octo-server source and explicitly disclaimed in both the PR body and the godoc. The architectural fix (single whitelist-aware CORS middleware) is rightly deferred. In production this is mitigated by terminating CORS at nginx / CLB; flagging here so reviewers downstream don't mistake the limitation for a bug introduced by this PR.

### Concurrency — `atomic.Pointer[box]` over `atomic.Value`

- The `errorRendererBox` / `tokenParserBox` indirection is the textbook fix for `atomic.Value`'s type-invariance: storing two different concrete renderer types into a single `atomic.Value` panics. `TestSetErrorRendererSwitchConcreteTypes` is a real regression test for the failure mode, and `TestSetErrorRendererConcurrent` exercises mixed Set/Render under `-race`. Both pass cleanly here.
- `(*WKHttp).errorRenderer.Store(nil)` in `SetErrorRenderer(nil)` is safe — `(*Context).renderer()` checks `box == nil || box.r == nil` before dereferencing, falling back to `defaultErrorRenderer{}`. `TestSetErrorRendererNilResetsToFallback` directly covers reset + re-injection.

### Pool reset

- `(*Context).reset()` zeroes both `Context` and `wk`. `WKHttpHandler` sets `hc.wk = l` after `reset` and before `defer Put`, so a pooled `*Context` cannot leak a stale `*WKHttp` reference into the next request. `RenderError`'s nil-`wk` branch protects against the bare `&Context{Context: g}` test fixtures explicitly used in `error_renderer_test.go`. ✅

## 2. Security review (security_sensitive label)

| Surface | Verdict | Notes |
|---|---|---|
| Auth abort divergence (information leak) | ✅ | All three branches use static `DefaultMessage` strings; injected-parser unknown errors default to a generic `err.shared.auth.required` rather than echoing `err.Error()`. |
| Token parser sentinel API | ✅ | `ErrTokenMissing/NotFound/Invalid` are exported sentinels recognised via `errors.Is`. Custom parsers can opt into the same i18n keys without leaking message bodies. |
| Rate-limit fail mode | ✅ | Redis fail-open documented; `unknownIPKey` shared-bucket bounds abuse when the proxy contract is violated. |
| Per-IP trust contract | ⚠️ Documented | Service exposed without a header-stripping proxy can be bypassed via spoofed XFF. Already loud in godoc and PR body; deployment-time concern, not a code defect. |
| `SecureCORSOverrideMiddleware` preflight ordering | ⚠️ Known limitation | Inherited; not introduced. Production mitigation is nginx / CLB CORS termination. Tracking issue mentioned in the PR body should be filed before octo-server starts depending on this for browser security. |
| Wildcard CORS subdomain matching | ✅ | Anchored, scheme-aware, length-checked. |
| Atomic-pointer concurrency on injected renderers | ✅ | Boxed pointer + dedicated regression test. |

No injection, deserialisation, secret-handling, or path-traversal surfaces are introduced. The PR consistently routes user-controlled error messaging through the typed `ErrorSpec` channel and keeps response-rendering decisions centralised.

### Worth a human eyeball at deploy

- The `code-review` label's `security_sensitive` flag is well earned — confirm with the platform / SRE owners that the `getClientIP` trust assumption holds in every deployment topology that consumes the rebased `RateLimitMiddleware` from octo-lib (especially any new gateway / IPv6 path). If a topology change is in flight, the comment on `ratelimit.go:191-195` is the right place to land a follow-up note.
- Before octo-server's local `ratelimit.go` / `secure_cors.go` are deleted in the follow-up PR mentioned in `## What's in`, run `grep -R "wkhttp.RateLimitMiddleware\b" ` in octo-server to make sure every call site has been moved to `(*WKHttp).RateLimitMiddleware`. The package-level → method-level rename is the only mechanical risk left.

## 3. Test coverage observations

- Concurrency: `TestSetErrorRendererSwitchConcreteTypes` + `TestSetErrorRendererConcurrent` — exactly the right pair to lock in the boxed-pointer fix.
- Backward compatibility: `TestDefaultRendererPreservesLegacyEnvelope` and `TestAuthMiddlewareLegacyFallback` together pin the wire format down to the byte level for every default-renderer path.
- Custom-parser pathway: `TestAuthMiddlewareUsesInjectedTokenParser` + `…ErrorFallsThrough` cover both the happy path and the "unknown error" sentinel-fallback.
- Gated tests: `TestRateLimitMiddlewareDeniesViaRenderer` / `…DefaultFallback` correctly `t.Skipf` when testenv Redis is unreachable. The 53.0 % local coverage / 66.1 % testenv-coverage gap is fully accounted for by these.
- `TestSecureCORSOverrideMiddleware` — the test honestly admits it cannot assert the strict header outcome under the contrived "handler also writes ACAO" setup. That is the right call (asserting the false-positive would lock in an incorrect contract); the production ordering rule lives in the godoc.

## 4. Out-of-scope / follow-up suggestions

These are explicitly NOT blockers:

- `(*Context).CheckLoginRole` / `CheckLoginRoleIsSuperAdmin` still return hardcoded Chinese `errors.New(...)` strings (`http.go:179-201`). The PR body explains this is option 2 from issue #46 §4 — translation deferred to the calling service's i18n-aware `ResponseErrorL`. Reasonable scope cut for the 401 / 429 hot paths in this PR; downstream services adopting `wkhttp` need to remember not to forward `c.CheckLoginRole().Error()` directly to the client.
- `SharedUIDRateLimiter` deliberately staying in octo-server is the right call to avoid a `lib → server` reverse import. Worth tracking in case a second service ever wants to share the wrapper.
- `ratelimit.go:251` comment label "fail-closed" vs the actual shared-bucket implementation — minor wording polish for a future PR.
- Add a follow-up tracking issue for the `CORSMiddleware` + `SecureCORSOverrideMiddleware` merge so credentialed preflights can land at the application layer even when nginx / CLB are not in front.

## Verdict

No P0 (correctness / security / data-loss / build-break) and no P1 (functional bug / blocker) findings. All raised items are P2 / documentation / future-iteration in nature, and every non-trivial trade-off is already disclosed in the PR body or inline godoc. Tests, build, vet, and `-race` are clean. Consolidating these primitives into octo-lib pays off as soon as octo-server's local copies are removed in the follow-up.

Approving — the few suggestions above are optional polish, not gating changes.

PR #55, opened 2026-06-01
ci: Phase 1 test quality — gotestsum + diff coverage + PR summary

lml2468:
## Summary

Phase 1 test quality CI upgrade for octo-lib.

### Changes

- **gotestsum**: Replaces bare `go test` — produces JUnit XML for structured test results
- **`-shuffle=on`**: Added alongside existing `-race` flag
- **Coverage**: Generates `coverage.out` with atomic mode
- **Artifacts**: Uploads `junit.xml` + `coverage.out` (14-day retention)
- **PR Summary**: Writes test results to `$GITHUB_STEP_SUMMARY`
- **Diff Coverage**: New job using `octocov-action` for PR comments (60% threshold)
- **Permissions**: Added `pull-requests: write` for octocov

### What's NOT changed
- All existing jobs preserved (build, vet)
- No services needed (standalone tests)

### Checklist
- [x] gotestsum replaces bare go test
- [x] JUnit XML + coverage.out produced and uploaded
- [x] Diff coverage job with octocov PR comments
- [x] PR summary in $GITHUB_STEP_SUMMARY
- [x] YAML validated

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -177,6 +177,11 @@ jobs:
     permissions:
       contents: read
       pull-requests: write
+    # Match .github reusable-quality workflows R6 pattern: expose IS_FORK at
+    # job level so the octocov config can gate the PR-comment step without
+    # 403'ing on fork PRs (where GITHUB_TOKEN is read-only).
+    env:
+      IS_FORK: ${{ github.event.pull_request.head.repo.full_name != github.repository }}
     steps:
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
         with:
@@ -193,6 +198,10 @@ jobs:
       # Diff-only gate: report total coverage but only fail the job when the
       # *diff* coverage drops below 60%. coverage.acceptable is intentionally
       # omitted so this job does not block PRs based on the repo-wide baseline.
+      #
+      # comment.if gates the PR-comment step on non-fork PRs only. On fork PRs
+      # GITHUB_TOKEN is read-only, so octocov would 403 trying to comment and
+      # fail the job; the diff-coverage summary still appears in the job log.
       - name: Generate octocov config
         run: |
           cat > .octocov.yml <<'EOF'
@@ -203,10 +212,10 @@ jobs:
             coverage:
               acceptable: 60
           comment:
-            enable: true
+            if: env.IS_FORK != 'true'
             hideFooterLink: false
           summary:
-            enable: true
+            if: true
           EOF
 
       - name: Run octocov
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -95,8 +95,122 @@ jobs:
       - uses: actions/setup-go@40f1582b2485089dde7abd97c1529aa768e1baff # v5
         with:
           go-version: ${{ env.GO_VERSION }}
-      - name: Test
-        run: go test -race -shuffle=on -count=1 -timeout 5m ./...
+
+      # gotestsum is pinned to v1.12.0 because v1.12.3+ require Go >= 1.23
+      # and v1.13.0 declares 'go 1.24.0' (rejected by Go 1.20 module loader).
+      # When GO_VERSION here is bumped, this pin can move forward.
+      - name: Install gotestsum
+        run: go install gotest.tools/gotestsum@v1.12.0
+
+      - name: Run tests
+        id: run-tests
+        run: |
+          gotestsum \
+            --junitfile junit.xml \
+            --format short-verbose \
+            -- -race -shuffle=on -count=1 -timeout 5m \
+               -coverprofile=coverage.out \
+               -covermode=atomic ./...
+
+      - name: Upload test artifacts
+        if: always()
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4
+        with:
+          name: test-results
+          path: |
+            junit.xml
+            coverage.out
+          retention-days: 14
+
+      - name: Write PR summary
+        if: always()
+        run: |
+          outcome="${{ steps.run-tests.outcome }}"
+
+          case "$outcome" in
+            success) verdict="✅ **PASSED**" ;;
+            failure) verdict="❌ **FAILED**" ;;
+            *)       verdict="⏭️ **${outcome}**" ;;
+          esac
+
+          total=0; failed=0; errors=0; skipped=0; passed=0
+          if [ -f junit.xml ]; then
+            # Root <testsuites> aggregates tests/failures/errors; skipped is
+            # only on per-suite <testsuite> elements, so sum across all suites.
+            total=$(grep -o 'tests="[0-9]*"' junit.xml | head -1 | grep -o '[0-9]*' || echo 0)
+            failed=$(grep -o 'failures="[0-9]*"' junit.xml | head -1 | grep -o '[0-9]*' || echo 0)
+            errors=$(grep -o 'errors="[0-9]*"' junit.xml | head -1 | grep -o '[0-9]*' || echo 0)
+            skipped=$(grep -o 'skipped="[0-9]*"' junit.xml | grep -o '[0-9]*' | paste -sd+ - | bc 2>/dev/null || echo 0)
+            passed=$(( total - failed - errors - skipped ))
+            [ "$passed" -lt 0 ] && passed=0
+          fi
+
+          {
+            echo "## 🧪 Test Results"
+            echo
+            echo "${verdict}"
+            echo
+            echo "| Metric | Count |"
+            echo "|--------|-------|"
+            echo "| 🧮 Total | ${total} |"
+            echo "| ✅ Passed | ${passed} |"
+            echo "| ❌ Failed | ${failed} |"
+            echo "| 💥 Errors | ${errors} |"
+            echo "| ⏭️ Skipped | ${skipped} |"
+            echo
+            echo "**Flags:** \`-race -shuffle=on\`"
+            echo
+            echo "📦 [Download artifacts](../actions/runs/${{ github.run_id }}#artifacts)"
+          } >> "$GITHUB_STEP_SUMMARY"
+
+  diff-coverage:
+    needs: [changes, test]
+    if: |
+      always() &&
+      github.event_name == 'pull_request' &&
+      needs.test.result == 'success' &&
+      needs.changes.outputs.code == 'true'
+    name: Diff Coverage
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    # Least-privilege: only this job needs to comment on PRs.
+    permissions:
+      contents: read
+      pull-requests: write
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+        with:
+          fetch-depth: 0
+      - uses: actions/setup-go@40f1582b2485089dde7abd97c1529aa768e1baff # v5
+        with:
+          go-version: ${{ env.GO_VERSION }}
+
+      - name: Download test results
+        uses: actions/download-artifact@d3f86a106a0bac45b974a628896c90dbdf5c8093 # v4
+        with:
+          name: test-results
+
+      # Diff-only gate: report total coverage but only fail the job when the
+      # *diff* coverage drops below 60%. coverage.acceptable is intentionally
+      # omitted so this job does not block PRs based on the repo-wide baseline.
+      - name: Generate octocov config
+        run: |
+          cat > .octocov.yml <<'EOF'
+          coverage:
+            paths:
+              - coverage.out
+          diff:
+            coverage:
+              acceptable: 60
+          comment:
+            enable: true
+            hideFooterLink: false
+          summary:
+            enable: true
+          EOF
+
+      - name: Run octocov
+        uses: k1LoW/octocov-action@b3b6ee60482a667950f87553abf1df63217235d9 # v1.5.1
 
   vet:
     needs: [changes]


Jerry-Xin (changes requested):
🔴 **CI Red — Test job fails, must fix before merge.**

**Root cause:** `gotestsum@latest` resolves to v1.13.0 which requires Go 1.24+, but this repo pins `GO_VERSION: '1.20'`. The error:

```
go: gotest.tools/gotestsum@latest (in gotest.tools/gotestsum@v1.13.0):
    go.mod:3: invalid go version '1.24.0': must match format 1.23
```

**Fix:** Pin gotestsum to a version compatible with Go 1.20. `gotestsum@v1.12.1` is the last release supporting Go 1.21 toolchains — verify it builds under 1.20 or check release notes. Alternatively, if Go 1.20 is about to be bumped repo-wide, coordinate that first.

```diff
- run: go install gotest.tools/gotestsum@latest
+ run: go install gotest.tools/gotestsum@v1.12.1
```

---

Beyond the CI failure, the workflow changes themselves are well-structured:

🔴 **Blocking:**

1. **Supply chain: `k1LoW/octocov-action@v1` uses mutable tag.** All other third-party actions in the `changes` job already use SHA pins (checkout `@34e111...`, dorny `@6852f9...`). Pin octocov to a SHA for consistency — `v1` can be silently replaced by a compromised tag.

2. **`pull-requests: write` is workflow-wide but only needed by `diff-coverage`.** The `changes` job already has a scoped `permissions` block. Move `pull-requests: write` into `diff-coverage` as a job-level permission to follow least-privilege (build/test/vet don't need it).

3. **`Write PR summary` step uses `${{ steps.run-tests.outcome }}` in a shell variable without quotes around the expression substitution.** While Actions context injection is generally safe here (outcome is a fixed enum), the pattern `outcome="${{ steps.run-tests.outcome }}"` is an expression injection vector if the source were attacker-controlled. More importantly, when `run-tests` is skipped (gotestsum install fails, step doesn't execute), `outcome` becomes `"skipped"` not `"failure"` — the summary says ❌ FAILED but semantically the tests never ran. Consider checking `steps.run-tests.conclusion` or adding an explicit skipped state.

💬 **Non-blocking:**

- `grep -o 'tests="[0-9]*"' junit.xml` parses JUnit XML with regex. This is fragile — a multi-suite report will have multiple `tests=` attributes and `head -1` grabs the first testsuite, not the root `<testsuites>` element. Works fine for a single-suite Go project today, but will silently undercount if the project grows. Consider `gotestsum --junitfile-testsuite-name=full` + a proper XML parser or just `go tool test2json` counts.
- The heredoc in `Write PR summary` has leading whitespace (the `cat >> "$GITHUB_STEP_SUMMARY" <<EOF` block is indented inside the `run:` block). GitHub renders this as a code block in the step summary because every line has leading spaces. Use `<<-EOF` with tabs, or dedent the heredoc body.
- `actions/upload-artifact@v4` + Node.js 20 deprecation warning already showing in logs — worth tracking for a follow-up bump.

✅ **Highlights:**
- gotestsum + JUnit XML + coverage.out is a solid testing foundation
- `diff-coverage` gated on `needs.test.result == 'success'` — correct dependency chain
- `-shuffle=on` alongside existing `-race` is a good addition for flake detection
- octocov config with 60% diff coverage threshold is reasonable for Phase 1

lml2468:
Gate check failed: CI is red, so I did not proceed into the diff review.

Blocking items:
1. CI gate: `Test`, `check-sprint / check-sprint`, and `welcome / welcome` failed on PR #55; `Vet` and CodeQL were still pending at gate-check time. Please fix the failing checks and rerun CI before code review continues.

Non-blocking notes:
- None.

Highlights:
- No code-level review performed because the Phase 1 gate failed.

yujiawei (changes requested):
# Code Review — PR #55 (octo-lib)

## Verdict: CHANGES_REQUESTED

This PR is a solid step toward better test-result visibility (gotestsum + JUnit + diff coverage + PR summary). However, there is a **blocking build break**: the test job will fail to start on every run because `go install ...@latest` now pulls a gotestsum version that cannot be built with the pinned Go toolchain.

---

## P0 — Blocking

### 1. `gotestsum@latest` is incompatible with the pinned `GO_VERSION: '1.20'` → test job fails on every run

`.github/workflows/ci.yml`:
```yaml
env:
  GO_VERSION: '1.20'
...
      - name: Install gotestsum
        run: go install gotest.tools/gotestsum@latest
```

`go install ...@latest` currently resolves to **gotestsum v1.13.0**, whose `go.mod` declares `go 1.24.0`. Go 1.20's module loader cannot parse a three-part `go 1.24.0` version line and aborts before installing.

Reproduced locally with the toolchain pinned to 1.20:
```
$ GOTOOLCHAIN=go1.20.14 go install gotest.tools/gotestsum@latest
go: gotest.tools/gotestsum@latest (in gotest.tools/gotestsum@v1.13.0):
    go.mod:3: invalid go version '1.24.0': must match format 1.23
exit status 1
```

Impact:
- The **Install gotestsum** step fails → the **Run tests** step never executes → the whole `test` job is red.
- Because `diff-coverage` gates on `needs.test.result == 'success'`, it is skipped too.
- Net effect: the PR's entire stated goal (structured results + diff coverage) does not run, and CI is broken for the repo on merge.

This is non-deterministic over time as well: `@latest` floats, so even a green run today can break tomorrow when a new gotestsum requires a newer Go than the pinned version. Pinning is both the fix and the prevention.

**Fix — pin to a version compatible with Go 1.20.** I verified the boundary empirically (all under `GOTOOLCHAIN=go1.20.14`):

| gotestsum | go.mod `go` directive | installs on Go 1.20? |
|---|---|---|
| v1.13.0 (`@latest`) | 1.24.0 | ❌ |
| v1.12.3 | 1.23.0 | ❌ |
| **v1.12.0** | 1.20 | ✅ |
| v1.11.0 | ✅ | ✅ |

```yaml
      - name: Install gotestsum
        run: go install gotest.tools/gotestsum@v1.12.0
```

Alternatively, if you prefer to stay on the latest gotestsum, bump `GO_VERSION` to `'1.24'` (or whatever the project supports) — but that is a larger decision affecting build/vet too, so pinning gotestsum is the lower-risk change for this PR.

---

## P2 — Non-blocking (recommend fixing before/with merge)

### 2. PR summary "Skipped" count is always 0

```yaml
skipped=$(grep -o 'skipped="[0-9]*"' junit.xml | head -1 | grep -o '[0-9]*' || echo 0)
```

gotestsum's JUnit output does **not** emit a `skipped="N"` attribute on `<testsuites>`/`<testsuite>`. Skipped tests are represented as `<skipped>` child elements inside each `<testcase>`. I confirmed this with a run containing a `t.Skip()` test: the rolled-up `skipped="..."` attribute does not exist, so the summary will report **Skipped: 0** even when tests are skipped.

Suggested fix — count the skipped elements instead:
```bash
skipped=$(grep -c '<skipped' junit.xml || echo 0)
```

### 3. PR summary labels total test count as "Passed"

```yaml
passed=$(grep -o 'tests="[0-9]*"' junit.xml | head -1 | grep -o '[0-9]*' || echo 0)
...
| ✅ Passed | ${passed} |
```

`tests="N"` on the root `<testsuites>` is the **total** number of tests, not the number that passed. On a green run total == passed, but on a run with failures/skips the "Passed" row will overcount. Either relabel the row to **Total**, or compute `passed = tests - failed - skipped`.

(Note: the `tests`/`failures` parsing via `head -1` is correct — it picks the aggregate root element, verified against real gotestsum output.)

### 4. `pull-requests: write` granted workflow-wide

```yaml
permissions:
  contents: read
  pull-requests: write
```

Only the `diff-coverage` job (octocov) needs PR write. `build`, `test`, and `vet` now inherit it unnecessarily. Least-privilege suggestion: keep the top-level default at `pull-requests: read` (or `contents: read` only) and grant `pull-requests: write` in a per-job `permissions:` block on `diff-coverage`. The `changes` job already demonstrates this per-job pattern.

### 5. Floating action/tool versions (supply-chain hygiene)

New steps use floating tags: `actions/upload-artifact@v4`, `actions/download-artifact@v4`, `k1LoW/octocov-action@v1`, and `go install ...@latest`. The existing `changes` job pins by commit SHA (`actions/checkout@34e1148... # v4`). This is consistent with the existing `build`/`vet` jobs (which also use `@v4`/`@v5`), so it's not a regression — but for a CI-hardening PR it would be worth SHA-pinning the new third-party actions (especially `k1LoW/octocov-action`) to match the `changes` job's standard.

---

## Additional observations (FYI, no action required)

- **octocov diff coverage** computes PR-line coverage from `coverage.out` + the diff, so it works without a configured `datastores:` baseline. The trend/total-coverage-vs-base comparison would need a datastore, but the 60% diff-coverage gate this PR targets does not — so the config is fine as-is.
- The `diff-coverage` job correctly downloads `coverage.out` from the `test` job's artifact and gates on `needs.test.result == 'success'`, so the artifact is guaranteed present when it runs. Good.
- `-shuffle=on` and `-covermode=atomic` (paired with `-race`) are correct choices. 

---

## Summary

| # | Severity | Issue |
|---|---|---|
| 1 | **P0** | `gotestsum@latest` (v1.13.0, needs Go 1.24) breaks under pinned Go 1.20 → test job fails. Pin to `@v1.12.0`. |
| 2 | P2 | "Skipped" count always 0 (no `skipped=` attr in gotestsum JUnit). |
| 3 | P2 | "Passed" row actually shows total test count. |
| 4 | P2 | `pull-requests: write` should be scoped to the octocov job. |
| 5 | P2 | SHA-pin new third-party actions to match the `changes` job. |

Requesting changes for **#1** (build break). The P2 items are recommended but not merge-blocking on their own.

lml2468:
Closing as deferred / obsolete.

The `feat/test-quality-phase1` series was intended to introduce gotestsum + diff coverage + PR summary. It has been deferred indefinitely as part of Mininglamp-OSS workflow cleanup; if/when test-quality phase 1 is restarted it will be re-scoped and re-opened on a fresh branch.

Branch has not been updated since 2026-05-30/31 and is divergent from current `main`.

lml2468:
Reopening — closed in error during workflow cleanup sweep.

`feat/test-quality-phase1` series is still in-scope; the test-quality phase 1 effort (gotestsum + diff coverage + PR summary) is being retained, not deferred. Re-opening to keep the PR queue intact.

Jerry-Xin (changes requested):
The PR is project-relevant, but the new test job currently breaks CI because the installed gotestsum version is incompatible with the workflow’s pinned Go version.

**🔴 Blocking**
- 🔴 Critical: [`.github/workflows/ci.yml:98`](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-486yuyrl/repo/.github/workflows/ci.yml:98) installs `gotest.tools/gotestsum@latest`, while [`.github/workflows/ci.yml:31`](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-486yuyrl/repo/.github/workflows/ci.yml:31) pins CI to Go `1.20`. The current `gotestsum@latest` resolves to `v1.13.0`, whose `go.mod` declares `go 1.24.0`, so `go install gotest.tools/gotestsum@latest` will fail on the configured runner before tests run. Pin gotestsum to a verified Go 1.20-compatible version, or intentionally raise the repo/workflow Go version.

**💬 Non-blocking**
- 🟡 Warning: [`.github/workflows/ci.yml:22`](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-486yuyrl/repo/.github/workflows/ci.yml:22) grants `pull-requests: write` at workflow scope, so jobs that execute PR code inherit broader permissions than needed. Prefer moving `pull-requests: write` to the `diff-coverage` job only, and keep build/test/vet at read-only permissions.
- 🟡 Warning: [`.github/workflows/ci.yml:133`](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-486yuyrl/repo/.github/workflows/ci.yml:133) labels the JUnit `tests` attribute as “Passed”, but that value is total tests, not passed tests. Compute `passed = tests - failures - errors - skipped` if the summary is meant to report passed tests accurately.
- 🔵 Suggestion: [`.github/workflows/ci.yml:195`](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-486yuyrl/repo/.github/workflows/ci.yml:195) uses `k1LoW/octocov-action@v1`. For CI supply-chain stability, consider pinning third-party actions to immutable SHAs, matching the existing pinned `dorny/paths-filter` style.

**✅ Highlights**
- The PR preserves the existing build and vet jobs.
- Uploading `junit.xml` and `coverage.out` with `if: always()` is the right shape for post-failure diagnostics.
- The diff coverage job is correctly gated behind successful tests and code-changing PRs.

yujiawei (changes requested):
# Code Review — PR #55 (octo-lib)

## Summary

This PR upgrades the `test` CI job (gotestsum + JUnit XML + shuffle + coverage artifacts + a `$GITHUB_STEP_SUMMARY` table) and adds a new `diff-coverage` job built on `k1LoW/octocov-action`. The mechanical parts (gotestsum invocation, artifact upload/download wiring, job gating) are sound. However, the centerpiece of the PR — "diff coverage with a 60% threshold" — does not do what the title/description claim, and the way the coverage gate is configured can block unrelated PRs. Requesting changes on that one functional issue; the rest are non-blocking nits.

## Findings

### P1 — `diff-coverage` does not measure diff/patch coverage, and the gate is on *total* coverage

`.octocov.yml` generated by the job:

```yaml
coverage:
  paths:
    - coverage.out
  acceptable: 60
diff:
  coverage:
    acceptable: 60
```

Two problems:

1. **`diff.coverage.acceptable` is not part of octocov's schema.** In octocov, the `diff:` block configures *report-vs-report comparison* and only accepts `diff.path:` (a baseline report file) or `diff.datastores:` (a stored baseline, e.g. from the default branch). There is no `diff.coverage.acceptable` key, and octocov does not compute per-changed-line ("patch") coverage at all — it compares whole-report metrics. So the `diff:` block here has no baseline to compare against and the nested `acceptable: 60` is silently ignored. The PR title and description ("Diff Coverage … 60% threshold") promise something octocov isn't being asked (or able) to do.

2. **The only gate that actually fires is `coverage.acceptable: 60` on *total* project coverage.** octocov exits non-zero when total coverage is below `acceptable`. For a library whose overall coverage is under 60%, this means the `diff-coverage` job will fail on **every** PR — including PRs that touch zero Go code paths — not just PRs that lower coverage. That's a footgun: a perpetually-red (or merge-blocking, if this check is/ becomes required) job decoupled from what the author changed.

**Recommendation (pick one):**
- If the goal is true delta/diff tracking vs `main`: configure `diff.datastores:` (e.g. an artifact or `gh` datastore) so octocov has a baseline, and keep `coverage.acceptable` only if the repo is already ≥60%.
- If the goal is just "comment coverage on PRs without blocking yet": drop `coverage.acceptable` (or set it to a value the repo currently clears) so the job reports without gating, and rename the job/PR away from "diff coverage" to avoid implying per-line patch gating that octocov doesn't provide.
- Also add `actions: read` to the job's permissions if you want octocov's test-execution-time metric (optional; without it that one metric is just omitted).

### P2 — `$GITHUB_STEP_SUMMARY` counts are inaccurate

In the "Write PR summary" step:

```bash
passed=$(grep -o 'tests="[0-9]*"' junit.xml | head -1 | grep -o '[0-9]*' || echo 0)
```

- `tests="N"` is the **total** test count (passed + failed + skipped), so labeling it "✅ Passed" overstates passes whenever anything fails or is skipped.
- `head -1` takes the first matching element. gotestsum emits one `<testsuite>` per package under a `<testsuites>` root; if the first match is a single package rather than the aggregate root, the numbers reflect only that package. Parsing JUnit XML with `grep` is fragile here.

This is summary-only cosmetics (no effect on pass/fail), but the displayed numbers will frequently be wrong. Consider deriving counts from gotestsum directly (e.g. `gotestsum --jsonfile` + a small parse) or computing `passed = tests - failures - skipped`.

### P2 — Inconsistent action pinning (supply-chain)

The pre-existing `changes` job pins third-party actions by full commit SHA (`dorny/paths-filter@6852f92…`, `actions/checkout@34e11487…`). The new/edited jobs use floating tags: `actions/checkout@v4`, `actions/setup-go@v5`, `k1LoW/octocov-action@v1`, and `go install gotest.tools/gotestsum@latest`. `@latest` in particular makes CI non-reproducible and pulls an unpinned binary on every run. For consistency with the repo's existing posture, pin these to SHAs (and gotestsum to a tagged version).

## Non-blocking observations

- Artifact wiring (`upload-artifact` in `test` → `download-artifact` in `diff-coverage`, same run) is correct; `coverage.out` will be present for octocov.
- Job gating (`always() && needs.test.result == 'success' && needs.changes.outputs.code == 'true'`, `pull_request` only) is reasonable. Note octocov can only comment on same-repo PRs — fork PRs will fall back to job-summary output, which is expected.
- `-shuffle=on` and JUnit artifacts are good additions and align with the stated Phase 1 goal.

## Verdict

The infra plumbing is solid, but the "diff coverage / 60% threshold" feature as configured neither measures the diff nor gates on it — it gates on total coverage and will misfire on PRs that don't touch coverage. That contradicts the PR's stated purpose, so I'm requesting changes to fix the octocov configuration (and the summary counts / pinning as smaller follow-ups).

Jerry-Xin (changes requested):
Project relevance gate passed: this PR updates octo-lib’s CI workflow, which is in scope.

🔴 Blocking

🔴 Critical — The diff coverage gate is not actually enforced. In [.github/workflows/ci.yml](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-uq27_3t0/repo/.github/workflows/ci.yml:202), the generated config uses:

```yaml
diff:
  coverage:
    acceptable: 60
```

That is not a valid octocov threshold shape for enforcing coverage. With this config, octocov ignores the nested `diff.coverage.acceptable` setting and exits successfully even when total coverage is below 60%; there is also no `diff.path` or `diff.datastores` configured for octocov’s report comparison mode. This means the PR advertises a 60% diff coverage gate, but the new `diff-coverage` job can pass without enforcing that gate. The CI behavior does not match the PR’s stated purpose.

💬 Non-blocking

🟡 Warning — The test summary parser in [.github/workflows/ci.yml](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-uq27_3t0/repo/.github/workflows/ci.yml:140) relies on `grep` over XML attributes. It is probably adequate for gotestsum’s current JUnit shape, but it is brittle compared with parsing the XML or using a structured gotestsum output mode.

✅ Highlights

The PR keeps the existing build and vet jobs intact, pins actions by SHA, scopes `pull-requests: write` to the octocov job, and the raw Go test command with `-race`, `-shuffle=on`, `-coverprofile`, and `-covermode=atomic` produced a valid coverage profile in local verification.

yujiawei (changes requested):
# Code Review — PR #55 (octo-lib)

## Summary

This PR upgrades the `test` job (gotestsum + JUnit XML + coverage artifacts + `$GITHUB_STEP_SUMMARY`) and adds a new `diff-coverage` job built on `k1LoW/octocov-action`. The test-quality plumbing is solid and the action pins / least-privilege `permissions` are good practice. However, the **headline feature — "diff coverage with a 60% threshold gate" — does not actually work**, and the PR summary's skipped/passed counts are computed from XML attributes that gotestsum does not emit. Requesting changes for the coverage gate; the rest are P2/nits.

Verified against: octocov config schema (struct `Diff{Path, Datastores, If}` at v0.59.0 **and** v0.75.8 — the `latest` the action installs), `goccy/go-yaml` unknown-key behavior, and gotestsum v1.12.0 JUnit XML structs.

---

## P1 — The advertised "diff coverage 60% threshold" is silently a no-op

The generated `.octocov.yml` is:

```yaml
diff:
  coverage:
    acceptable: 60
```

octocov does **not** have a `diff.coverage.acceptable` key. Its `Diff` config supports only `path`, `datastores`, and `if` (`config/config.go`, identical in v0.59.0 and the `latest`/v0.75.8 the action pulls via `version: default`). Two consequences, both verified:

1. **The threshold is silently ignored.** octocov uses `goccy/go-yaml`, which does **not** error on unknown keys (reproduced locally: unmarshalling `coverage:\n  acceptable: 60` into the real `Diff` struct returns `err=<nil>` and an empty struct). So `acceptable: 60` is dropped on the floor — the job will **never fail** on low coverage. The inline comment ("only fail the job when the *diff* coverage drops below 60%") describes behavior that cannot happen.
2. **octocov does not compute patch/diff coverage at all.** octocov's `diff` feature compares the *whole-repo* coverage report against a **previously stored baseline** (`diff.path` or `diff.datastores`). Neither is configured here, so there is no baseline to diff against — the job just reports current total coverage in a PR comment. It is not measuring "coverage of the changed lines," which is what "diff coverage" implies and what the PR description and checklist claim.

Net: the job posts a coverage comment, but the "60% diff-coverage gate" advertised in the title, description, and checklist is inert. Please either:
- wire up a real changed-lines gate (a dedicated patch-coverage tool, or octocov with a proper datastore baseline + a valid `coverage.acceptable` expression such as `coverage.acceptable: "diff >= 0"`), **or**
- drop the "60% threshold" framing and relabel this as a report-only coverage comment, removing the misleading config + comment.

---

## P2 — PR summary: `Skipped` is always 0 and inflates `Passed`

In the "Write PR summary" step:

```bash
skipped=$(grep -o 'skipped="[0-9]*"' junit.xml | grep -o '[0-9]*' | paste -sd+ - | bc ...)
passed=$(( total - failed - errors - skipped ))
```

gotestsum v1.12.0 does not emit a `skipped="N"` attribute on either `<testsuites>` or `<testsuite>`. Skipped tests are emitted as child `<skipped>` elements inside `<testcase>` (per `internal/junitxml/report.go`: the `JUnitTestSuite` struct has only `tests/failures/time/name/timestamp` attrs — no `skipped`, no `errors`). So:
- `Skipped` will **always render 0**, and
- skipped tests fall into `passed = total - failed - errors - 0`, **over-counting Passed** by the number of skips.

The inline comment ("skipped is only on per-suite `<testsuite>` elements, so sum across all suites") is factually incorrect for this XML format. To count skips, count `<skipped` child elements instead (e.g. `grep -c '<skipped'`). `total/failed/errors` read from the root `<testsuites>` are correct.

---

## P2 — octocov comment will have nothing to diff against

Because no `report.datastores` / `diff.datastores` baseline is configured, the very first and every subsequent run has no prior report to compare to. The comment degrades to "current total coverage" with no trend. If the intent is a trend/gate, a datastore (e.g. artifact, GitHub, or `report.path`) is required. Tied to the P1 fix.

---

## Nits

- `diff-coverage` `if:` combines `always() && ... && needs.test.result == 'success'`. The `always()` is redundant given the explicit `needs.test.result == 'success'` guard — it doesn't break anything but is contradictory to read. Consider dropping `always()`.
- The gotestsum `@v1.12.0` pin rationale is **correct** and well-documented — v1.12.0's `go.mod` declares `go 1.13`, compatible with the repo's Go 1.20; newer gotestsum requires Go ≥1.23. Nice.
- `-covermode=atomic` alongside `-race` is the right combination. 👍
- Scoping `pull-requests: write` to only the `diff-coverage` job while the top-level stays `contents: read` is good least-privilege. 👍

---

## Verdict

The test-quality scaffolding (gotestsum, artifacts, summary) is a genuine improvement, but the central advertised feature — the 60% diff-coverage gate — is non-functional as written, and the summary's skipped/passed math is wrong. Requesting changes to fix or correctly re-scope the coverage gate (P1) and the skipped-count parsing (P2).

Jerry-Xin (changes requested):
The CI additions are project-relevant, but the PR summary currently reports skipped tests as passed.

**🔴 Blocking**
- 🔴 Critical: `.github/workflows/ci.yml:138-145` counts skipped tests from `skipped="N"` attributes, but `gotestsum --junitfile` emits skipped tests as `<skipped ...>` elements in this repo’s JUnit output. As a result, skipped tests are reported as `0`, and `passed` is inflated. A local run produced 244 tests with 3 skipped, but this script would summarize them as 244 passed / 0 skipped. Count skipped elements instead, for example with `grep -c '<skipped' junit.xml || true`, before computing passed.

**✅ Highlights**
- The test command now produces both JUnit and coverage artifacts while preserving `-race`, `-shuffle=on`, and `-count=1`.
- The diff coverage job is scoped to successful PR test runs and uses a narrower `pull-requests: write` permission only where commenting is needed.
- The `gotestsum` version pin is appropriate for the repository’s current `go 1.20` baseline.

yujiawei (changes requested):
# Code Review — PR #55 (octo-lib)

## Summary

This revision fixes the earlier build break (gotestsum is now pinned to `v1.12.0`, installable under the repo's `GO_VERSION: '1.20'`), scopes `pull-requests: write` to the `diff-coverage` job, and SHA-pins the new third-party actions. Those are real improvements.

However, the **headline feature — "diff coverage with a 60% threshold gate" — is still a no-op as configured**, and the PR summary's `Skipped`/`Passed` math is still wrong. Requesting changes on the coverage gate (functional defect in the centerpiece of the PR); the summary parsing is a non-blocking follow-up.

Findings verified against octocov's config source (`config/config.go` `Diff` struct), its non-strict YAML loader, and gotestsum `v1.12.0`'s JUnit XML structs.

---

## P1 — The advertised "diff coverage 60% threshold" gate does nothing

The generated `.octocov.yml` (`.github/workflows/ci.yml:207-219`):

```yaml
coverage:
  paths:
    - coverage.out
diff:
  coverage:
    acceptable: 60
```

octocov's `diff:` block has **no `coverage` subkey**. The `Diff` config struct supports only `path`, `datastores`, and `if`:

```go
type Diff struct {
    Path       string   `yaml:"path,omitempty"`
    Datastores []string `yaml:"datastores,omitempty"`
    If         string   `yaml:"if,omitempty"`
}
```

octocov loads config with `goccy/go-yaml`'s non-strict `yaml.Unmarshal` (no `Strict()` option), so the unknown `diff.coverage` mapping is **silently discarded** — it does not error, and it does not configure anything. The only coverage gate octocov enforces is the **top-level** `coverage.acceptable`, which this config does **not** set; with an empty condition, octocov's `coverageAcceptable()` returns early and the job passes unconditionally.

Two consequences:

1. **The 60% threshold never fires.** The job will pass regardless of coverage. The inline comment at `ci.yml:198-200` ("only fail the job when the *diff* coverage drops below 60%") describes behavior that cannot occur with this config.
2. **octocov does not compute patch/diff-line coverage at all.** Its `diff` feature compares the *whole-report* coverage against a stored baseline (`diff.path` / `diff.datastores`), neither of which is configured here. So even conceptually, "coverage of the changed lines" is not what this job measures — it just posts current total coverage as a PR comment.

Net: the PR title, description, and checklist advertise a "60% diff coverage gate," but the job neither measures diff coverage nor gates on it. Please pick one:

- **If a real gate is intended:** drop the invalid `diff.coverage` block and set a top-level `coverage.acceptable: "60%"` (gates on total coverage — only safe if the repo is already ≥60%), and/or wire up `diff.datastores` with a `main` baseline for trend comparison; or adopt a dedicated changed-lines coverage tool if true patch-coverage gating is the goal.
- **If report-only is intended for Phase 1:** remove the misleading `diff.coverage.acceptable` block and the inline comment, and relabel the job / PR away from "diff coverage … 60% threshold" so it doesn't imply a gate that isn't there.

---

## P2 — PR summary: `Skipped` is always 0 and `Passed` is over-counted

In the "Write PR summary" step (`ci.yml:143-144`):

```bash
skipped=$(grep -o 'skipped="[0-9]*"' junit.xml | grep -o '[0-9]*' | paste -sd+ - | bc 2>/dev/null || echo 0)
passed=$(( total - failed - errors - skipped ))
```

gotestsum `v1.12.0` does **not** emit a `skipped="N"` attribute on `<testsuites>` or `<testsuite>`. Its `JUnitTestSuite`/`JUnitTestSuites` structs carry only `tests`/`failures`/`errors`/`time`/`name`; skipped tests are represented as `<skipped>` **child elements** inside each `<testcase>`. Therefore:

- `Skipped` will always render **0**, and
- skipped tests are folded into `passed = total - failed - errors - 0`, so **`Passed` is over-counted** by the number of skips.

The inline comment at `ci.yml:138-139` ("skipped is only on per-suite `<testsuite>` elements, so sum across all suites") is incorrect for this XML shape. To count skips, count the child elements instead, e.g. `skipped=$(grep -c '<skipped' junit.xml || echo 0)`. The `total`/`failed`/`errors` reads from the root `<testsuites>` are correct.

This is summary-only cosmetics (no effect on the actual pass/fail of the job), hence P2 — but the displayed numbers will be wrong whenever any test is skipped.

---

## Confirmed-good (previously raised, now fixed)

- **gotestsum pin** (`ci.yml:99-103`): `v1.12.0` declares `go 1.13` in its `go.mod` and installs cleanly under the pinned Go 1.20 toolchain. The earlier `@latest` → v1.13.0 (`go 1.24.0`) build break is resolved, and the pin rationale comment is accurate.
- **Least-privilege**: top-level stays `contents: read`; `pull-requests: write` is now scoped to the `diff-coverage` job only (`ci.yml:177-179`).
- **Supply chain**: `octocov-action`, `upload-artifact`, `download-artifact`, `checkout`, `setup-go` are SHA-pinned, matching the `changes` job's posture.
- **Fork gating**: `comment.if: env.IS_FORK != 'true'` + job-level `IS_FORK` correctly avoids the 403 octocov would hit commenting on fork PRs with a read-only token.
- Artifact wiring (`test` upload → `diff-coverage` download in the same run) and job gating (`needs.test.result == 'success'`) are correct. `-shuffle=on` + `-covermode=atomic` alongside `-race` are good choices.

---

## Nit

- `diff-coverage`'s `if:` (`ci.yml:168-172`) combines `always() && … && needs.test.result == 'success'`. The `always()` is redundant given the explicit `needs.test.result == 'success'` guard — harmless, but contradictory to read.

---

## Verdict

The test-quality scaffolding (gotestsum, JUnit, artifacts, summary, fork-safe octocov comment) is a solid Phase 1 improvement and the prior build break is fixed. But the central advertised feature — the 60% diff-coverage gate — is non-functional as configured (an unknown octocov key that is silently ignored), and the summary's skipped/passed counts are wrong. Requesting changes to fix or honestly re-scope the coverage gate (P1); the summary parsing (P2) should be fixed alongside.

lml2468:
Closing per project decision (2026-06-06).

This PR is part of the test-quality / T15 workstream. The owner has decided to stop and re-scope the test-quality initiative; this PR is being closed as part of that wind-down. The work is preserved in branch history and the .github#55 reusable workflows already merged to main remain in place — only the consumer-side opt-in PRs and the T15 Phase 2 / R7 follow-ups are being closed here.

If this PR's content is later wanted, re-open against latest main with a fresh rebase.

PR #67, opened 2026-06-08
ci: pass TRIAGE_WEBHOOK_URL to octo-issue-notify

lml2468:
The issue notification caller only forwarded OCTO_BOT_TOKEN, so the triage webhook mode was silently skipped and the triage agent was never invoked. Forward the org-level TRIAGE_WEBHOOK_URL secret so triage runs on new issues.

### DIFF
--- a/.github/workflows/octo-issue-feed.yml
+++ b/.github/workflows/octo-issue-feed.yml
@@ -18,3 +18,4 @@ jobs:
       event_action: ${{ github.event.action }}
     secrets:
       OCTO_BOT_TOKEN: ${{ secrets.OCTO_BOT_TOKEN }}
+      TRIAGE_WEBHOOK_URL: ${{ secrets.TRIAGE_WEBHOOK_URL }}


Jerry-Xin (approved):
This PR is in scope for `Mininglamp-OSS/octo-lib`: it updates repository CI/automation for Octo issue notification, which is part of this project’s GitHub workflow surface.

🔴 Blocking

None.

💬 Non-blocking

None.

✅ Highlights

🔵 Suggestion / Positive: `.github/workflows/octo-issue-feed.yml:19-21` now forwards `TRIAGE_WEBHOOK_URL` alongside `OCTO_BOT_TOKEN`, matching the PR’s stated intent to enable the reusable issue notification workflow’s optional triage webhook path.

🔵 Suggestion / Positive: The change is narrowly scoped, does not alter workflow permissions, and does not introduce code execution from issue content. No additional tests are needed beyond existing workflow sanity validation for this one-line reusable-workflow secret mapping.

yujiawei (approved):
# Code Review — PR #67 (octo-lib)

## Summary

Single-line change: forwards the org-level `TRIAGE_WEBHOOK_URL` secret from the `octo-issue-feed.yml` caller to the reusable `octo-issue-notify.yml@v1` workflow, so the optional triage-webhook path is actually wired up. The stated bug — only `OCTO_BOT_TOKEN` was forwarded, so triage was silently skipped — is real and this is the correct, minimal fix.

```yaml
     secrets:
       OCTO_BOT_TOKEN: ${{ secrets.OCTO_BOT_TOKEN }}
+      TRIAGE_WEBHOOK_URL: ${{ secrets.TRIAGE_WEBHOOK_URL }}
```

## Verification (cross-file contract)

I verified the change against the called reusable workflow at the pinned tag, since a secret forward is only correct if the callee declares and consumes it.

- ✅ **Declared by callee.** `Mininglamp-OSS/.github/.github/workflows/octo-issue-notify.yml@v1` declares `TRIAGE_WEBHOOK_URL` under `on.workflow_call.secrets` with `required: false`. The secret name matches exactly.
- ✅ **Consumed by callee.** It is passed to `webhook-url:` of the `Mininglamp-OSS/.github/.github/actions/octo-notify@v1` composite action in the `Send` step (guarded by `if: steps.build.outputs.skip != 'true'`). The forward is not dead.
- ✅ **No breakage when absent.** Because the callee marks it `required: false` and the comment explicitly says "Omit to disable," an empty/undefined secret degrades gracefully (triage disabled) rather than failing the workflow.
- ✅ **Capability-secret hygiene.** The URL is passed from `secrets.*`, never hardcoded, consistent with the callee's documented "capability secret — never hardcode it" guidance. The caller continues to map secrets explicitly (not `secrets: inherit`), which is the least-privilege approach.
- ✅ **YAML well-formed.** Correct indentation under the existing `secrets:` block; no syntax risk, no build break.

## Findings

No P0/P1 issues. The change is correct, minimal, and matches the reusable-workflow contract.

## Notes for a human reviewer (security-sensitive, non-blocking)

These are outside this diff and cannot be confirmed from the PR alone — worth a human glance given the security-sensitive classification, but none block merge:

- **Org secret must actually exist.** This PR wires the plumbing, but triage only runs if `TRIAGE_WEBHOOK_URL` is configured at the org/repo level with this repo granted access. If it is not set, the secret resolves to empty and triage stays silently disabled — i.e. the very symptom this PR targets would persist. Please confirm the org secret is populated and visible to this repo.
- **Downstream URL handling is trusted to the callee.** The actual egress safety (POST, retry/backoff, masking, any URL allowlist) lives in `octo-notify@v1`, not here. This PR correctly delegates transport; no concern within this diff.

## Verdict

APPROVED — correct minimal fix, cross-file contract verified, no blocking issues.

PR #74, opened 2026-06-10
feat(envelope): canonical response envelope types + wkhttp helpers (R1/R5)

liuooo:
## What

The OCTO API spec (octo-openapi-dev-skill, rules R1/R2/R5) mandates canonical response wire shapes — `{"data": ...}` for 2xx, `{"data": [...], "pagination": {...}}` for lists, `{"error": {...}}` for failures — and the spectral toolchain now lint-blocks anything else. But octo-lib offers no envelope abstraction: octo-server's 444 handlers have no compliant type to reference (~793 bare legacy calls), and octo-matter / octo-smart-summary each hand-roll their own response helpers (matter's hand-rolled shapes happen to match the spec — evidence the shared abstraction is needed).

This PR ships the official implementation of the spec's "recommended unified response abstraction" (api-spec §B). Additive only: no existing method changes behavior.

## How

- **`pkg/envelope/envelope.go`** (new): dependency-free wire-shape types — `Data[T]`, `EmptyResp` (`Data[EmptyResp]` → `{"data":{}}`), `CursorList[T]` / `OffsetList[T]` with `CursorPagination` (`has_more`, omitempty `next_cursor`) / `OffsetPagination` (`total`, `page`, `page_size`), and `Error` / `ErrorBody` (`code`, `message`, omitempty `details` / `hint`). Gin-free on purpose so non-wkhttp services can reference the same types in swag annotations. The doc comment states the invariant: the contract is the wire shape, not the type name — lint validates structure.
- **`pkg/wkhttp/response_envelope.go`** (new):
  - Single-object `Context` methods: `ResponseData` (200), `ResponseCreated` (201), `ResponseEmpty` (200 `{"data":{}}`). `ResponseEmpty` is deliberately NOT named `ResponseOK`: the legacy `ResponseOK` emits `{"status":200}` and existing clients depend on it.
  - List helpers as package-level **generic functions** `ResponseCursor[T]` / `ResponseOffset[T]` (methods cannot take type parameters): `[]T` is compiler-checked, the canonical `envelope.CursorList[T]` / `OffsetList[T]` types are emitted directly (no hand-copied shape), and a nil slice — Go's default "no results" state — is normalized once at the helper so empty lists always render `"data":[]`, never `null` (R1).
  - R5 query readers `GetCursorParams` / `GetOffsetParams` (`page_size` default 20, capped at 100; `page` default 1; invalid values fall back).
[...]

PR #79, opened 2026-06-18
fix(log): synchronize lazy logger init to remove data race

### LINKED ISSUE #77: Data race in octo-lib pkg/log: unsynchronized lazy logger init under concurrent logging
> **Note:** the root cause is in **octo-lib `pkg/log`**, not octo-server. Filed here for tracking — to be **moved to the octo-lib repo**. octo-server only carries a workaround (see below).

## Summary

`octo-lib/pkg/log` lazily configures its package-global loggers on first use **without synchronization**. When the first log call(s) in a process happen concurrently (before anything called `log.Configure`), multiple goroutines race on the lazy init. `go test -race` flags this and it intermittently fails CI.

## Where it surfaces

`modules/notify` `deliverNotification` fans out up to 20 goroutines that each log send failures via `n.Warn`. In the test binary the logger is not pre-configured, so the concurrent `n.Warn` calls race on the lazy init:

- Failing test: `modules/notify` `TestIntegration_Deliver_SendFailure_MarksFiltered`
- CI job: `Test` (`go test -race -shuffle=on`)

## Root cause (octo-lib `pkg/log/logger.go`)

```go
// package-global loggers — no mutex / sync.Once
var logger, errorLogger, warnLogger *zap.Logger

func Warn(msg string, fields ...zap.Field) {
    if warnLogger == nil {          // ~line 141: READ of the global
        Configure(NewOptions())     // ~line 142: Configure WRITES the globals (~line 81)
    }
    warnLogger.Warn(msg, fields...)
}
```

`Configure()` assigns `logger` / `errorLogger` / `warnLogger` with no lock. `Info` / `Debug` / `Error` / `Warn` all use the same unguarded `if X == nil { Configure() }` lazy pattern. Concurrent first-time callers read the global (nil check) while another writes it inside `Configure()` → data race.

Race detector evidence:

```
Read at 0x...  by goroutine 37:
   octo-lib/pkg/log.Warn()       logger.go:141   // if warnLogger == nil
Previous write at 0x...  by goroutine 36:
   octo-lib/pkg/log.Configure()  logger.go:81    // assigns warnLogger (called from Warn:142)
```

## Impact

- **Production: none.** `octo-server/main.go:82` calls `log.Configure()` once at startup (single-threaded) before any concurrency, so the globals are set and subsequent logging is concurrent-read-only (safe).
- **CI/tests: real, flaky.** `testutil.NewTestServer` does not call `log.Configure`, so the globals stay nil until the first log call. With `-shuffle=on`, whether a single-threaded log call runs before a concurrent fan-out varies → the race appears intermittently.

## Reproduction

```bash
# MySQL(root/demo, db test) + Redis on 127.0.0.1
OCTO_MASTER_KEY=0123456789abcdef0123456789abcdef \
  go test -race -shuffle=on -count=1 ./modules/notify/ -run TestIntegration_Deliver_SendFailure_MarksFiltered
# => WARNING: DATA RACE  (octo-lib/pkg/log.Warn vs log.Configure)
```

## Proper fix (octo-lib)

Synchronize the lazy init — e.g. guard `Configure()` (and the `if X == nil` fast-paths) with a `sync.Once` or a mutex, so concurrent first-time logging cannot race on the globals. (Reads after a one-time configure remain lock-free.)

## octo-server workaround (already done)

`modules/notify` no longer logs from inside the fan-out goroutines: the send error is carried out in `sendResult` and failures are logged serially in the drain loop. This removes the concurrent-logging trigger in notify. Branch: `fix/notify-log-race`. This is a local mitigation; it does not fix the systemic octo-lib race (any other concurrent-first-logging site remains latent in `-race` tests).

an9xyz:
## Summary

`pkg/log` configured its package-global loggers lazily on first use **without synchronization**. When the first log calls in a process happened concurrently (before anything called `Configure`), goroutines raced on the `if X == nil { Configure() }` lazy init: one read the nil global while another wrote it inside `Configure`. `go test -race` flagged this (e.g. `modules/notify` fanning out concurrent `Warn` calls) and it intermittently failed CI.

Closes #77.

## Root cause

```go
var logger, errorLogger, warnLogger *zap.Logger // unguarded globals

func Warn(msg string, fields ...zap.Field) {
    if warnLogger == nil {      // read of the global
        Configure(NewOptions()) // Configure writes the globals
    }
    warnLogger.Warn(msg, fields...)
}
```

Concurrent first-time callers read a global (the nil check) while another goroutine wrote it inside `Configure()` → data race. `Info` / `Debug` / `Error` / `Warn` all shared the same unguarded pattern.

- **Production: none.** `octo-server` calls `log.Configure()` once at startup (single-threaded) before any concurrency, so subsequent logging is concurrent read-only.
- **CI/tests: real, flaky.** Test servers don't pre-call `Configure`, so the globals stay nil until the first (possibly concurrent) log call; with `-shuffle=on` the race appears intermittently.

## Fix

- Serialize (re)configuration with a `sync.Mutex` (`configMu`); `Configure` delegates to an internal `configureLocked` that holds the lock.
- Publish all loggers as one `loggerSet` through a **single `atomic.Pointer`**, so the hot logging path is a single lock-free atomic load and readers always observe a fully-built set — never a half-initialized one.
- `current()` performs the one-time lazy init under `configMu` (double-checked) and returns the published set; `Info` / `Debug` / `Error` / `Warn` read from it.

Publishing one set (rather than several independent atomics gated by a single-logger sentinel) keeps the sentinel and the reads on the **same pointer**, so correctness no longer depends on the order in which loggers are stored. This mirrors the boxed `atomic.Pointer` pattern already used in `pkg/wkhttp`.

The public API (`Configure` / `Info` / `Debug` / `Error` / `Warn` / `TLog`) is unchanged, and lazy first-use plus explicit `Configure` semantics are preserved.

## Testing

- New `pkg/log/logger_concurrent_test.go` reproduces the race under `-race` by forcing the lazy path from many goroutines concurrently, and alongside an explicit `Configure`. It fails (DATA RACE) on the old code and passes on the fix.
- `go test -race -shuffle=on` — green across 20+ repeated runs (`-count=5` per run).
- `go build ./...` and `go vet ./pkg/log/` clean.

## Notes

- During review, `testLogger` was found to be pre-existing dead code (written, never read). Left untouched here to keep this change scoped to the concurrency fix; suggest a separate cleanup.

https://claude.ai/code/session_01FpoBeeNmZ1aH6gKjcWe7fx

---
_Generated by [Claude Code](https://claude.ai/code/session_01FpoBeeNmZ1aH6gKjcWe7fx)_

### DIFF
--- a/pkg/log/logger.go
+++ b/pkg/log/logger.go
@@ -43,9 +43,9 @@ func Configure(opts *Options) {
 	configureLocked(opts)
 }
 
-// configureLocked builds the loggers and publishes them as one set. Callers must
-// hold configMu.
-func configureLocked(opts *Options) {
+// configureLocked builds the loggers, publishes them as one set, and returns it.
+// Callers must hold configMu.
+func configureLocked(opts *Options) *loggerSet {
 	atom.SetLevel(opts.Level)
 
 	stdoutCore := zapcore.NewCore(
@@ -114,12 +114,14 @@ func configureLocked(opts *Options) {
 		warn = zap.New(warnCore)
 	}
 
-	loggers.Store(&loggerSet{
+	set := &loggerSet{
 		infoLogger:  info,
 		errorLogger: errLogger,
 		warnLogger:  warn,
 		testLogger:  test,
-	})
+	}
+	loggers.Store(set)
+	return set
 }
 
 // current returns the published logger set, lazily configuring it with default
@@ -132,11 +134,11 @@ func current() *loggerSet {
 		return ls
 	}
 	configMu.Lock()
-	if loggers.Load() == nil {
-		configureLocked(NewOptions())
+	defer configMu.Unlock()
+	if ls := loggers.Load(); ls != nil { // another goroutine configured while we waited
+		return ls
 	}
-	configMu.Unlock()
-	return loggers.Load()
+	return configureLocked(NewOptions())
 }
 
 func newEncoderConfig() zapcore.EncoderConfig {
--- a/pkg/log/logger.go
+++ b/pkg/log/logger.go
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"os"
 	"path"
+	"sync"
+	"sync/atomic"
 	"time"
 
 	"go.uber.org/zap"
@@ -11,23 +13,51 @@ import (
 	"gopkg.in/natefinch/lumberjack.v2"
 )
 
-var logger *zap.Logger
-var errorLogger *zap.Logger
-var warnLogger *zap.Logger
-var testLogger *zap.Logger
-var atom = zap.NewAtomicLevel()
+// loggerSet bundles the configured loggers so they can be published as a single
+// unit. Readers load the whole set through an atomic.Pointer, so they observe
+// either a fully-built set or none at all — never a half-initialized one. This
+// mirrors the boxed atomic.Pointer pattern already used in pkg/wkhttp.
+type loggerSet struct {
+	infoLogger  *zap.Logger
+	errorLogger *zap.Logger
+	warnLogger  *zap.Logger
+	testLogger  *zap.Logger
+}
 
+// The package loggers are configured lazily on first use (or eagerly via
+// Configure). configMu serializes (re)configuration so concurrent first-time
+// callers cannot race on the lazy init, while loggers is read lock-free on the
+// hot logging path once configured. atom is shared by the cores and is itself
+// goroutine-safe.
+var (
+	configMu sync.Mutex
+	loggers  atomic.Pointer[loggerSet]
+	atom     = zap.NewAtomicLevel()
+)
+
+// Configure (re)configures the package loggers. It is safe to call concurrently
+// and alongside lazy first-use initialization.
 func Configure(opts *Options) {
+	configMu.Lock()
+	defer configMu.Unlock()
+	configureLocked(opts)
+}
+
+// configureLocked builds the loggers and publishes them as one set. Callers must
+// hold configMu.
+func configureLocked(opts *Options) {
 	atom.SetLevel(opts.Level)
-	core := zapcore.NewCore(
+
+	stdoutCore := zapcore.NewCore(
 		zapcore.NewJSONEncoder(newEncoderConfig()),
 		zapcore.NewMultiWriteSyncer(zapcore.AddSync(os.Stdout)),
 		atom,
 	)
+	var test *zap.Logger
 	if opts.LineNum {
-		testLogger = zap.New(core, zap.AddCaller())
+		test = zap.New(stdoutCore, zap.AddCaller())
 	} else {
-		testLogger = zap.New(core)
+		test = zap.New(stdoutCore)
 	}
 
 	infoWriter := zapcore.AddSync(&lumberjack.Logger{
@@ -36,15 +66,16 @@ func Configure(opts *Options) {
 		MaxBackups: 3,
 		MaxAge:     28, // days
 	})
-	core = zapcore.NewCore(
+	infoCore := zapcore.NewCore(
 		zapcore.NewJSONEncoder(newEncoderConfig()),
 		zapcore.NewMultiWriteSyncer(zapcore.AddSync(os.Stdout), zapcore.AddSync(infoWriter)),
 		atom,
 	)
+	var info *zap.Logger
 	if opts.LineNum {
-		logger = zap.New(core, zap.AddCaller())
+		info = zap.New(infoCore, zap.AddCaller())
 	} else {
-		logger = zap.New(core)
+		info = zap.New(infoCore)
 	}
 
 	errorWriter := zapcore.AddSync(&lumberjack.Logger{
@@ -53,15 +84,16 @@ func Configure(opts *Options) {
 		MaxBackups: 3,
 		MaxAge:     28, // days
 	})
-	core = zapcore.NewCore(
+	errorCore := zapcore.NewCore(
 		zapcore.NewJSONEncoder(newEncoderConfig()),
 		zapcore.NewMultiWriteSyncer(zapcore.AddSync(os.Stdout), zapcore.AddSync(errorWriter)),
 		zap.ErrorLevel,
 	)
+	var errLogger *zap.Logger
 	if opts.LineNum {
-		errorLogger = zap.New(core, zap.AddCaller())
+		errLogger = zap.New(errorCore, zap.AddCaller())
 	} else {
-		errorLogger = zap.New(core)
+		errLogger = zap.New(errorCore)
 	}
 
 	warnWriter := zapcore.AddSync(&lumberjack.Logger{
@@ -70,17 +102,41 @@ func Configure(opts *Options) {
 		MaxBackups: 3,
 		MaxAge:     28, // days
 	})
-	core = zapcore.NewCore(
+	warnCore := zapcore.NewCore(
 		zapcore.NewJSONEncoder(newEncoderConfig()),
 		zapcore.NewMultiWriteSyncer(zapcore.AddSync(os.Stdout), zapcore.AddSync(warnWriter)),
 		zap.WarnLevel,
 	)
+	var warn *zap.Logger
 	if opts.LineNum {
-		warnLogger = zap.New(core, zap.AddCaller())
+		warn = zap.New(warnCore, zap.AddCaller())
 	} else {
-		warnLogger = zap.New(core)
+		warn = zap.New(warnCore)
 	}
 
+	loggers.Store(&loggerSet{
+		infoLogger:  info,
+		errorLogger: errLogger,
+		warnLogger:  warn,
+		testLogger:  test,
+	})
+}
+
+// current returns the published logger set, lazily configuring it with default
+// options on first use. Because the set is published as a single atomic pointer,
+// the returned value always has every logger built — there is no window where one
+// logger is set and another is still nil. The lock-free fast path keeps the hot
+// logging path free of contention once configured.
+func current() *loggerSet {
+	if ls := loggers.Load(); ls != nil {
+		return ls
+	}
+	configMu.Lock()
+	if loggers.Load() == nil {
+		configureLocked(NewOptions())
+	}
+	configMu.Unlock()
+	return loggers.Load()
 }
 
 func newEncoderConfig() zapcore.EncoderConfig {
@@ -110,38 +166,22 @@ func timeEncoder(t time.Time, enc zapcore.PrimitiveArrayEncoder) {
 
 // Info Info
 func Info(msg string, fields ...zap.Field) {
-
-	if logger == nil {
-		Configure(NewOptions())
-	}
-	logger.Info(msg, fields...)
+	current().infoLogger.Info(msg, fields...)
 }
 
 // Debug Debug
 func Debug(msg string, fields ...zap.Field) {
-
-	if logger == nil {
-		Configure(NewOptions())
-	}
-	logger.Debug(msg, fields...)
-
+	current().infoLogger.Debug(msg, fields...)
 }
 
 // Error Error
 func Error(msg string, fields ...zap.Field) {
-	if errorLogger == nil {
-		Configure(NewOptions())
-	}
-	errorLogger.Error(msg, fields...)
+	current().errorLogger.Error(msg, fields...)
 }
 
 // Warn Warn
 func Warn(msg string, fields ...zap.Field) {
-
-	if warnLogger == nil {
-		Configure(NewOptions())
-	}
-	warnLogger.Warn(msg, fields...)
+	current().warnLogger.Warn(msg, fields...)
 }
 
 // Log Log
--- a/pkg/log/logger_concurrent_test.go
+++ b/pkg/log/logger_concurrent_test.go
@@ -37,6 +37,10 @@ func TestConcurrentLazyInit(t *testing.T) {
 	}
 	close(start)
 	wg.Wait()
+
+	if loggers.Load() == nil {
+		t.Fatal("loggers should be configured after concurrent lazy-init use")
+	}
 }
 
 // TestConcurrentConfigureAndLog ensures explicit Configure is safe alongside
@@ -63,4 +67,8 @@ func TestConcurrentConfigureAndLog(t *testing.T) {
 	}
 	close(start)
 	wg.Wait()
+
+	if loggers.Load() == nil {
+		t.Fatal("loggers should be configured after concurrent Configure/log use")
+	}
 }
--- a/pkg/log/logger_concurrent_test.go
+++ b/pkg/log/logger_concurrent_test.go
@@ -1,0 +1,66 @@
+package log
+
+import (
+	"sync"
+	"testing"
+)
+
+// resetLoggers forces the lazy-init path by clearing the published logger set,
+// independent of test ordering or -shuffle. It is white-box (same package) on
+// purpose: the globals are unexported.
+func resetLoggers() {
+	configMu.Lock()
+	defer configMu.Unlock()
+	loggers.Store(nil)
+}
+
+// TestConcurrentLazyInit reproduces issue #77: before the sync fix, the first
+// concurrent log calls raced on the unsynchronized lazy init (log.Warn reading a
+// global that Configure was concurrently writing). Run with -race to guard
+// against regressions.
+func TestConcurrentLazyInit(t *testing.T) {
+	resetLoggers()
+
+	const goroutines = 50
+	var wg sync.WaitGroup
+	start := make(chan struct{})
+	for i := 0; i < goroutines; i++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			<-start // release all at once to maximize contention on first-use init
+			Info("concurrent info")
+			Warn("concurrent warn")
+			Error("concurrent error")
+			Debug("concurrent debug")
+		}()
+	}
+	close(start)
+	wg.Wait()
+}
+
+// TestConcurrentConfigureAndLog ensures explicit Configure is safe alongside
+// concurrent logging (the lazy init and an explicit reconfigure both publish the
+// globals; they must not race).
+func TestConcurrentConfigureAndLog(t *testing.T) {
+	resetLoggers()
+
+	const goroutines = 25
+	var wg sync.WaitGroup
+	start := make(chan struct{})
+	for i := 0; i < goroutines; i++ {
+		wg.Add(2)
+		go func() {
+			defer wg.Done()
+			<-start
+			Configure(NewOptions())
+		}()
+		go func() {
+			defer wg.Done()
+			<-start
+			Warn("racing with configure")
+		}()
+	}
+	close(start)
+	wg.Wait()
+}


lml2468 (dismissed):
✅ **APPROVE** — `fix(log)` synchronize lazy logger init to remove data race

Verified locally on head `e6eb27f` (go 1.26.4) — and I reproduced both directions:

- 🟢 **The race is real on base.** Reconstructed the pre-fix `logger.go` (base `f62e626`) and ran a public-API-only probe (50 goroutines hitting `Info/Warn/Error/Debug` cold) under `-race` → **WARNING: DATA RACE** (many). Confirms #77 is a genuine bug, not theoretical.
- 🟢 **The fix removes it.** `go test -race -shuffle=on -count=5 ./pkg/log/` → green across all runs. `go build ./...` ok, `go vet ./pkg/log/` clean, `gofmt -l` clean.

**Why the fix is correct**
- 🟢 Publishing one `loggerSet` through a single `atomic.Pointer` is the right call — readers observe a fully-built set or nothing, never a half-initialized one. Avoids the subtler bug of N independent atomics gated by one sentinel where store-ordering matters.
- 🟢 `current()` is textbook double-checked locking: lock-free fast path (`loggers.Load()`), and the slow path re-checks `loggers.Load() == nil` **under** `configMu` before configuring — so two concurrent first-callers can't both build. Correct.
- 🟢 Hot logging path is now a single atomic load + method call, no mutex contention once configured. Good.
- 🟢 Behavior preserved: `Debug` routes through `infoLogger` (matches the old `logger` global), lazy-first-use + explicit `Configure` semantics intact, public API unchanged. `TLog` methods delegate to the package funcs, so they pick up the fix for free — no stale global refs remain (grepped).
- 🟢 Mirrors the existing boxed-`atomic.Pointer` pattern in `pkg/wkhttp` — consistent with the codebase.
- 🟢 `Closes #77` present → `check-sprint` is **green** here (unlike #74/#76).

**Nits (non-blocking)**
- 🟡 `TestConcurrentConfigureAndLog` has no post-condition assertion — under `-race` its value is the race check itself, which is fine, but it would pass even if logging silently no-op'd. Optional: assert `current()` non-nil at the end. Minor.
- 🟡 The new tests rely on `-race` to have teeth; without `-race` they're near-no-ops. CI runs `-race` (per PR notes), so this is covered — just worth keeping `-race` in the `log` package's CI lane permanently so a future regression can't slip in green.
- 🟡 PR notes `testLogger` is pre-existing dead code (built in `configureLocked`, never read via a public accessor). Correctly left out of scope; agree a separate cleanup is the right move.

Verdict: **approve**. Real race, minimal correct fix, regression test that genuinely fails on the old code, no behavior change. Clean work.

Jerry-Xin (dismissed):
The PR is in scope for `Mininglamp-OSS/octo-lib`: it fixes concurrency behavior in the existing `pkg/log` package.

🔴 Blocking

None.

💬 Non-blocking

None.

✅ Highlights

🔵 `pkg/log/logger.go:32-43` correctly serializes explicit reconfiguration with `configMu`.

🔵 `pkg/log/logger.go:117-139` publishes the logger set through one `atomic.Pointer`, avoiding mixed or partially initialized logger state on concurrent first use.

🔵 `pkg/log/logger_concurrent_test.go:21-66` adds focused race-regression coverage for lazy initialization and concurrent `Configure` plus logging.

Verification performed: `go test ./pkg/log`, `go test -race ./pkg/log`, and `go test ./...` all passed.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# octo-lib PR#79 Review Report — fix(log): synchronize lazy logger init to remove data race
**Head SHA:** `e6eb27f61b8d118c69d97cb4242fd14ccd716e3c`
**Files:** `pkg/log/logger.go` (+77/−37), `pkg/log/logger_concurrent_test.go` (+66, new)

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| Data race fixed | ✅ | `current()` uses double-checked locking via `configMu` + `atomic.Pointer[loggerSet]`; single atomic publish of all loggers as one struct. `go test -race -count=3 ./pkg/log/` passes clean. |
| Public API preserved | ✅ | `Configure`, `Info`, `Debug`, `Error`, `Warn`, `TLog` signatures unchanged. Callers in `pkg/wkhttp/ratelimit.go`, `common/msg.go` need no changes. |
| Lazy init preserved | ✅ | `current()` fast-path: `loggers.Load()` returns non-nil → lock-free. Slow-path: `configMu.Lock()` + double-check + `configureLocked(NewOptions())`. |
| Re-configuration safe | ✅ | `Configure()` always acquires `configMu`; concurrent `Configure` calls serialized. Atomic `Store` replaces the full set atomically. |
| Build / vet / tests | ✅ | `go build ./...` clean, `go vet ./pkg/log/` clean, `go test -race -count=3 ./pkg/log/` passes. |
| Half-initialized window eliminated | ✅ | Base code wrote globals one-by-one (`testLogger`, `logger`, `errorLogger`, `warnLogger`). Fix builds all loggers as locals, then publishes one `loggerSet` via single `atomic.Store` — no partial-visibility window. |

## 2. Findings

### No P0/P1 findings.

### P2 — `current()` slow-path does redundant atomic load

**Diff-scope:** new (introduced by this PR's `current()` implementation)

`current()` at `logger.go:131-139`:
```go
func current() *loggerSet {
    if ls := loggers.Load(); ls != nil {
        return ls
    }
    configMu.Lock()
    if loggers.Load() == nil {
        configureLocked(NewOptions())
    }
    configMu.Unlock()
    return loggers.Load()  // ← third atomic load
}
```

After `configureLocked` returns, the set was just stored. The final `loggers.Load()` is a third atomic load. In production this is negligible (atomic loads are ~1ns), but `configureLocked` could return the built set directly, avoiding the extra load and making the invariant ("the set we just built is the one we return") more explicit.

**Severity:** P2 (micro-optimization / clarity). Non-blocking.

### P2 — `testLogger` field is dead code (pre-existing, acknowledged)

**Diff-scope:** pre-existing (base code wrote `testLogger` but no public function reads it)

`loggerSet.testLogger` is populated in `configureLocked()` (line ~63-68) but no public API function consumes it. The PR body explicitly notes this and defers cleanup. Agreed — out of scope for this fix.

**Severity:** P2 (pre-existing dead code, not amplified by this PR).

### Nit — Test assertions are race-only (no behavioral check)

**Diff-scope:** new (test file)

`TestConcurrentLazyInit` and `TestConcurrentConfigureAndLog` verify absence of data races but don't assert that log output is correct or that `current()` returns non-nil. This is standard practice for `-race` regression tests and acceptable here. A minor enhancement would be `if ls := loggers.Load(); ls == nil { t.Fatal("loggers not configured after concurrent use") }` after `wg.Wait()`.

**Severity:** nit. Non-blocking.

## 3. Suggestions

1. **Optional refactor:** Have `configureLocked` return `*loggerSet` so `current()` can capture and return it directly, eliminating the third atomic load:
   ```go
   func configureLocked(opts *Options) *loggerSet {
       // ... build loggers ...
       set := &loggerSet{...}
       loggers.Store(set)
       return set
   }
   ```
   This makes the "what you build is what you return" invariant explicit in the type system.

2. **Future cleanup (separate PR):** Remove `testLogger` from `loggerSet` or wire it to a public API, as noted in the PR body.

## 4. Additional Observations

- **`atom` (zap.AtomicLevel) is shared across reconfigurations:** All logger cores reference the same `atom`. When `Configure()` is called again, `atom.SetLevel()` changes the level for all previously-created cores too. This is pre-existing behavior, not introduced by this PR. `zap.AtomicLevel` uses internal atomics, so `SetLevel` concurrent with log writes is safe.
- **Lumberjack writer lifecycle:** Each `Configure()` creates new `lumberjack.Logger` writers. Old writers become unreachable and are GC'd. Pending buffered writes on old writers may be lost. Pre-existing behavior, unchanged by this PR.

## 5. Data-Flow Trace

| Consumer | Data consumed | Upstream source | Verified? |
|----------|--------------|-----------------|-----------|
| `Info()` → `current().infoLogger` | `loggerSet.infoLogger` | `configureLocked()` builds `info` local → `loggers.Store(&loggerSet{infoLogger: info, ...})` | ✅ Non-nil after `current()` returns |
| `Debug()` → `current().infoLogger` | Same as Info | Same | ✅ |
| `Error()` → `current().errorLogger` | `loggerSet.errorLogger` | `configureLocked()` builds `errLogger` local → Store | ✅ Non-nil after `current()` returns |
| `Warn()` → `current().warnLogger` | `loggerSet.warnLogger` | `configureLocked()` builds `warn` local → Store | ✅ Non-nil after `current()` returns |
| `TLog.Info/Debug/Error/Warn` | Delegates to package-level `Info/Debug/Error/Warn` | Same as above | ✅ Transitive safety |
| `current()` fast path | `loggers.Load()` | `atomic.Pointer[loggerSet]` — set by `configureLocked` or `Configure` | ✅ Returns non-nil or falls through to slow path |
| `current()` slow path | `configureLocked(NewOptions())` → `loggers.Store(...)` | `NewOptions()` returns `&Options{}` (zero-value defaults) | ✅ Always stores non-nil set |

**Key invariant verified:** `configureLocked()` always ends with `loggers.Store(&loggerSet{...})` where all four fields are assigned from local vars built unconditionally above. There is no early-return or conditional that could produce a nil field within the set.

## 6. R5 Blind-Spot Checklist

- **C1 (dual-path parity):** N/A — no symmetric add/remove or create/delete paths. Single init path (`current()`) + single reconfigure path (`Configure()`), both funnel through `configureLocked()` under `configMu`.
- **C2 (control-flow ordering / nested reuse):** N/A — `current()` is not nested or re-entrant. `Configure()` is not called from within `current()` (it calls `configureLocked` directly), so no recursive lock deadlock.
- **C3 (authorization boundary ≠ capability boundary):** N/A — logging package has no authorization boundary.
- **C4 (authorization lifecycle / container-member cascade):** N/A — no authorization hierarchy.

## 7. Cross-Round Blocker Recheck

---

**[Octo-Q] verdict: APPROVE**

The fix correctly eliminates the data race using the standard Go pattern (mutex + atomic.Pointer with double-checked locking). The single-pointer publication of a bundled `loggerSet` is a clean improvement over the base code's per-global nil-check pattern. Public API is preserved, tests reproduce and verify the fix under `-race`, and build/vet are clean. No P0/P1 findings. Two P2 items (redundant atomic load in `current()` slow path, pre-existing dead `testLogger`) are non-blocking.

yujiawei (dismissed):
# Code Review — PR #79 (octo-lib)

**Scope reviewed:** `pkg/log/logger.go`, `pkg/log/logger_concurrent_test.go` at head `e6eb27f` against `main`.

## Verdict: APPROVED ✅

The fix is correct and well-scoped. The data race described in #77 is genuinely eliminated, behavioral equivalence with the old API is preserved, and the chosen pattern (single `atomic.Pointer[loggerSet]` published under a `sync.Mutex`, with double-checked lazy init in `current()`) is the right one — it matches the existing `pkg/wkhttp` boxed-pointer precedent. No P0/P1 (correctness/security/blocker) issues found. Remaining notes are P2 advisories that do not block merge.

## 1. Verification

- ✅ **Race eliminated.** The unguarded `if X == nil { Configure() }` lazy-init pattern is replaced by a single published `loggerSet`. Readers do one lock-free `atomic.Load`; the set is fully constructed before `Store`, so a reader can never observe a half-initialized set. `current()` uses correct double-checked locking: lock-free fast path, then `configMu` + re-check nil + `configureLocked` on the slow path.
- ✅ **Behavioral equivalence.** Public API (`Configure` / `Info` / `Debug` / `Error` / `Warn` / `TLog`) unchanged. `Debug` previously read the info `logger` global and now reads `current().infoLogger` — a faithful rename, same level routing. Lazy-first-use and explicit-`Configure` semantics preserved.
- ✅ **`atom` safety.** `zap.AtomicLevel` is goroutine-safe; `SetLevel` is called under `configMu`. No race on the shared level.
- ✅ **Tests.** `go build ./...`, `go vet ./pkg/log/` clean. `go test -race -count=1 ./pkg/log/` green, including the two new concurrency tests. The white-box `resetLoggers()` correctly forces the lazy path regardless of test ordering/`-shuffle`.
- ✅ **No cross-file fallout.** No external references to the removed unexported globals; all call sites go through the unchanged public functions.

## 2. Findings (P2 — advisory, non-blocking)

**P2-1 — `lumberjack.Logger` instances are not closed on reconfigure (pre-existing).**
Each `configureLocked` call builds fresh `lumberjack.Logger` writers (`logger.go:63`, `:81`, `:99`) without closing the previously published ones. Repeated `Configure` calls (e.g. dynamic level reloads) would leak file handles, and multiple live rotators on the same paths can contend during rotation. **This is not introduced by this PR** — the old `Configure` had the identical pattern — and lumberjack opens the FD lazily on first `Write`, so in production (single startup `Configure`) the impact is nil. Out of scope for a race fix; worth a separate follow-up if `Configure` is ever called repeatedly at runtime.

**P2-2 — shared `atom` blurs snapshot isolation.**
`atom` is a single global mutated by `atom.SetLevel` on every `configureLocked`, and is shared across the published `loggerSet` cores. A reconfigure changes the effective level of the *previous* set still in use by in-flight goroutines, so the `atomic.Pointer` snapshot is not fully immutable w.r.t. level. Benign today (level rarely changes after startup) and not a race, but scoping a fresh `AtomicLevel` per set would make the snapshot guarantee complete.

**P2-3 — `testLogger` remains dead code.**
Acknowledged in the PR description and deliberately deferred. Reasonable to keep this change scoped; suggest a separate cleanup.

**P2-4 — concurrency tests amplify P2-1.**
`TestConcurrentConfigureAndLog` calls `Configure` 25×, leaving orphaned rotators in the test process. Harmless at this scale but compounds with P2-1 if expanded. Optional.

## 3. Suggestions

- If runtime reconfiguration is ever a real use case, address P2-1/P2-2 together: close the prior `loggerSet` writers after `Store`, and give each set its own `AtomicLevel`.
- None of the above is required for this PR.

## 4. Coverage / blind spots

- Verified there are no other exported logging entry points (`Fatal`/`Panic`/top-level `Log`) that bypass `current()` — all of `Info`/`Debug`/`Error`/`Warn` (and `TLog` wrappers) route through it.
- Default `NewOptions()` writes `info.log`/`warn.log`/`error.log` into the CWD (empty `LogDir`); the tests therefore drop log files in `pkg/log/`. Pre-existing behavior, not introduced here.
- Not assessed (out of diff): application-shutdown `Sync()`/`Close()` lifecycle — unchanged by this PR.

lml2468 (approved):
✅ **APPROVE (re-review)** — head `b5e7245` (was `e6eb27f`)

The new commit is a clean response to the prior non-blocking nits — nothing else changed, no behavior change:

1. 🟢 **`current()` slow path no longer does a redundant `atomic.Load()`** — `configureLocked` now returns the built `*loggerSet`, and the slow path returns it directly. Switched to `defer configMu.Unlock()` and added a proper double-check-after-lock (`if ls := loggers.Load(); ls != nil { return ls }`) so a goroutine that lost the init race returns the winner's set instead of rebuilding. This is exactly the textbook DCL shape — cleaner than before.
2. 🟢 **Both concurrent tests now assert `loggers.Load() != nil`** after the goroutines join — addresses the "race-only, no behavior assertion" nit. Tests now fail if logging silently no-op'd, not just on a race.

Re-verified locally on `b5e7245` (go 1.26.4):
- `go test -race -shuffle=on -count=5 ./pkg/log/` → green every run
- `go build ./...` ok · `go vet ./pkg/log/` clean · `gofmt -l` clean

Correctness unchanged from the version I already proved (real race on base via `-race`, eliminated by the fix). The refactor only tightens the slow path and strengthens the tests. No new concerns.

Verdict: **approve**. Note: head moved to `b5e7245`, so any prior approvals on `e6eb27f` may be stale under branch protection and need re-voting on the new head.

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-lib PR#79 Review Report
**PR:** Mininglamp-OSS/octo-lib #79 — fix(log): synchronize lazy logger init to remove data race
**Head SHA:** `b5e72455f48968a6f5071a8d19231b83ea925222`
**Base:** `f62e626` (main)
**Files changed:** 2 (+153 / −37)
**Commits:** 2 — `e6eb27f` (sync fix) + `b5e7245` (refactor: return set from configureLocked)

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| Data race fix correctness | ✅ | `current()` double-checked locking at logger.go:131–139; `atomic.Pointer[loggerSet]` publish at :125 |
| Atomicity of logger set | ✅ | All 4 loggers built as locals, bundled into `loggerSet`, single `loggers.Store(set)` at :125 |
| Mutex discipline | ✅ | `Configure()` acquires `configMu` at :44; `current()` acquires at :134; `configureLocked` asserts caller holds lock |
| Hot-path lock-free | ✅ | `current()` fast path `loggers.Load()` at :132 — no lock after first init |
| Test coverage | ✅ | `TestConcurrentLazyInit` (50 goroutines) + `TestConcurrentConfigureAndLog` (25×2 goroutines) |
| `go test -race ./...` | ✅ | All packages pass (pkg/log 1.029s) |
| `go vet ./pkg/log/` | ✅ | Clean |
| API compatibility | ✅ | `Info/Debug/Error/Warn/Configure/TLog` signatures unchanged; callers in `pkg/wkhttp`, `config/`, `common/` unaffected |

## 2. Findings

### No P0 or P1 issues found.

### P2 — Pre-existing observations (not introduced by this PR)

**P2-1: `testLogger` field is dead code**
- `loggerSet.testLogger` (logger.go:24) is built and stored but never consumed by any exported function. No `Test()` or similar accessor exists.
- **Diff-scope:** pre-existing — the base version had the same unused `var testLogger *zap.Logger`.
- **Suggestion:** Consider removing the field or adding an exported accessor if it's planned for future use.

**P2-2: `timeEncoder` function is dead code**
- `timeEncoder` (logger.go:163) is defined but never referenced. The encoder config uses an inline lambda instead.
- **Diff-scope:** pre-existing — unchanged by this PR.

**P2-3: `Configure(nil)` panics**
- `configureLocked` dereferences `opts.Level` and `opts.LogDir` without nil-checking `opts`. Same as base behavior.
- **Diff-scope:** pre-existing.

## 3. Data-Flow Trace

| Consumer | Data consumed | Upstream source | Verified? |
|----------|--------------|-----------------|-----------|
| `Info()` → `current().infoLogger.Info()` | `infoLogger` field | `configureLocked()` builds `info` local (both branches of `opts.LineNum` assign it), stores in `loggerSet.infoLogger` at :119 | ✅ Always non-nil |
| `Error()` → `current().errorLogger.Error()` | `errorLogger` field | Same pattern, :120 | ✅ Always non-nil |
| `Warn()` → `current().warnLogger.Warn()` | `warnLogger` field | Same pattern, :121 | ✅ Always non-nil |
| `Debug()` → `current().infoLogger.Debug()` | `infoLogger` field | Same as Info | ✅ Always non-nil |
| `current()` fast path | `loggers.Load()` | `loggers.Store(set)` at :125 under `configMu` | ✅ Atomic publish; memory barrier ensures all fields visible |
| `current()` slow path | `configureLocked(NewOptions())` return value | Builds complete set, stores, returns pointer | ✅ Never returns nil |
| `Configure()` re-config | New `loggerSet` replaces old via `Store()` | Old set GC'd when no goroutine holds it | ✅ Safe swap; `atom` shared by value, level change visible to all cores |

## 4. Suggestions

- Consider removing `testLogger` and `timeEncoder` dead code in a follow-up cleanup.
- No blocking changes requested.

## 5. R5 Blind-Spot Checklist

- **C1 (dual-path parity):** N/A — no symmetric add/remove or create/delete paths in this logging utility.
- **C2 (control-flow ordering / nesting):** Clear — `current()` is the single entry point; no nested re-invocation path. `configureLocked` is only called under `configMu`. No double-invocation risk.
- **C3 (authorization boundary):** N/A — no auth/permission model in a logging package.
- **C4 (authorization lifecycle):** N/A — no container/member state hierarchy.

## 6. Cross-Round Blocker Recheck (R6)

---

**[Octo-Q] verdict: APPROVE** — The data race fix is correct, idiomatic, and well-tested. The `sync.Mutex` + `atomic.Pointer[loggerSet]` + double-checked locking pattern is the standard Go solution. All loggers are atomically published as a single set, eliminating the half-initialized window. Tests pass with `-race`. Only pre-existing P2 observations (dead code) noted.

Jerry-Xin (approved):
Summary: The PR is in scope for `octo-lib` and correctly fixes the `pkg/log` lazy initialization race without changing the public API.

🔴 Blocking

None.

💬 Non-blocking

None.

✅ Highlights

- `pkg/log/logger.go:117` publishes all logger instances through one `atomic.Pointer[loggerSet]`, which avoids readers observing a partially updated global logger state.
- `pkg/log/logger.go:132` uses a double-checked lazy initialization path under `configMu`, keeping the hot path lock-free after setup.
- `pkg/log/logger_concurrent_test.go:21` and `pkg/log/logger_concurrent_test.go:49` cover both concurrent lazy initialization and concurrent explicit `Configure` with logging.

Verification performed:

- `go test ./...`
- `go test -race ./...`

yujiawei (approved):
# Code Review — PR #79 (octo-lib)

**Scope:** Reviewed at head SHA `b5e72455f48968a6f5071a8d19231b83ea925222`, against merge-base `f62e626` (clean, branch is 2 commits ahead of `main`, `MERGEABLE`).

`fix(log): synchronize lazy logger init to remove data race` — replaces four unguarded package-global `*zap.Logger`s and their `if x == nil { Configure() }` lazy-init pattern with a single `atomic.Pointer[loggerSet]` published under a `sync.Mutex`. Net `+153/-37` across `pkg/log/logger.go` and a new white-box `logger_concurrent_test.go`.

## 1. Verification

| Item | Result | Evidence |
|---|---|---|
| Root cause is real | ✅ | `logger.go` (pre-fix): `Info/Debug/Error/Warn` each read an unguarded global in the nil-check while a concurrent `Configure` writes it — a genuine read/write data race on first concurrent use. |
| Concurrency primitives correct | ✅ | `current()` (`logger.go:135-148`) is a textbook double-checked load: lock-free `loggers.Load()` fast path → `configMu.Lock()` → re-check `Load()` → `configureLocked`. Publish-as-one-`*loggerSet` via `atomic.Store`/`Load` gives a correct happens-before edge under Go's memory model, so readers never observe a half-built set. |
| Hot path stays lock-free | ✅ | `Info/Debug/Error/Warn` reduce to one atomic load + method call; `atom` (`zap.AtomicLevel`) is itself goroutine-safe. |
| Behavior preserved | ✅ | Public API signatures unchanged; `Debug` routes to `infoLogger`, matching the old `logger` global; lazy first-use + explicit `Configure` semantics intact. |
| Builds & vets clean | ✅ | `go build ./...` and `go vet ./pkg/log/` clean (local, go1.26). |
| Test reproduces the race / guards regression | ✅ | `go test -race -shuffle=on -count=3 ./pkg/log/` green across repeated runs locally. `resetLoggers()` correctly clears state under `configMu` (`logger_concurrent_test.go:11-15`). The two tests exercise the concurrent lazy-init path and concurrent `Configure`+log. |

The fix is correct and the chosen "publish one set through a single pointer" approach (mirroring `pkg/wkhttp`'s boxed `atomic.Pointer`) is the right one — it removes any dependence on store-ordering between individual loggers.

## 2. Findings

No blocking issues (no P0/P1). The items below are **pre-existing** behaviors that this PR neither introduces nor is obligated to fix; noting them as non-blocking suggestions.

### P2 — `lumberjack` file descriptors not closed on re-`Configure` (pre-existing)
`configureLocked` builds fresh `lumberjack.Logger` instances on every call (`logger.go:62-110`) and never closes the previous set's writers. Repeated `Configure` calls (e.g. a future config hot-reload) would abandon FDs to GC finalizers, and two live `lumberjack` instances writing the same file would conflict on size-based rotation. **This is unchanged from the prior code** — the old `Configure` had the identical pattern — and there is currently no repeated/concurrent `Configure` caller in the tree, so it is not a regression and not a blocker. Worth a follow-up if hot-reload is ever added (close old writers, or reuse the sink).

### P2 — `Configure(nil)` panics (pre-existing)
`configureLocked` dereferences `opts.Level` immediately (`logger.go:49`) with no nil guard, so an external `Configure(nil)` panics. Same as the pre-fix code; a one-line `if opts == nil { opts = NewOptions() }` would harden the exported entry point. Non-blocking.

### Nit — `testLogger` / `stdoutCore` remain dead code
The `loggerSet.testLogger` field and its `stdoutCore` are built on every `configureLocked` but never read (no exported accessor). The PR description already acknowledges this and scopes it out, which is reasonable; flagging only so the eventual cleanup also drops the now-unused build cost.

## 3. Suggestions
- If a config hot-reload path is introduced later, pair it with closing the previous `loggerSet`'s `lumberjack` writers to avoid the FD/rotation issue above.
- Consider a separate cleanup PR for `testLogger` dead code (as the author suggests).

## 4. Coverage / blind spots
- Verified every public logging entry point (`Info`/`Debug`/`Error`/`Warn` and the `TLog` wrapper) routes through `current()` — no path still touches a raw global. The `Log` interface is a type declaration with no global state; `testLogger` is unread dead code. So the truncated-diff concern about other functions bypassing the safe accessor does not hold here.
- This review covers `pkg/log` only; no downstream consumer in this repo calls `Configure` more than once, so the pre-existing re-`Configure` caveats have no live trigger today.
- Independent second-opinion passes were incorporated: one line-level pass found no remaining races or incorrect memory publication; a long-context pass raised the FD-leak and nil-`opts` items, both verified here as pre-existing and downgraded to non-blocking. `-race` could not run in one sandbox; it was run locally instead (green).

## Verdict
**APPROVED.** The data-race fix is correct, well-tested, preserves the public API, and follows an established in-repo pattern. The remaining items are pre-existing, non-blocking, and out of scope for this change.

PR #78, opened 2026-06-18
feat(searchmsg): bump contract to v2 with SpaceID/Visibles/MessageSeq (sprint-w25)

yujiawei:
## What

Bump the `searchmsg` Kafka contract to **SchemaVersion 2** and add three reader-required safety/correctness fields to `Message`:

| Go field | JSON (Kafka wire) | type | purpose |
|---|---|---|---|
| `SpaceID` | `space_id,omitempty` | `string` | space isolation; reader fail-closed on empty for p2p |
| `Visibles` | `visibles,omitempty` | `[]string` | group-message visibility allowlist; reader **fail-OPEN** on empty (the gap this closes) |
| `MessageSeq` | `message_seq,omitempty` | `uint64` | channel-offset / clear-conversation gate |

## Why

The downstream indexer's `esindex.Doc` already declares these three fields and gates real-time write on them (`SafetyFieldsSchemaVersion == 2`, checked via `searchmsg.SchemaVersion >= 2`). Until the contract carries them, the real-time path writes empty `visibles`, which the reader treats as fail-OPEN (normal members can search admin-only group system messages). Bumping the contract is the prerequisite that lets the indexer repin and auto-unblock that path.

## Alignment

Field **Go types** match `esindex.Doc` exactly (`string` / `[]string` / `uint64`), so the indexer's `DocFromMessage` maps with no casts after it repins.

**JSON tags stay snake_case** to match the existing 11 contract fields (`message_id`, `channel_id`, `msg_timestamp`, …). `esindex.Doc`'s camelCase form (`spaceId`, `messageSeq`) is a **separate ES wire boundary**: the consumer unmarshals Kafka bytes into `searchmsg.Message` (snake_case), then `DocFromMessage` bridges at the Go-field level into `Doc` (camelCase) for OpenSearch. The two encodings are intentionally distinct and never pass JSON through directly.

## Scope

Contract-only. No producer/indexer change is required to compile — the fields are additive with `omitempty`. Producer enrichment and indexer repin land separately.

## Tests

`go test ./contract/searchmsg/...` green:
- round-trip over all three fields (incl. max `uint64`, no float truncation)
- `SchemaVersion == 2` lock
- `omitempty` omits empty safety fields
- snake_case wire-key assertion (rejects accidental camelCase)

PR #80, opened 2026-06-18
feat(searchmsg): shared fail-closed ExtractVisibility for producer + indexer

yujiawei:
## What

Extract the message-visibility parser into the shared `searchmsg` contract as `ExtractVisibility`, so all three sides of the search pipeline use **one** source of truth:

- octo-server `modules/searchetl` producer (real-time + backfill enrichment)
- octo-search-indexer `internal/backfill` enrichment
- any future reader-side re-check

Previously the logic lived only in octo-search-indexer (`internal/backfill/doc_enrich.go::extractVisibility`, a Go-internal package the independent indexer image could not share with the server). Re-implementing it in octo-server would fork the security semantic across repos — the root cause class behind the #1124 over-exposure.

## Fail-closed semantics (unchanged intent, now stricter + shared)

The reader treats **empty `visibles` as fail-OPEN** (no gate). So:

| payload | result |
|---|---|
| not a JSON object (corrupt/array/scalar) | **error → caller DLQs** |
| object, **no** `visibles` key | broadcast → pass (`visibles=nil`) |
| `visibles` present but `null` / non-array / `[]` / all-non-string / all-empty | **error → caller DLQs** (valid-but-empty also blocked) |
| `visibles` with ≥1 valid string UID | pass |

`space_id` is type-tolerant: only taken when a JSON string, otherwise degrades to empty **without** dropping a legitimate `visibles` (isolates the V3b fail-OPEN root cause).

This is stricter than the old indexer parser, which let a valid-but-empty `[]` through as a broadcast — exactly the gap ReviewBot YUJ-4953 flagged.

## Shared test vectors

`FailClosedVisibilityVectors()` ships as non-test source so octo-server and octo-search-indexer both import and run the **identical** vector set, locking one semantic across repos.

## Test

`go build ./...`, `go vet ./contract/...`, `go test ./contract/searchmsg/` all green.

Part of the Route A real-time unblock chain (producer enrichment depends on this).

### DIFF
--- a/contract/searchmsg/visibility.go
+++ b/contract/searchmsg/visibility.go
@@ -1,0 +1,114 @@
+package searchmsg
+
+import (
+	"bytes"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+)
+
+// ErrVisibilityFailClosed 表示一条**非加密**消息的可见性无法被可信解析，调用方必须
+// fail-closed（落 DLQ，绝不进正文 topic）。判 errors.Is(err, ErrVisibilityFailClosed)
+// 可与「真序列化/IO 错误」区分，但当前所有 fail-closed 返回都包它，调用方只需判 err!=nil。
+var ErrVisibilityFailClosed = errors.New("searchmsg: visibility fail-closed")
+
+// ExtractVisibility 从一条消息的**原始 payload 字节**解析 reader 鉴权所需的可见性字段
+// （SpaceID / Visibles）——这是 octo-im 消息检索管线 fail-closed 可见性口径的**单一真源**。
+//
+// 三方共用（强制 (a) 抽 octo-lib，禁 (b) 各仓重实现，防 #1124 口径分叉）：
+//   - octo-server searchetl producer（实时 + backfill 富化时填 searchmsg.Message.SpaceID/Visibles）
+//   - octo-search-indexer backfill（docFromRow 富化 esindex.Doc.SpaceID/Visibles）
+//   - 未来 reader 侧若需从 payload 复核可见性
+//
+// 仅对**非加密**消息调用：Signal 加密 DM 的 payload 是密文，调用方应在调用前判加密并直接
+// 走 raw_excluded（SpaceID/Visibles 留空 = reader fail-closed，安全方向），绝不把密文当损坏
+// JSON 喂进来误判。
+//
+// 🔴 fail-closed 返回值契约（reader visibility.go 对空 Visibles 是 **fail-OPEN**：普通成员
+// 能搜出「仅指定成员可见」的群定向系统消息，正是 #1124 泄漏）：
+//
+//	① payload 顶层不是 JSON 对象（损坏 / 截断 / 数组 / 标量）
+//	    → err != nil。可见性整体不可信，调用方 fail-closed（DLQ）。
+//	② payload 是对象，但**不含** visibles 键
+//	    → spaceID（容忍解析）, nil, nil。这是「广播给频道全员」的正常消息（普通群聊正文、
+//	      GroupCreate/GroupUpdate 等无定向系统消息）——reader 无 gate 即放行属预期，**安全**。
+//	③ payload 含 visibles 键，但解析不出 **≥1 个有效字符串 UID**
+//	    （visibles 为 null / 非数组 / 空数组 [] / 数组里全是非字符串或空串元素）
+//	    → err != nil。**关键**：visibles 键的「存在」即表示「本消息本应受白名单约束」（octo-lib
+//	      config/msg_group.go 仅在 GroupMemberBeRemove/Invite/Exit 等定向系统消息上写 visibles=
+//	      subscribers，且发送前已保证 subscribers 非空）。若键在却解析为空，说明白名单已损坏/漂移
+//	      → 必须 fail-closed（DLQ），绝不写空 Visibles 让 reader fail-OPEN。这正是 ReviewBot
+//	      YUJ-4953 钉死的「valid-but-empty visibles 也要拦」口径，比旧 indexer extractVisibility
+//	      （把空数组当广播放行）更严。
+//	④ payload 含合法 visibles（≥1 个有效字符串 UID）
+//	    → spaceID, visibles, nil。
+//
+// space_id 的类型**容忍**：仅当 JSON 字符串时取值，其余类型（数字 / 对象 / null / 上游漂移）
+// 退化为空 spaceID（reader 对 p2p 空 spaceID 走 fail-closed，安全方向）——绝不让 space_id 的
+// 怪异类型炸掉整条 payload 的解析、连累合法 visibles 被清空（这是 V3b fail-OPEN 的历史根因）。
+func ExtractVisibility(payload []byte) (spaceID string, visibles []string, err error) {
+	// 先把 payload 解成顶层对象（容忍每个字段各自的 JSON 类型）。顶层不是对象才是真正
+	// 「可见性不可信」→ fail-closed。单个字段的类型怪异不在此判死。
+	var top map[string]json.RawMessage
+	dec := json.NewDecoder(bytes.NewReader(payload))
+	if derr := dec.Decode(&top); derr != nil {
+		return "", nil, fmt.Errorf("%w: payload not a JSON object: %v", ErrVisibilityFailClosed, derr)
+	}
+	// 顶层 `null` 会 Decode 成 nil map 而**不**报错——它不是合法的可见性对象，必须 fail-closed
+	// （否则落进「无 visibles 键」分支被当广播放行 = fail-OPEN）。
+	if top == nil {
+		return "", nil, fmt.Errorf("%w: payload is JSON null, not an object", ErrVisibilityFailClosed)
+	}
+	// 拒绝尾部多余字节（如 `{...} garbage`、`{...}]`、`{...}}`）：合法 payload 必须**恰好**是
+	// 一个 JSON 对象，其后只能是空白 + EOF。不能用 dec.More()——它把 `]`/`}` 当「无下一元素」
+	// 返回 false，会放过 `{...}]` 这类损坏 payload（落进「无 visibles 键」分支被当广播 = fail-OPEN）。
+	// 改为再 Decode 一次并要求返回 io.EOF：任何可再解出的值或非 EOF 错误都说明有尾部残留 → fail-closed。
+	var trailing json.RawMessage
+	if derr := dec.Decode(&trailing); !errors.Is(derr, io.EOF) {
+		return "", nil, fmt.Errorf("%w: payload has trailing data after JSON object", ErrVisibilityFailClosed)
+	}
+
+	// space_id：容忍类型——仅 JSON 字符串取值，其余类型留空（不报错、不连累 visibles）。
+	if raw, ok := top["space_id"]; ok {
+		var s string
+		if uerr := json.Unmarshal(raw, &s); uerr == nil {
+			spaceID = s
+		}
+		// 非字符串 space_id（数字 / 对象 / null）→ spaceID 留空（reader p2p fail-closed）。
+	}
+
+	// visibles：键**不存在** = 广播消息（无 gate 意图），放行进正文流（安全：本就发给全员）。
+	rawVis, ok := top["visibles"]
+	if !ok {
+		return spaceID, nil, nil
+	}
+
+	// 键存在即表示「本应受白名单约束」。从这里起，任何「解析不出 ≥1 个有效 UID」都 fail-closed。
+	// null：键在却无值 = 白名单损坏/漂移 → fail-closed（绝不当广播放行）。
+	if string(bytes.TrimSpace(rawVis)) == "null" {
+		return "", nil, fmt.Errorf("%w: visibles present but null", ErrVisibilityFailClosed)
+	}
+	var elems []json.RawMessage
+	if uerr := json.Unmarshal(rawVis, &elems); uerr != nil {
+		// 非数组（对象 / 标量 / 字符串）= 白名单结构损坏 → fail-closed。
+		return "", nil, fmt.Errorf("%w: visibles is not a JSON array: %v", ErrVisibilityFailClosed, uerr)
+	}
+	out := make([]string, 0, len(elems))
+	for _, raw := range elems {
+		var s string
+		if uerr := json.Unmarshal(raw, &s); uerr != nil {
+			// 与 octo-server visiblesAllows 口径一致：非字符串元素跳过（不当作有效约束）。
+			continue
+		}
+		if s != "" {
+			out = append(out, s)
+		}
+	}
+	if len(out) == 0 {
+		// visibles 键在、却解析不出任何有效 UID（空数组 [] / 全非字符串 / 全空串）
+		// = valid-but-empty 白名单损坏 → fail-closed（ReviewBot YUJ-4953 钉死口径）。
+		return "", nil, fmt.Errorf("%w: visibles present but resolves to zero valid UIDs", ErrVisibilityFailClosed)
+	}
+	return spaceID, out, nil
+}
--- a/contract/searchmsg/visibility_test.go
+++ b/contract/searchmsg/visibility_test.go
@@ -1,0 +1,62 @@
+package searchmsg
+
+import (
+	"errors"
+	"reflect"
+	"testing"
+)
+
+// TestExtractVisibility_SharedVectors 跑共享 fail-closed 向量集，锁口径单一真源。
+// octo-server searchetl 与 octo-search-indexer backfill 各自 import 同一组向量跑同样断言，
+// 防 #1124 在不同仓重新分叉（验收门 (ii)）。
+func TestExtractVisibility_SharedVectors(t *testing.T) {
+	for _, v := range FailClosedVisibilityVectors() {
+		v := v
+		t.Run(v.Name, func(t *testing.T) {
+			spaceID, visibles, err := ExtractVisibility(v.Payload)
+			if v.WantErr {
+				if err == nil {
+					t.Fatalf("%s: want fail-closed err, got spaceID=%q visibles=%v", v.Name, spaceID, visibles)
+				}
+				if !errors.Is(err, ErrVisibilityFailClosed) {
+					t.Fatalf("%s: err must wrap ErrVisibilityFailClosed, got %v", v.Name, err)
+				}
+				// fail-closed 时绝不返回任何可见性值（防调用方误用）。
+				if spaceID != "" || visibles != nil {
+					t.Fatalf("%s: fail-closed must return empty values, got spaceID=%q visibles=%v", v.Name, spaceID, visibles)
+				}
+				return
+			}
+			if err != nil {
+				t.Fatalf("%s: want ok, got err %v", v.Name, err)
+			}
+			if spaceID != v.WantSpaceID {
+				t.Fatalf("%s: spaceID=%q want %q", v.Name, spaceID, v.WantSpaceID)
+			}
+			if !reflect.DeepEqual(visibles, v.WantVisibles) {
+				t.Fatalf("%s: visibles=%v want %v", v.Name, visibles, v.WantVisibles)
+			}
+		})
+	}
+}
+
+// TestExtractVisibility_EmptyVisiblesFailClosed 单独钉死 ReviewBot 最强调的口径：
+// 非加密群消息 visibles **valid-but-empty**（键在、空数组）必须 fail-closed，绝不放空 visibles。
+func TestExtractVisibility_EmptyVisiblesFailClosed(t *testing.T) {
+	_, _, err := ExtractVisibility([]byte(`{"content":"x","visibles":[]}`))
+	if err == nil {
+		t.Fatal("valid-but-empty visibles MUST fail-closed (else reader fail-OPEN, #1124)")
+	}
+}
+
+// TestExtractVisibility_NoVisiblesIsBroadcast 钉死：无 visibles 键 = 广播，放行不报错。
+// 否则全部正常群聊正文会被灌进 DLQ。
+func TestExtractVisibility_NoVisiblesIsBroadcast(t *testing.T) {
+	spaceID, visibles, err := ExtractVisibility([]byte(`{"type":1,"content":"normal chat"}`))
+	if err != nil {
+		t.Fatalf("broadcast (no visibles key) must pass, got %v", err)
+	}
+	if spaceID != "" || visibles != nil {
+		t.Fatalf("broadcast must yield empty spaceID + nil visibles, got %q %v", spaceID, visibles)
+	}
+}
--- a/contract/searchmsg/visibility_vectors.go
+++ b/contract/searchmsg/visibility_vectors.go
@@ -1,0 +1,150 @@
+package searchmsg
+
+// 本文件是**非测试**源码（不带 _test.go 后缀），故可被任意外部仓 import。它把
+// ExtractVisibility 的 fail-closed 测试向量收敛为单一真源，供三方共用，锁口径：
+//   - octo-server  modules/searchetl 的 producer fail-closed 单测
+//   - octo-search-indexer backfill 富化路径单测
+// 两侧跑**同一组**向量（验收门 (ii)：同口径锁），防 #1124 在不同仓重新分叉。
+
+// VisibilityVector 是一条 ExtractVisibility 的期望行为向量。
+type VisibilityVector struct {
+	// Name 用例名（断言失败时定位）。
+	Name string
+	// Payload 原始消息 payload 字节（非加密）。
+	Payload []byte
+	// WantErr 期望 fail-closed（true → 调用方必须落 DLQ，绝不进正文 topic）。
+	WantErr bool
+	// WantSpaceID / WantVisibles 仅在 WantErr=false 时校验。WantVisibles=nil 表示
+	// 「广播消息，无白名单」（reader 无 gate 放行属预期安全行为）。
+	WantSpaceID  string
+	WantVisibles []string
+}
+
+// FailClosedVisibilityVectors 返回锁口径的共享测试向量集。
+//
+// 安全核心（ReviewBot YUJ-4953 裁决逐字钉死，不可弱化）：
+//   - 非加密群消息 visibles **解析失败** → fail-closed（DLQ）。
+//   - 非加密群消息 visibles **valid-but-empty**（键在但空数组 / null / 全非字符串）
+//     → 同样 fail-closed（DLQ）。population 校验主落点在此，空数组也要拦。
+//   - visibles 键**缺失** = 广播消息 → 放行（visibles=nil），这是正常群聊正文/无定向系统
+//     消息，reader 无 gate 放行属预期；若把它也拦死会把全部正常消息灌进 DLQ。
+//   - space_id 非字符串类型不连累 visibles（V3b fail-OPEN 根因隔离）。
+func FailClosedVisibilityVectors() []VisibilityVector {
+	return []VisibilityVector{
+		// ---- 放行类（visibles 键缺失 = 广播；或含合法白名单）----
+		{
+			Name:         "broadcast/no_visibles_key",
+			Payload:      []byte(`{"type":1,"content":"hello group"}`),
+			WantErr:      false,
+			WantVisibles: nil,
+		},
+		{
+			Name:         "broadcast/no_visibles_with_space_id",
+			Payload:      []byte(`{"type":1,"content":"hi","space_id":"space_42"}`),
+			WantErr:      false,
+			WantSpaceID:  "space_42",
+			WantVisibles: nil,
+		},
+		{
+			Name:         "valid/visibles_single_uid",
+			Payload:      []byte(`{"type":99,"content":"you were removed","visibles":["u_alice"]}`),
+			WantErr:      false,
+			WantVisibles: []string{"u_alice"},
+		},
+		{
+			Name:         "valid/visibles_multi_uid_with_space",
+			Payload:      []byte(`{"content":"x","visibles":["u_a","u_b"],"space_id":"s1"}`),
+			WantErr:      false,
+			WantSpaceID:  "s1",
+			WantVisibles: []string{"u_a", "u_b"},
+		},
+		{
+			// space_id 非字符串（数字）必须退化为空 spaceID，且**绝不**连累合法 visibles。
+			Name:         "tolerant/space_id_number_does_not_drop_visibles",
+			Payload:      []byte(`{"content":"x","space_id":123,"visibles":["u_a","u_b"]}`),
+			WantErr:      false,
+			WantSpaceID:  "",
+			WantVisibles: []string{"u_a", "u_b"},
+		},
+		{
+			// 数组里混入非字符串元素：跳过非法元素，仍保留有效 UID（与 visiblesAllows 一致）。
+			Name:         "valid/visibles_mixed_skips_non_string",
+			Payload:      []byte(`{"content":"x","visibles":["u_a",123,null,"u_b"]}`),
+			WantErr:      false,
+			WantVisibles: []string{"u_a", "u_b"},
+		},
+
+		// ---- fail-closed 类：解析失败（unparseable）----
+		{
+			Name:    "failclosed/payload_not_json",
+			Payload: []byte(`{not valid json`),
+			WantErr: true,
+		},
+		{
+			Name:    "failclosed/payload_top_level_array",
+			Payload: []byte(`["a","b"]`),
+			WantErr: true,
+		},
+		{
+			Name:    "failclosed/payload_top_level_scalar",
+			Payload: []byte(`"just a string"`),
+			WantErr: true,
+		},
+		{
+			// 顶层 JSON null：Decode 成 nil map 不报错，但不是合法可见性对象 → fail-closed。
+			Name:    "failclosed/payload_json_null",
+			Payload: []byte(`null`),
+			WantErr: true,
+		},
+		{
+			// 尾部多余字节：Decode 只吃第一个值，残留 = payload 不是恰好一个对象 → fail-closed。
+			Name:    "failclosed/payload_trailing_data",
+			Payload: []byte(`{"content":"x"} garbage`),
+			WantErr: true,
+		},
+		{
+			// 尾部多余 `]`：dec.More() 会误判为「无更多」，必须靠二次 Decode 要求 EOF 才能拦住。
+			Name:    "failclosed/payload_trailing_bracket",
+			Payload: []byte(`{"content":"x"}]`),
+			WantErr: true,
+		},
+		{
+			// 尾部多余 `}`：同上，损坏 payload 不可信 → fail-closed。
+			Name:    "failclosed/payload_trailing_brace",
+			Payload: []byte(`{"content":"x"}}`),
+			WantErr: true,
+		},
+		{
+			Name:    "failclosed/visibles_object_not_array",
+			Payload: []byte(`{"content":"x","visibles":{"u_a":true}}`),
+			WantErr: true,
+		},
+		{
+			Name:    "failclosed/visibles_string_not_array",
+			Payload: []byte(`{"content":"x","visibles":"u_a"}`),
+			WantErr: true,
+		},
+
+		// ---- fail-closed 类：valid-but-empty（ReviewBot 特别强调，空数组也要拦）----
+		{
+			Name:    "failclosed/visibles_empty_array",
+			Payload: []byte(`{"content":"x","visibles":[]}`),
+			WantErr: true,
+		},
+		{
+			Name:    "failclosed/visibles_null",
+			Payload: []byte(`{"content":"x","visibles":null}`),
+			WantErr: true,
+		},
+		{
+			Name:    "failclosed/visibles_all_non_string",
+			Payload: []byte(`{"content":"x","visibles":[123,456,null]}`),
+			WantErr: true,
+		},
+		{
+			Name:    "failclosed/visibles_all_empty_string",
+			Payload: []byte(`{"content":"x","visibles":["",""]}`),
+			WantErr: true,
+		},
+	}
+}


lml2468 (dismissed):
✅ **APPROVE with should-fix findings** — `feat(searchmsg)` shared fail-closed ExtractVisibility

Verified locally on head `11fdbfe` (go 1.26.4): `go test ./contract/searchmsg/` green (all 14 shared vectors + 2 unit tests), `go build ./...` ok, `go vet` clean, `gofmt -l` clean.

The core design is exactly right: one shared parser as the single source of truth, shipped with `FailClosedVisibilityVectors()` as **non-test source** so octo-server and octo-search-indexer run the identical vector set — this is the correct way to stop the #1124 semantic from forking across repos. The documented truth table for normal object payloads is correct and the stricter "valid-but-empty `visibles` → fail-closed" (YUJ-4953) is properly enforced and tested. `space_id` type-tolerance correctly isolates the V3b root cause without dropping legit `visibles`.

But I probed edge cases the vectors don't cover, and found **two payloads that violate the parser's own row ① ("top-level not a JSON object → fail-closed")**. Both come from using `json.NewDecoder(...).Decode()`:

**🟠 Should-fix #1 — JSON `null` payload bypasses fail-closed (spec violation).**
`ExtractVisibility([]byte("null"))` returns `("", nil, nil)` — i.e. treated as a **broadcast pass**, not an error. `null` is a JSON scalar; row ① of your own doc says scalars must fail-closed. Root cause: `json.Unmarshal("null", &map)` (and `Decoder.Decode`) leaves the map nil with **no error**, so both key lookups miss and it falls through to the broadcast return. A message whose payload is `null` would be indexed with no visibility gate. Low real-world likelihood, but this is a fail-closed parser whose entire value is not relying on "unlikely."
Fix: after decoding, `if top == nil { return "", nil, fmt.Errorf("%w: payload is JSON null", ErrVisibilityFailClosed) }`.

**🟡 Should-fix #2 — trailing garbage after the first JSON value is silently accepted.**
`{"visibles":["u_a"]} <anything>` parses clean. `json.NewDecoder().Decode()` reads only the first value and ignores the rest (`dec.More()` would be true). `json.Unmarshal` rejects this with `invalid character ... after top-level value`. A truncated/corrupt/concatenated payload is exactly the "untrusted, can't fully parse → DLQ" case row ① is meant to catch. Not attacker-controlled in the normal producer flow (payload comes from the message table), so this is more robustness than active exposure — but it's still a documented-contract gap.
Fix: switch to `json.Unmarshal` and add the nil-map check above; or after `Decode`, reject if `dec.More()`.

Both fixes are ~2 lines and I'd add a vector for each (`failclosed/payload_json_null`, `failclosed/trailing_garbage`) so the shared set locks them across repos too.

**Why approve and not request-changes:** the core semantics for real object payloads are correct, well-tested, and the stricter empty-visibles rule is the actual point of this PR. This code is additive and not yet wired into producers (the PR notes producer enrichment depends on it landing). So it's safe to merge the foundation — but I'd **strongly recommend tightening both before the producer actually starts relying on `ExtractVisibility`**, since this is the very parser meant to close the fail-OPEN gap from #78/#1124. If the team prefers them fixed in-PR, that's a reasonable call too — they're trivial.

**CI**: `dependency-review` ❌ is org infra (Dependency graph not enabled), same as #74/#76/#78 — not code. `check-sprint` pending; keep the `Closes #` / issue linkage.

Verdict: **approve**, with two should-fix edge cases in the fail-closed path (JSON `null` bypass is the one I'd not let ship to a live producer).

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# octo-lib PR#80 Review Report

**PR:** https://github.com/Mininglamp-OSS/octo-lib/pull/80
**Head SHA:** `11fdbfe6610e9b402bcc6164a986cdfaa2f1c1cd`
**Scope:** 3 new files, +288/-0 lines (`contract/searchmsg/visibility.go`, `visibility_test.go`, `visibility_vectors.go`)

---

## 1. Verification

| Item | Status | Evidence |
|---|---|---|
| `go build ./...` | ✅ | Clean build, exit 0 |
| `go vet ./contract/...` | ✅ | No issues |
| `go test ./contract/searchmsg/ -v` | ✅ | All 22 subtests + 2 standalone tests PASS |
| Fail-closed semantics match doc contract | ✅ (with one P2 gap, see §2) | 15 vector cases + 2 standalone tests cover all documented paths |
| Shared vectors file is non-test source (importable) | ✅ | `visibility_vectors.go` (not `_test.go`), exports `FailClosedVisibilityVectors()` |
| Error wrapping (`ErrVisibilityFailClosed`) | ✅ | `errors.Is()` works in test assertions (visibility_test.go:23) |

## 2. Findings

### F1 — P2: `null` payload bypasses fail-closed (documented contract violation)

**Diff-scope:** new (function is new in this PR; gap is in new code, not pre-existing)

**What:** `json.NewDecoder(bytes.NewReader([]byte("null"))).Decode(&top)` where `top` is `map[string]json.RawMessage` **succeeds** with `top=nil, err=nil` (verified empirically). Nil-map key lookup returns `ok=false` for both `"space_id"` and `"visibles"`, so the function returns `"", nil, nil` — treating `null` as a broadcast message.

**Doc contract says** (visibility.go:28): *"payload 顶层不是 JSON 对象（损坏 / 截断 / 数组 / 标量）→ err != nil"*. JSON `null` is not an object — this should fail-closed.

**Severity rationale (P2, not P1):**
- Not practically reachable: octo-im payloads are always constructed via `util.ToJson(map[string]interface{}{...})` (config/msg_group.go), producing JSON objects. A `null` payload would require storage corruption or a bug in an upstream component.
- Even if reached: the message has no content to index; the only risk is a phantom "broadcast" entry in the search index with no searchable content.
- Does not satisfy R1 ("让本来能工作的路径在生产中变得不可用 / 产生用户可见的错误数据 / 比修复前更糟").

**Fix (trivial, one line):** After `dec.Decode(&top)`, add `if top == nil { return "", nil, fmt.Errorf("%w: payload is JSON null", ErrVisibilityFailClosed) }`.

**Missing test vector:** Add `{"failclosed/payload_null", []byte("null"), true}` to `FailClosedVisibilityVectors()`.

### No P0/P1 findings.

## 3. Suggestions

1. **Add `null` payload guard** (F1 fix above). One-line defensive check closes the contract gap and locks it with a shared vector.
2. Consider adding an empty-payload vector (`[]byte("")`) to the shared vectors — currently covered implicitly by `json.Decoder` returning `io.EOF` wrapped in fail-closed, but an explicit vector prevents regression if the decode strategy changes.

## 4. Additional Observations

- **Upstream data-flow verified:** `visibles` in payloads is written by `SendGroupMemberBeRemove` (config/msg_group.go:228, guarded by `len(req.Members) <= 0` early return at :204), `SendGroupMemberInviteReq` (:343, `req.Subscribers`), and `SendGroupExit` (:375, `visibleUids`). All produce JSON arrays of non-empty UID strings when functioning correctly. The parser's fail-closed on empty/malformed arrays is the correct defensive posture.
- **`space_id` type tolerance is well-designed:** Non-string `space_id` degrades to empty without dropping valid `visibles` — this isolates the V3b fail-OPEN root cause as documented.
- **Test design is strong:** The shared vector approach (`visibility_vectors.go` as non-test source) ensures octo-server and octo-search-indexer run identical assertions, preventing the #1124 semantic fork. The standalone `EmptyVisiblesFailClosed` and `NoVisiblesIsBroadcast` tests pin the two most critical invariants independently.

## 5. Data-Flow Trace

| Consumed data | Upstream source | Reaches consumer? |
|---|---|---|
| `payload []byte` | octo-server searchetl producer / octo-search-indexer backfill (reads `message.payload` column) | Yes — raw bytes from DB |
| `top["space_id"]` | Payload JSON field, written by message sender (octo-lib config or client SDK) | Yes — parsed from payload |
| `top["visibles"]` | Payload JSON field, written by `SendGroupMemberBeRemove`/`SendGroupMemberInviteReq`/`SendGroupExit` | Yes — parsed from payload |
| `ErrVisibilityFailClosed` | Sentinel error, `errors.New` at package init | Yes — wrapped via `%w`, `errors.Is()` works |
| Return `spaceID` | `json.Unmarshal(raw, &s)` on `top["space_id"]` | Yes — only when string type |
| Return `visibles` | Loop over `[]json.RawMessage` elements, string-only filter | Yes — skips non-string/empty elements |

## 6. Blind-Spot Checklist (R5 — security_sensitive)

- **C1 (dual-path parity):** N/A — this is a new standalone parser function. No symmetric add/remove paths. The function will be called by two consumers (producer + indexer), and the shared vector set ensures parity.
- **C2 (control-flow ordering / nesting):** Clear — single-pass linear parse (decode → space_id → visibles). No nested calls, no re-entrancy, no ordering hazards.
- **C3 (authorization boundary ≠ capability boundary):** N/A — this is a data parser, not an authorization endpoint. It extracts visibility metadata; the reader-side `visibility.go` enforces authorization.
- **C4 (authorization lifecycle / container-member cascade):** N/A — parser does not perform authorization checks. It extracts fields for downstream authorization consumers.

## 7. Cross-Round Blocker Recheck (R6)

---

**[Octo-Q] verdict: APPROVE**

Well-designed fail-closed visibility parser with strong test coverage and clear security rationale. One P2 finding (null payload contract gap) is not blocking — trivially fixable, not practically reachable. The shared vector approach effectively prevents the #1124 semantic fork risk.

Jerry-Xin (changes requested):
Summary: The PR is relevant to `octo-lib`, but the new fail-closed parser currently fail-opens some malformed top-level payloads.

🔴 Blocking

🔴 Critical — `ExtractVisibility` accepts non-object/corrupt payloads as trusted broadcasts. In `contract/searchmsg/visibility.go:53-70`, `json.Decoder.Decode(&top)` is used without checking that the decoded map is non-nil and without checking for trailing tokens. As a result, `null` decodes successfully into a nil map and returns `spaceID="", visibles=nil, err=nil`, and payloads like `{"content":"x"} trailing` or `{"visibles":["u"]}{"x":1}` also pass after the first object. This violates the stated contract that non-object/corrupt payloads must fail-closed, and the no-`visibles` case can become reader fail-open.

Add coverage to `FailClosedVisibilityVectors()` for top-level `null` and valid-object-prefix-with-trailing-garbage, then make the parser reject them. For example, use `json.Unmarshal` plus `top != nil`, or keep the decoder and verify the next decode reaches EOF after whitespace.

✅ Highlights

The shared vector approach is a good fit for keeping producer and indexer behavior aligned, and the valid-but-empty `visibles` cases are covered well.

Verification run: `go test ./contract/searchmsg`, `go test ./...`, and `go vet ./contract/...` all passed on the checked-out PR.

yujiawei (commented):
# Code Review — PR #80 (octo-lib)

**Scope reviewed:** head `11fdbfe6610e9b402bcc6164a986cdfaa2f1c1cd` vs `main`. 3 new files, +288/-0, all in `contract/searchmsg/`. `go build ./...`, `go vet ./contract/...`, and `go test ./contract/searchmsg/` (17 cases) all green locally.

This PR extracts message-visibility ACL parsing into `searchmsg.ExtractVisibility` as the **single source of truth** shared by the producer and indexer. Because it is the authoritative fail-closed gate behind the #1124 over-exposure class, the bar is "no path returns empty `visibles` without an error for a message that should be gated." The stricter-than-old semantics (valid-but-empty `visibles` → fail-closed) are correct and well tested. However, two top-level-payload paths still fail **OPEN**, and one contradicts the function's own documented contract.

## Verdict: CHANGES_REQUESTED

One P1 fail-open in the security gate. The fix is small (≈2 lines) and is the reason to hold rather than approve-with-nits: this is the designated source of truth, so the defect propagates to every future caller that (by design) does not re-implement the parse.

---

## P1 — Top-level `null` (and trailing data) payloads fail OPEN

`contract/searchmsg/visibility.go:52-56`

```go
var top map[string]json.RawMessage
dec := json.NewDecoder(bytes.NewReader(payload))
if derr := dec.Decode(&top); derr != nil {
    return "", nil, fmt.Errorf("%w: payload not a JSON object: %v", ErrVisibilityFailClosed, derr)
}
```

Two distinct corrupt payloads slip through as ungated broadcasts:

1. **Top-level `null`.** `Decode(&top)` on the literal `null` returns `err == nil` with `top == nil`. Execution then reaches the `visibles` lookup, the key is absent, and the function returns `("", nil, nil)` — i.e. **broadcast pass, no error**. This directly contradicts the doc-comment contract on this same function, case ①: *"payload 顶层不是 JSON 对象（损坏 / 截断 / 数组 / 标量）→ err != nil"*. `null` is a JSON scalar; it is documented to fail-closed but actually fails open. The shared vectors include `payload_top_level_array` and `payload_top_level_scalar` (a bare string) but **not** `null`, so the gap is untested.

2. **Trailing data after the first value.** `json.Decoder.Decode` reads exactly one JSON value and ignores the rest of the buffer. `{}{"visibles":["admin_only"]}` and `{"visibles":["admin"]} <garbage>` both parse only the first token and return without error — the first being an empty object → broadcast pass. A truncated/concatenated envelope is silently classified as ungated.

Reproduced locally against this exact head:

```
payload="null"                         => spaceID="" visibles=[] err=<nil>   # expected: err != nil
payload="{}{\"visibles\":[\"admin_only\"]}" => spaceID="" visibles=[] err=<nil>   # expected: err != nil
payload=""                             => spaceID="" visibles=[] err=EOF     # OK (fails closed)
```

**Why P1, not defense-in-depth:** the only caller today (`docFromRow` → `extractMessage`) happens to reject `null`/trailing upstream via a strict `json.Unmarshal(row.Payload, &m)` that returns `err`/empty-map → DLQ. But the stated purpose of this PR is that callers **must not** re-implement the parse — the producer added in the follow-up ticket, and any future reader-side re-check, will call `ExtractVisibility` directly and inherit this fail-open. A source-of-truth ACL parser that returns "ungated broadcast" for a corrupt payload is the #1124 failure mode it exists to prevent.

**Suggested fix (closes both sub-cases at once):**

```go
var top map[string]json.RawMessage
if derr := json.Unmarshal(payload, &top); derr != nil {
    return "", nil, fmt.Errorf("%w: payload not a JSON object: %v", ErrVisibilityFailClosed, derr)
}
if top == nil { // literal `null` decodes without error but yields a nil map
    return "", nil, fmt.Errorf("%w: payload is JSON null", ErrVisibilityFailClosed)
}
```

`json.Unmarshal` rejects trailing data (`invalid character '{' after top-level value`), and the explicit `top == nil` guard catches the `null` literal (which `Unmarshal` also accepts into a nil map). Verified locally. Please also add `failclosed/payload_top_level_null` and `failclosed/payload_trailing_data` to `FailClosedVisibilityVectors()` so the shared vector set locks the gap across both repos.

---

## P2 — Missing-`visibles` key is an unconditional broadcast even for known whitelist-gated system types

`contract/searchmsg/visibility.go:67-70`

The missing-key branch treats *every* payload without a `visibles` key as a safe broadcast. But some non-encrypted system message types are produced specifically because they carry a whitelist — `config/msg_group.go` writes `"visibles": subscribers` for `GroupMemberBeRemove` (line 228), `GroupMemberInvite` (line 343), and `GroupMemberQuit` (line 365). If an old or drifted payload of one of those types loses its `visibles` key, this returns `("", nil, nil)` and the reader fails open to ordinary group members.

This is intentionally in tension with the package's "decouple from payload internals / don't enumerate types" design, which is why it's P2 not P1: type-awareness here couples octo-lib to message-type semantics it deliberately avoids, and the realistic occurrence (a gated type that somehow dropped its own `visibles`) is a narrow drift case the producer enrichment should not emit. Recommend either (a) a documented decision to accept missing-key-as-broadcast for all types, or (b) a caller-side type allowlist in the producer/indexer rather than in the shared parser. Worth an explicit note in the doc-comment so the boundary is a recorded decision, not an implicit one.

## P2 / nit — Shared test vectors compile into production binaries

`contract/searchmsg/visibility_vectors.go`

`FailClosedVisibilityVectors()` ships as non-test source on purpose (so both repos import the identical vectors — a good call for locking semantics). The trade-off is that the fixtures and their byte-literal payloads are linked into every production binary that imports `searchmsg`. Acceptable given the goal, but consider isolating them in a dedicated sub-package (e.g. `searchmsg/searchmsgtest`) so the contract package proper stays fixture-free. Not blocking.

---

## Coverage notes (what I did and did not verify)

- **Verified:** build/vet/test green at head; cross-checked the new strict semantics against the existing `octo-search-indexer` `internal/backfill/doc_enrich.go::extractVisibility` (new version is correctly stricter on empty/null — matches the intended convergence); confirmed `space_id` type-tolerance does not drop a legitimate `visibles` (the V3b root cause is correctly isolated); confirmed the gated system-message types in `config/msg_group.go` that motivate the P2.
- **Not verified (out of this diff):** the follow-up producer that will import `ExtractVisibility` (not in this repo yet) — the P1 matters precisely because that caller is new code; encrypted-DM exclusion happens correctly *before* this function is called (verified in the indexer caller, but each future caller must uphold it — `ExtractVisibility` itself has no encryption guard and trusts the caller, which is documented but caller-dependent).

lml2468 (dismissed):
✅ **APPROVE (re-review)** — head `3c15eb2` (was `11fdbfe`). My 🔴 findings are fixed.

Both fail-closed gaps I flagged are correctly closed, with the exact fix the reviewers converged on:
- 🟢 **JSON `null` → fail-closed.** `if top == nil` after `Decode` now returns `ErrVisibilityFailClosed`. Re-ran my probe: `ExtractVisibility([]byte("null"))` now errors instead of passing as a broadcast.
- 🟢 **Trailing data → fail-closed.** `if dec.More()` rejects `{...} garbage` and concatenated objects `{...}{...}`.
- 🟢 Two locking vectors added (`failclosed/payload_json_null`, `failclosed/payload_trailing_data`) so both repos enforce these via the shared set.

Re-verified on `3c15eb2` (go 1.26.4):
- `go test ./contract/searchmsg/` green
- `go build ./...` ok · `go vet ./contract/...` clean · `gofmt -l` clean
- **Re-ran my full edge-case probe — zero mismatches.** Every fail-closed case (`null`, `""`, whitespace-only, top-level scalar/array, trailing garbage, concatenated objects) now errors; every pass case (real object, `{}`, broadcast, valid visibles) still passes.
- 🟢 **No regression on the case I was most worried about:** a payload with *trailing whitespace only* (`{"visibles":["u_a"]}   `) still passes — `dec.More()` correctly returns false for trailing whitespace, so the stricter check didn't break legitimate pretty-printed/whitespace-padded payloads.

The fix kept the `Decoder` approach and added the two guards rather than switching to `json.Unmarshal` — equally correct, and `dec.More()` is actually the more precise trailing-data check (whitespace-tolerant), so I'm happy with this choice over the Unmarshal alternative.

Verdict: **approve**. The blocking fail-closed gaps are closed and locked with vectors. Good, fast turnaround.

Note: head moved `11fdbfe` → `3c15eb2`, so prior reviews (incl. the CHANGES_REQUESTED) are against the old head — everyone who blocked should re-review on `3c15eb2` to clear the gate. Still not wired into any producer (zero call sites), so this lands as a correct, now-hardened library primitive; the producer-enrichment PR remains where the deploy-order gate must be enforced.

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-lib PR#80 Review Report
**PR:** Mininglamp-OSS/octo-lib #80
**Head SHA:** `3c15eb247e4b99627a2af86dc97a8d938a1fe5ce`
**Title:** feat(searchmsg): shared fail-closed ExtractVisibility for producer + indexer
**Routing:** complexity=security_sensitive (automated review)

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| `go build ./...` | ✅ PASS | exit 0, no output |
| `go vet ./contract/...` | ✅ PASS | exit 0, no output |
| `go test -v ./contract/searchmsg/` | ✅ PASS | 25/25 subtests pass (17 shared vectors + 3 standalone + 5 pre-existing) |
| Fail-closed semantic correctness | ✅ PASS | All 4 documented cases (①–④) implemented correctly |
| Data-flow: upstream writer alignment | ✅ PASS | See §5 below |

## 2. Findings

### P2-1 — Doc comment overstates upstream guarantee (pre-existing, not amplified)

**Location:** `visibility.go:38-39`

The comment states:
> "octo-lib config/msg_group.go 仅在 GroupMemberBeRemove/Invite/Exit 等定向系统消息上写 visibles=subscribers，**且发送前已保证 subscribers 非空**"

**Diff-scope check:** new (this PR's doc comment). Not amplified — the code itself is correct regardless.

**Evidence:** Tracing the three write-sites in `config/msg_group.go`:
- `SendGroupMemberBeRemove` (line 203-240): Has `len(req.Members) <= 0` early-return guard at line 204 → subscribers **guaranteed non-empty** ✅
- `SendGroupMemberInviteReq` (line 322-345): Writes `req.Subscribers` directly — **no non-empty guard** ⚠️
- `SendGroupExit` (line 355-372): Writes `visibleUids` parameter directly — **no non-empty guard** ⚠️

**Impact:** Documentation inaccuracy only. The `ExtractVisibility` code correctly handles empty/nil visibles by fail-closing (DLQ), so the security property holds regardless. If Invite/Exit callers ever pass empty slices, those messages DLQ (availability hit, not security leak). The comment should say "SendGroupMemberBeRemove guarantees non-empty; Invite/Exit callers should too" rather than blanket "已保证".

**Severity:** P2 (doc nit, no runtime impact)

### P2-2 — Missing test vector: `space_id` absent + valid `visibles`

**Location:** `visibility_vectors.go`

**Diff-scope check:** new (this PR's test suite).

The vector set covers `space_id` present-as-string + visibles, and `space_id` present-as-number + visibles, but lacks an explicit vector with **no `space_id` key at all** + valid `visibles`. The `valid/visibles_single_uid` vector implicitly covers this (no `space_id` key, WantSpaceID defaults to `""`), so this is a readability/completeness nit, not a gap.

**Severity:** P2 (test readability nit)

## 3. Suggestions

1. **Fix doc comment** at `visibility.go:38-39`: Qualify the "已保证 subscribers 非空" claim to specify it applies to `SendGroupMemberBeRemove` only, and note that Invite/Exit callers are expected (but not enforced) to pass non-empty lists.

2. **Add explicit vector** `valid/visibles_no_space_id` with `WantSpaceID: ""` for clarity, even though `valid/visibles_single_uid` implicitly covers this.

## 4. Additional Observations

- **Design quality:** The fail-closed semantic is well-implemented. The four-case contract (① not-object → err, ② no visibles key → broadcast, ③ visibles present but empty/invalid → err, ④ valid → pass) is correctly coded with no edge-case leaks.
- **`json.Decoder` + `dec.More()` trailing-data check** is a robust pattern — catches `{...} garbage` payloads that `json.Unmarshal` would silently ignore.
- **Top-level `null` guard** (`visibility.go:62-64`) correctly catches the `json.Decoder` quirk where `null` decodes to nil map without error.
- **`space_id` type tolerance** correctly isolates V3b fail-OPEN root cause: a weird `space_id` never causes the entire parse to fail, preserving valid `visibles`.
- **Shared test vectors as non-test source** (`visibility_vectors.go`) is a clean pattern for locking semantic parity across repos.

## 5. Data-Flow Trace

### Upstream: Who writes `visibles` in the payload?

Three write-sites in `config/msg_group.go`:

| Function | Line | `visibles` value | Non-empty guard |
|----------|------|-------------------|-----------------|
| `SendGroupMemberBeRemove` | 228 | `subscribers` (from `req.Members[].UID`) | ✅ `len(req.Members) <= 0` early-return at line 204 |
| `SendGroupMemberInviteReq` | 343 | `req.Subscribers` | ⚠️ None |
| `SendGroupExit` | 365 | `visibleUids` (parameter) | ⚠️ None |

### Downstream: Who consumes `ExtractVisibility` output?

Per PR description, two consumers (both in separate repos, not yet wired):
- **octo-server `modules/searchetl`** producer: fills `searchmsg.Message.SpaceID` / `Visibles` for Kafka
- **octo-search-indexer `internal/backfill`**: fills `esindex.Doc.SpaceID` / `Visibles` for ES

**Consumer contract:** On `err != nil` → caller DLQs the message (never indexes with empty/missing visibility). On `err == nil` → `visibles == nil` means broadcast (reader passes), `visibles != nil` means whitelist-gated.

**Security property verification:** The reader treats empty `Visibles` as fail-OPEN (no gate). `ExtractVisibility` ensures this state is **unreachable**: it either returns `visibles != nil` (valid whitelist) or `err != nil` (DLQ). The only `visibles == nil, err == nil` path is "no visibles key in payload" = genuine broadcast. This closes #1124. ✅

### Error wrapping verification

All error returns use `fmt.Errorf("%w: ...", ErrVisibilityFailClosed, ...)`. The test asserts `errors.Is(err, ErrVisibilityFailClosed)` — correctly passes for all 11 fail-closed vectors. ✅

## 6. R5 Blind-Spot Checklist (security_sensitive)

- **C1 — Dual-path parity:** N/A. This is a pure extraction function with no add/remove or create/delete symmetry. No dual paths to check.
- **C2 — Control-flow ordering / nested reuse:** N/A. Single-pass parser, no nested calls, no security controls (regex/escape/sanitize) to trial with non-canonical input. The JSON decoder handles all encoding edge cases.
- **C3 — Authorization boundary ≠ capability boundary:** N/A. This is a library function that extracts data; it does not enforce authorization. The authorization boundary is at the reader (visibility.go in octo-server/octo-search-indexer), which is outside this PR's scope.
- **C4 — Authorization lifecycle / container-member state cascade:** N/A. This function parses message payloads, not authorization state. No container/member status checks are relevant here.

## 7. Cross-Round Blocker Recheck (R6)

---

**[Octo-Q] verdict: APPROVE**

Rationale: No P0/P1 findings. The fail-closed security semantic is correctly implemented with thorough test coverage. Two P2 nits (doc comment overstatement, missing explicit test vector) are non-blocking. The function correctly ensures that empty `Visibles` can never reach the reader's fail-OPEN path, closing the #1124 over-exposure class.

lml2468 (approved):
✅ **APPROVE (re-review)** — head `feffcc1` (was `3c15eb2`). And a correction I need to own.

**I was wrong last round.** On `3c15eb2` I approved the trailing-data guard and specifically praised `dec.More()` as "the more precise, whitespace-tolerant check." That was a mistake: `dec.More()` returns **false** for a closing `]` or `}`, so the `3c15eb2` version still let `{"content":"x"}]` and `{"content":"x"}}` through as fail-OPEN broadcasts. I just proved it empirically against the old code:

```
"{\"content\":\"x\"}]"  err=<nil>  *** FAIL-OPEN (hole) ***
"{\"content\":\"x\"}}"  err=<nil>  *** FAIL-OPEN (hole) ***
```

So the previous "fix" closed `} garbage` but not `}]` / `}}`. Good catch by whoever pushed `feffcc1` — this is exactly the kind of residual hole that hides behind a green test suite.

**The new commit closes it correctly.** It replaces `dec.More()` with a second `dec.Decode(&trailing)` that must return `io.EOF` — any further decodable value OR any non-EOF error means trailing residue → fail-closed. This is the right primitive: it's strict about "exactly one JSON object then EOF" while still tolerating trailing whitespace/newlines (decoder skips whitespace, then hits EOF).

Re-verified on `feffcc1` (go 1.26.4):
- `go test ./contract/searchmsg/` green · `go build ./...` ok · `go vet` clean · `gofmt -l` clean
- **Full edge probe — zero mismatches**, now including the newly-caught cases:
  - fail-closed: `null`, `""`, whitespace-only, top-level scalar/array, `} garbage`, `}]`, `}}`, concatenated objects `{...}{...}`
  - still pass (no regression): real object, `{}`, broadcast, valid visibles, **trailing-whitespace** `{...}   `, and **pretty-printed multiline** `{\n  "visibles": [...]\n}\n`
- Two new locking vectors (`payload_trailing_bracket`, `payload_trailing_brace`) added to the shared set so both repos enforce these.

Verdict: **approve**. The fail-closed path is now genuinely airtight for every malformed top-level shape I can construct, and the whitespace/pretty-print pass-cases are preserved.

mochashanyao's two nits on this head are fair and non-blocking: (1) the doc comment's "upstream guarantees subscribers non-empty" overstates what this function can assume — worth softening to "callers must…"; (2) add an explicit vector for *no `space_id` + valid `visibles`* (currently only covered transitively). Neither blocks merge.

Process note: head is now `feffcc1`; everyone who reviewed `3c15eb2`/`11fdbfe` (incl. any prior CHANGES_REQUESTED) needs to re-review on `feffcc1` to clear the gate. Still zero producer call sites — lands as a hardened library primitive.

Jerry-Xin (approved):
The PR is in scope for `octo-lib`: it extends the existing shared `contract/searchmsg` package used as the search-message contract.

💬 Non-blocking
- 🟡 Warning: `ExtractVisibility` treats whitespace-only strings as valid visible UIDs because it only filters `s != ""` at `contract/searchmsg/visibility.go:99-105`. If UID syntax never permits whitespace, consider `strings.TrimSpace(s) != ""` and add a vector for `["   "]`. This is not an exposure risk if reader checks exact UID equality, but it can let malformed visibility data bypass DLQ and become an unintentionally invisible document.
- 🔵 Suggestion: `FailClosedVisibilityVectors` is intentionally exported from non-test source at `contract/searchmsg/visibility_vectors.go:32`, but that makes test fixtures part of the package API. Since external repos need it, this is reasonable; consider documenting that it is a test-support contract so downstream users do not treat it as runtime data.

✅ Highlights
- `ExtractVisibility` correctly distinguishes missing `visibles` as broadcast while fail-closing malformed, null, empty, or all-invalid `visibles` at `contract/searchmsg/visibility.go:81-112`.
- The parser rejects top-level non-objects, JSON `null`, and trailing garbage before allowing broadcast behavior at `contract/searchmsg/visibility.go:53-70`.
- Tests cover the important fail-closed vectors, including valid-but-empty `visibles`, trailing data, and tolerant `space_id` handling at `contract/searchmsg/visibility_test.go:12-62`.

Verification run locally:
- `go test ./contract/searchmsg`
- `go vet ./contract/...`
- `go test ./...`

lml2468 (approved):
✅ **APPROVE (re-review)** — current head `a57dc30` (was `feffcc1`).

Head moved again, but I checked what actually changed: the `feffcc1 → a57dc30` raw diff shows `pkg/log/logger.go`, which looked alarming — but that's just **rebase noise** from #79 having merged into main. The PR's *net* diff vs merge-base (`b639b8f`) is still **only the three `contract/searchmsg/` files (+326)**, and `git diff feffcc1 a57dc30 -- contract/searchmsg/` is **empty** — the security code is byte-identical to the `feffcc1` revision I already verified airtight.

Re-verified on `a57dc30` anyway (discipline: every head gets re-run):
- `visibility.go` still has the correct fix — `if top == nil` (JSON null) + second `dec.Decode` requiring `io.EOF` (the `}]`/`}}`-catching trailing check, not the broken `dec.More()`).
- `go test ./contract/searchmsg/` green · `go build ./...` ok · `go vet` clean · `gofmt -l` clean.
- Key fail-closed probe — zero mismatches: `null`, `}]`, `}}`, `} g`, concatenated `{...}{...}`, `""`, top-level array all fail-closed; `{}`, valid visibles, trailing-whitespace all still pass.

Verdict: **approve** on `a57dc30`. No code change since `feffcc1`; the fail-closed path remains airtight and rebasing onto the merged #79 didn't disturb it.

Process: this is now the head everyone should re-vote on. Any review pinned to `11fdbfe`/`3c15eb2`/`feffcc1` (including standing CHANGES_REQUESTED) is stale — re-review on `a57dc30` to clear the gate. Still zero producer call sites; lands as a hardened library primitive, with the #78 deploy-order gate to be enforced on the producer-enrichment PR.

yujiawei (commented):
# Code Review — PR #80 (octo-lib)

**Scope reviewed:** head `3c15eb247e4b99627a2af86dc97a8d938a1fe5ce` vs `main` (merge-base `f41032c`, 1 commit ahead, no sibling-PR conflicts). 3 new files, +310/-0, all in `contract/searchmsg/`. Re-review of the previously-blocked round (prior head `11fdbfe`). Verified locally (go 1.26.3): `go build ./...`, `go vet ./contract/...`, `gofmt -l`, and `go test ./contract/searchmsg/` (all 17 shared vectors + 2 unit tests) all green.

This PR extracts message-visibility ACL parsing into `searchmsg.ExtractVisibility` as the **single source of truth** shared by the producer and indexer. Because the reader treats an empty `visibles` as **fail-OPEN**, the bar is: *no path may return `err == nil` with empty/nil `visibles` for a payload that should be access-gated or is untrusted/corrupt.*

The prior round was blocked for two fail-OPEN cases — top-level `null` and trailing data. The `null` case is now correctly closed (`if top == nil`). The trailing-data fix, however, is **incomplete**: it uses `dec.More()`, which has a hole for two specific trailing bytes.

## Verdict: CHANGES_REQUESTED

One P1 fail-OPEN remains in the security gate. The fix is ~2 lines and is the reason to hold rather than approve: this is the designated source of truth, so the defect propagates to every future caller that (by design) does not re-implement the parse.

---

## P1 — Trailing `}` or `]` after the JSON object fails OPEN

`contract/searchmsg/visibility.go:62-66`

```go
// 拒绝尾部多余字节（如 `{...} garbage`）：Decode 只消费第一个 JSON 值，残留数据说明 payload
// 不是「恰好一个 JSON 对象」，不可信 → fail-closed。
if dec.More() {
    return "", nil, fmt.Errorf("%w: payload has trailing data after JSON object", ErrVisibilityFailClosed)
}
```

`json.Decoder.More()` reports whether there is another element **in the current array or object being parsed** — it returns `false` as soon as the next non-whitespace byte is `}` or `]` (Go stdlib `stream.go`: `return err == nil && c != ']' && c != '}'`). It is not a stream-terminator check. So a payload whose trailing junk *starts* with a brace/bracket slips through the guard.

Reproduced locally against this exact head:

```
payload="{\"type\":1,\"content\":\"x\"}}"  => spaceID="" visibles=[] err=<nil>   # expected: err != nil
payload="{\"type\":1,\"content\":\"x\"}]"  => spaceID="" visibles=[] err=<nil>   # expected: err != nil
payload="{\"type\":1,\"content\":\"x\"} garbage" => err != nil   # OK (fails closed)
payload="{\"type\":1,\"content\":\"x\"},"        => err != nil   # OK (fails closed)
```

Both `}`- and `]`-terminated cases return `("", nil, nil)` — i.e. **broadcast pass, no error** — for a truncated/concatenated/corrupt envelope that the function's own doc-comment (`visibility.go:62`) says must fail-closed. This is the same fail-OPEN class that blocked the prior round; the chosen `dec.More()` fix closes ` garbage`/`,` tails but not brace/bracket tails.

**Why the shared vectors didn't catch it:** `failclosed/payload_trailing_data` (`visibility_vectors.go:99`) uses `{"content":"x"} garbage` — a non-brace tail — which happens to be exactly the case `dec.More()` *does* catch. The brace/bracket tail is untested, so the gap is invisible to the locked vector set in both repos.

**Suggested fix (closes all trailing-data cases, preserves legitimate trailing whitespace):**

```go
var top map[string]json.RawMessage
if derr := json.Unmarshal(payload, &top); derr != nil {
    return "", nil, fmt.Errorf("%w: payload not a JSON object: %v", ErrVisibilityFailClosed, derr)
}
if top == nil { // literal `null` decodes without error but yields a nil map
    return "", nil, fmt.Errorf("%w: payload is JSON null, not an object", ErrVisibilityFailClosed)
}
```

`json.Unmarshal` rejects any trailing data with `invalid character '}' after top-level value` (verified locally for `{...}}`, `{...}]`, `{...} garbage`), while still accepting legitimate trailing whitespace (`{...}   `, `{...}\n` → `err == nil`). This also subsumes the existing `null` guard. Please add `failclosed/payload_trailing_brace` (`{...}}`) and `failclosed/payload_trailing_bracket` (`{...}]`) to `FailClosedVisibilityVectors()` so the gap is locked across both repos.

**Severity rationale:** P1, not P0 — the only caller today (`docFromRow` → `extractMessage`) rejects such payloads upstream via a strict `json.Unmarshal` → DLQ, so there is no live exposure right now. But the stated purpose of this PR is that callers **must not** re-implement the parse; the follow-up producer and any future reader-side re-check will call `ExtractVisibility` directly and inherit this fail-OPEN. A source-of-truth ACL parser that returns "ungated broadcast" for a corrupt payload is precisely the failure mode it exists to prevent. Worth fixing before any producer starts relying on it.

---

## What's correct (verified)

- **`null` fail-closed (prior P1) is fixed** — `ExtractVisibility([]byte("null"))` now errors (`if top == nil`), confirmed by re-running the probe and by the new `failclosed/payload_json_null` vector.
- **Core fail-closed semantics are sound and well-tested:** non-object payloads, non-array `visibles`, and the YUJ-4953 valid-but-empty cases (`[]`, `null`, all-non-string, all-empty-string) all fail-closed. Re-ran the full suite — green.
- **`space_id` type-tolerance** correctly degrades a non-string `space_id` to empty **without** dropping a legitimate `visibles` (isolates the V3b root cause). Verified: `{"space_id":123,"visibles":["u_a","u_b"]}` → `("", ["u_a","u_b"], nil)`.
- **Shared-vectors-as-non-test-source** is the right mechanism to stop the #1124 semantic from forking across repos. The empty-`visibles`-fail-closed invariant is independently pinned by `TestExtractVisibility_EmptyVisiblesFailClosed`.
- **Duplicate JSON keys** behave per Go stdlib (last value wins) with no fail-OPEN path.

## Non-blocking (P2 / nit)

- **`missing visibles key = unconditional broadcast` is type-blind.** If a gated system-message type (the producers writing `visibles` are `GroupMemberBeRemove`/`GroupMemberInvite`/`GroupMemberQuit` in `config/msg_group.go`) ever drops its `visibles` key, this returns a broadcast pass. This is a documented design tension (the parser intentionally avoids coupling to message-type semantics), and producer drift toward *empty* `visibles` fails closed anyway — so the residual risk is narrow. Recommend recording the boundary explicitly in the doc-comment as an accepted decision. Note: only `SendGroupMemberBeRemove` guards non-empty subscribers before writing `visibles`; `SendGroupMemberInviteReq` and `SendGroupExit` do not — harmless today because the reader fails closed on empty `visibles`, but worth a producer-side guard for defense-in-depth.
- **Whitespace-only UID strings** (e.g. `"   "`) are accepted as valid UIDs (only `== ""` is filtered). This is the *safe* direction (message visible to nobody), so nit-level only.
- **Test vectors compile into production binaries** (`visibility_vectors.go` is non-test source by design). Acceptable for the cross-repo locking goal; consider a `searchmsgtest` sub-package later. Not blocking.

## Coverage notes (what I did / did not verify)

- **Verified:** build/vet/test/fmt green at head `3c15eb2`; reproduced the P1 brace/bracket fail-OPEN and confirmed the `json.Unmarshal` fix closes it while preserving trailing-whitespace tolerance; confirmed the prior `null` P1 is closed; cross-checked the `space_id` tolerance and valid-but-empty fail-closed; confirmed the gated `visibles`-writing producers in `config/msg_group.go`; confirmed the failing CI checks (`osv-scan`/`dependency-review`/`check-sprint`) are org-infra (this PR touches no `go.mod`/`go.sum` and imports only the stdlib).
- **Not verified (out of this diff):** the follow-up producer that will import `ExtractVisibility` (not in this repo yet) — the P1 matters precisely because that caller is new code; the reader-side `visibility.go` enforcement and the encrypted-DM exclusion live in callers, which this function documents but does not itself guard.

Issue #96, opened 2026-06-26
DB/Redis instrumentation seam: follow-ups & known limitations


an9xyz:
## Context

octo-lib#97 (merged) added metrics-observer seams at the MySQL and Redis client-construction sites so a downstream consumer (octo-server `#442` `DependencyMetrics`) can record dependency-call latency without octo-lib depending on Prometheus:

- `pkg/db/instrument.go` — `DBObserver` + `SetDBObserver`; `instrumentedConnector` (reports `op="connect"`) wrapping the conn in `instrumentedConn` (times `Exec`/`QueryContext` → `op="query"`, execute-only, carries error status).
- `pkg/redis/instrument.go` — `RedisObserver` + `SetRedisObserver`; `instrumentClient` (go-redis v6 `WrapProcess`/`WrapProcessPipeline`).
- `pkg/db/mysql.go`, `pkg/redis/redis.go` — wiring.

Consumer side wired in octo-server#474 (merged/open). This issue tracks quality / observability follow-ups, ranked by impact.

## Follow-ups

- [ ] **Raw Lua-capable Redis clients aren't instrumented (medium — coverage gap).** Only clients built through `pkg/redis` (`redis.New`/`NewWithOptions`, i.e. `instrumentClient`) emit the Redis observer. But `pkg/redis.Conn` doesn't expose `Eval`/`Script`/`SetNX`/pipelined primitives, so octo-server constructs **~15 raw `rd.NewClient(...)` clients** (rate-limit, OIDC state/bind/logout/sync_lock, bot registry, user/group/space auth, health, Lua locks) that bypass instrumentation entirely — `dependency=redis` is blind to those paths (pool stats still cover them). Fix options in octo-lib: (a) export `redis.Instrument(*rd.Client)` so consumers can opt a raw client in, and/or (b) widen `pkg/redis.Conn` to expose the Lua/SetNX primitives those call sites need so they don't have to drop to a raw client. Until then this is a documented coverage limitation, not a regression.

[...]

PR #97, opened 2026-06-26
feat: add MySQL/Redis client-seam latency observers

an9xyz:
## What

Adds no-op-by-default **observer seams** at the MySQL and Redis client-construction sites so a downstream consumer (octo-server `#442` `DependencyMetrics`) can record dependency-call latency — **without octo-lib taking a Prometheus dependency**. octo-lib only exposes callbacks; octo-server injects the real implementations at startup.

This gives a clean three-way split of the request tail: **connect handshake** (`op="connect"`) vs **pool-wait** (existing pool `WaitDuration`) vs **execute** (`op="query"`), each measured at a non-overlapping seam.

## Changes

- **`pkg/db/instrument.go`** (new) — `DBObserver` + `SetDBObserver` (`atomic.Pointer`, safe no-op when unset). `instrumentedConnector` wraps `driver.Connector`: times `Connect` as `op="connect"`, and wraps the returned `driver.Conn` in `instrumentedConn`, which times `ExecContext`/`QueryContext` as `op="query"`. Because the timing sits **below** pool acquisition (the conn is already in hand), `op="query"` is pure execute latency — no pool-wait, no cold-pool connect double-count. Driver-level returns the real error, so `op="query"` carries ok/error status.
- **`pkg/db/mysql.go`** — `NewMySQL` opens via `sql.OpenDB(instrumentedConnector{mysql.MySQLDriver{}.OpenConnector(addr)})` and hand-builds `*dbr.Connection` with a plain `NullEventReceiver` (query timing no longer comes from dbr). Pool params and migration behavior unchanged; the `"mysql"` driver stays registered.
- **`pkg/redis/instrument.go`** (new) — `RedisObserver` + `SetRedisObserver`; `instrumentClient` uses go-redis **v6** `WrapProcess`/`WrapProcessPipeline` to time each command; `redis.Nil` normalized to non-error; `pipelineErr` reports the first non-`redis.Nil` command error so a leading cache-miss can't mask a later failure.
- **`pkg/redis/redis.go`** — `New`/`NewWithOptions` install the hook.
- Both seams wrap the observer dispatch in `defer/recover` so a panicking downstream observer is contained to one dropped sample.
- Tests cover connector/conn/observer plumbing (success, error, `ErrSkip`, panic-isolation, pipeline fallback) without a live backend; the Redis command path is exercised behind `REDIS_ADDR`.

## Design notes

- **`op="query"` is execute-only, by construction.** dbr's own `Timing` starts before `database/sql` acquires a connection, so it would fold in pool-wait (and double-count connect on a cold pool). Timing at the `driver.Conn` seam avoids that — the connection is already acquired when `ExecContext`/`QueryContext` runs. (Resolves the round-1 blocker.)
- **Optional-interface preservation.** Wrapping `driver.Conn` would otherwise hide the optional interfaces `database/sql` probes for. `instrumentedConn` re-exposes every one go-sql-driver/mysql implements (`ExecerContext`, `QueryerContext`, `ConnPrepareContext`, `ConnBeginTx`, `Pinger`, `NamedValueChecker`, `SessionResetter`, `Validator`), delegating all but the two timed methods.
- **Callback signature** is `(op/cmd string, dur time.Duration, err error)` — the octo-server adapter feeds `dur` straight into `DependencyMetrics` (same metric family, `dependency=mysql/redis` label values, no new family).

## Coverage caveat

dbr interpolates args into the SQL and calls `Exec`/`QueryContext` directly (no prepared statement), so **all dbr traffic is timed**. Execs issued through an explicit `driver.Stmt` bypass the conn-level hook; dbr does not use that path.

## Follow-ups (tracked in #96)

- blocking Redis commands (`BLPop`/`BRPoplpush`) time the block-wait as latency
- `NewSqlite` not yet instrumented
- octo-server: inject `SetDBObserver`/`SetRedisObserver` → `metrics.ObserveDB/ObserveRedisCmd` (done on the octo-server branch, pending this tag)

~~`op="query"` carries no error status~~ — resolved by the driver-level move.

Refs #96

### DIFF
--- a/pkg/db/instrument.go
+++ b/pkg/db/instrument.go
@@ -37,7 +37,12 @@ func SetDBObserver(o DBObserver) {
 }
 
 // reportDB 把一次调用转发给已注入的 observer；未注入时为 no-op。
+//
+// 在接缝处兜住下游 observer 的 panic:observer 跑在每条 DB 调用的热路径上,一个
+// nil-deref 不该顺着所有查询炸上来,最多丢掉这一条埋点样本。godoc 已要求实现方
+// "不 panic",这里再加一道防线而非纯靠信任。
 func reportDB(op string, dur time.Duration, err error) {
+	defer func() { _ = recover() }()
 	if p := dbObserver.Load(); p != nil {
 		(*p)(op, dur, err)
 	}
--- a/pkg/db/instrument.go
+++ b/pkg/db/instrument.go
@@ -1,0 +1,84 @@
+package db
+
+import (
+	"context"
+	"database/sql/driver"
+	"sync/atomic"
+	"time"
+
+	"github.com/gocraft/dbr/v2"
+)
+
+// DBObserver 接收一次 MySQL 客户端调用的耗时。op 是低基数枚举：
+//
+//   - "connect"：建立一条新连接（驱动握手 + 鉴权），由 instrumentedConnector 上报，
+//     用于把「建连握手」从「取连接等待 / 执行」中隔离出来；
+//   - "query"：一条 dbr 查询（select / exec）的执行耗时，由 metricEventReceiver 上报。
+//
+// dur 为本次调用耗时；err 为调用结果（nil 表示成功）。实现方应保证轻量且不 panic，
+// 它在每条 DB 调用的热路径上同步执行。
+//
+// octo-lib 不依赖 Prometheus —— 仅暴露这个回调，由 octo-server 在启动时用
+// SetDBObserver 注入真实实现（接 #442 的 DependencyMetrics，label dependency=mysql）。
+type DBObserver func(op string, dur time.Duration, err error)
+
+// dbObserver 持有进程级 observer。用 atomic.Pointer：启动时 SetDBObserver 设置一次，
+// 业务热路径只读；未注入（指标关闭 / 单测）时 reportDB 为安全 no-op，绝不 panic。
+var dbObserver atomic.Pointer[DBObserver]
+
+// SetDBObserver 注入进程级 DB observer。传 nil 可清除（恢复 no-op）。
+// 通常在 main 启动早期调用一次。
+func SetDBObserver(o DBObserver) {
+	if o == nil {
+		dbObserver.Store(nil)
+		return
+	}
+	dbObserver.Store(&o)
+}
+
+// reportDB 把一次调用转发给已注入的 observer；未注入时为 no-op。
+func reportDB(op string, dur time.Duration, err error) {
+	if p := dbObserver.Load(); p != nil {
+		(*p)(op, dur, err)
+	}
+}
+
+// metricEventReceiver 实现 dbr.EventReceiver，把每条查询的执行耗时经
+// Timing/TimingKv 上报为 op="query"。其余事件（Event*/EventErr*）由内嵌的
+// NullEventReceiver 兜底为 no-op。
+//
+// 为何只在 Timing 上报、不在 EventErr 上报：dbr 的 select/exec 用 defer 触发
+// TimingKv，因此「失败的查询」会同时触发 TimingKv（带真实耗时）和 EventErrKv。
+// 若在两处都上报，一条失败查询会产生两条样本（一条 ok + 一条 error），既重复计数
+// 又把失败误标成 ok。故只取 Timing 这个「每条查询必触发一次」的信号，统一记为耗时
+// 样本。代价：本指标不区分查询成功/失败（错误另由 dbr 自身日志与 connect 段体现），
+// 与发版说明里的 caveat 一致。
+type metricEventReceiver struct {
+	*dbr.NullEventReceiver
+}
+
+// Timing 接收一条查询的执行耗时（纳秒），上报为 op="query"。
+func (metricEventReceiver) Timing(eventName string, nanoseconds int64) {
+	reportDB("query", time.Duration(nanoseconds), nil)
+}
+
+// TimingKv 接收一条查询的执行耗时（纳秒）及附带 kv，上报为 op="query"。
+// kv（含 sql 文本）刻意不进 label，避免基数爆炸。
+func (metricEventReceiver) TimingKv(eventName string, nanoseconds int64, kvs map[string]string) {
+	reportDB("query", time.Duration(nanoseconds), nil)
+}
+
+// instrumentedConnector 包一层 driver.Connector，对每次建连（Connect）计时并上报
+// op="connect"，从而把建连握手从执行耗时中隔离出来。Driver() 由内嵌 Connector 提升。
+type instrumentedConnector struct {
+	driver.Connector
+}
+
+// Connect 计时底层建连并上报。失败的建连同样带耗时（握手/鉴权失败也是真实耗时），
+// 故 connect 段保留完整的 ok/error 状态。
+func (c instrumentedConnector) Connect(ctx context.Context) (driver.Conn, error) {
+	start := time.Now()
+	conn, err := c.Connector.Connect(ctx)
+	reportDB("connect", time.Since(start), err)
+	return conn, err
+}
--- a/pkg/db/instrument.go
+++ b/pkg/db/instrument.go
@@ -5,21 +5,27 @@ import (
 	"database/sql/driver"
 	"sync/atomic"
 	"time"
-
-	"github.com/gocraft/dbr/v2"
 )
 
 // DBObserver 接收一次 MySQL 客户端调用的耗时。op 是低基数枚举：
 //
-//   - "connect"：建立一条新连接（驱动握手 + 鉴权），由 instrumentedConnector 上报，
-//     用于把「建连握手」从「取连接等待 / 执行」中隔离出来；
-//   - "query"：一条 dbr 查询（select / exec）的执行耗时，由 metricEventReceiver 上报。
+//   - "connect"：建立一条新连接（驱动握手 + 鉴权），由 instrumentedConnector 上报；
+//   - "query"：拿到连接之后一条语句的纯执行耗时（ExecContext / QueryContext），由
+//     instrumentedConn 上报。
 //
-// dur 为本次调用耗时；err 为调用结果（nil 表示成功）。实现方应保证轻量且不 panic，
-// 它在每条 DB 调用的热路径上同步执行。
+// 关键：query 在连接已从池中取得之后才开始计时,因此「建连握手 / 取连接等待 / 执行」
+// 三段不重叠 —— connect=握手,取连接等待看连接池的 WaitDuration,query=执行。
+//
+// dur 为本次调用耗时；err 为调用结果（nil 表示成功）。connect 与 query 都携带真实
+// err（query 走驱动层,拿得到执行错误）。实现方应保证轻量且不 panic，它在每条 DB
+// 调用的热路径上同步执行。
 //
 // octo-lib 不依赖 Prometheus —— 仅暴露这个回调，由 octo-server 在启动时用
 // SetDBObserver 注入真实实现（接 #442 的 DependencyMetrics，label dependency=mysql）。
+//
+// 覆盖范围：dbr 通过插值把参数内联进 SQL 后直接走 ExecContext/QueryContext,故其全部
+// 查询都被计时。显式预编译语句（driver.Stmt）的执行走 Stmt 而非 conn,不在此覆盖
+// （dbr 不预编译）。
 type DBObserver func(op string, dur time.Duration, err error)
 
 // dbObserver 持有进程级 observer。用 atomic.Pointer：启动时 SetDBObserver 设置一次，
@@ -48,42 +54,120 @@ func reportDB(op string, dur time.Duration, err error) {
 	}
 }
 
-// metricEventReceiver 实现 dbr.EventReceiver，把每条查询的执行耗时经
-// Timing/TimingKv 上报为 op="query"。其余事件（Event*/EventErr*）由内嵌的
-// NullEventReceiver 兜底为 no-op。
-//
-// 为何只在 Timing 上报、不在 EventErr 上报：dbr 的 select/exec 用 defer 触发
-// TimingKv，因此「失败的查询」会同时触发 TimingKv（带真实耗时）和 EventErrKv。
-// 若在两处都上报，一条失败查询会产生两条样本（一条 ok + 一条 error），既重复计数
-// 又把失败误标成 ok。故只取 Timing 这个「每条查询必触发一次」的信号，统一记为耗时
-// 样本。代价：本指标不区分查询成功/失败（错误另由 dbr 自身日志与 connect 段体现），
-// 与发版说明里的 caveat 一致。
-type metricEventReceiver struct {
-	*dbr.NullEventReceiver
-}
-
-// Timing 接收一条查询的执行耗时（纳秒），上报为 op="query"。
-func (metricEventReceiver) Timing(eventName string, nanoseconds int64) {
-	reportDB("query", time.Duration(nanoseconds), nil)
-}
-
-// TimingKv 接收一条查询的执行耗时（纳秒）及附带 kv，上报为 op="query"。
-// kv（含 sql 文本）刻意不进 label，避免基数爆炸。
-func (metricEventReceiver) TimingKv(eventName string, nanoseconds int64, kvs map[string]string) {
-	reportDB("query", time.Duration(nanoseconds), nil)
-}
-
-// instrumentedConnector 包一层 driver.Connector，对每次建连（Connect）计时并上报
-// op="connect"，从而把建连握手从执行耗时中隔离出来。Driver() 由内嵌 Connector 提升。
+// instrumentedConnector 包一层 driver.Connector：对每次建连（Connect）计时并上报
+// op="connect"，并把返回的连接包成 instrumentedConn 以便对执行计时。Driver() 由内嵌
+// Connector 提升。
 type instrumentedConnector struct {
 	driver.Connector
 }
 
 // Connect 计时底层建连并上报。失败的建连同样带耗时（握手/鉴权失败也是真实耗时），
-// 故 connect 段保留完整的 ok/error 状态。
+// 故 connect 段保留完整的 ok/error 状态。成功则把连接包成 instrumentedConn。
 func (c instrumentedConnector) Connect(ctx context.Context) (driver.Conn, error) {
 	start := time.Now()
 	conn, err := c.Connector.Connect(ctx)
 	reportDB("connect", time.Since(start), err)
-	return conn, err
+	if err != nil {
+		return nil, err
+	}
+	return instrumentedConn{conn}, nil
+}
+
+// instrumentedConn 包一层 driver.Conn，对 ExecContext/QueryContext 计时并上报
+// op="query"。连接此时已从池中取得,测的是纯执行耗时(不含取连接等待/握手)。
+//
+// database/sql 通过类型断言探测 driver.Conn 的可选接口。用接口类型内嵌 driver.Conn
+// 只会提升 Prepare/Close/Begin,其余可选接口（ExecerContext / QueryerContext /
+// ConnPrepareContext / ConnBeginTx / Pinger / NamedValueChecker / SessionResetter /
+// Validator）会被「藏掉」,导致 database/sql 退化(丢预编译上下文 / 连接复用 reset /
+// 参数检查 / 坏连接探活等)。因此这里把 go-sql-driver/mysql 实现的全部可选接口逐一
+// 重新暴露并委托;只有 Exec/Query 两条加计时。断言失败时给出与「未实现」等价的安全
+// 回退(对 mysql 而言是 dead path,纯防御)。
+type instrumentedConn struct {
+	driver.Conn
+}
+
+var (
+	_ driver.ExecerContext      = instrumentedConn{}
+	_ driver.QueryerContext     = instrumentedConn{}
+	_ driver.ConnPrepareContext = instrumentedConn{}
+	_ driver.ConnBeginTx        = instrumentedConn{}
+	_ driver.Pinger             = instrumentedConn{}
+	_ driver.NamedValueChecker  = instrumentedConn{}
+	_ driver.SessionResetter    = instrumentedConn{}
+	_ driver.Validator          = instrumentedConn{}
+)
+
+// ExecContext 计执行耗时并上报 op="query"。driver.ErrSkip 表示「请退回 Prepare 路径」,
+// 不是真实执行,不计样本。
+func (c instrumentedConn) ExecContext(ctx context.Context, query string, args []driver.NamedValue) (driver.Result, error) {
+	ec, ok := c.Conn.(driver.ExecerContext)
+	if !ok {
+		return nil, driver.ErrSkip
+	}
+	start := time.Now()
+	res, err := ec.ExecContext(ctx, query, args)
+	if err != driver.ErrSkip {
+		reportDB("query", time.Since(start), err)
+	}
+	return res, err
+}
+
+// QueryContext 计执行耗时并上报 op="query"。耗时覆盖到结果集首包返回（执行+首响应）,
+// 不含调用方后续逐行读取。
+func (c instrumentedConn) QueryContext(ctx context.Context, query string, args []driver.NamedValue) (driver.Rows, error) {
+	qc, ok := c.Conn.(driver.QueryerContext)
+	if !ok {
+		return nil, driver.ErrSkip
+	}
+	start := time.Now()
+	rows, err := qc.QueryContext(ctx, query, args)
+	if err != driver.ErrSkip {
+		reportDB("query", time.Since(start), err)
+	}
+	return rows, err
+}
+
+// 以下为可选接口的透传委托(不计时),用于在包装后保持底层驱动的完整能力。
+
+func (c instrumentedConn) PrepareContext(ctx context.Context, query string) (driver.Stmt, error) {
+	if cpc, ok := c.Conn.(driver.ConnPrepareContext); ok {
+		return cpc.PrepareContext(ctx, query)
+	}
+	return c.Prepare(query)
+}
+
+func (c instrumentedConn) BeginTx(ctx context.Context, opts driver.TxOptions) (driver.Tx, error) {
+	if cbt, ok := c.Conn.(driver.ConnBeginTx); ok {
+		return cbt.BeginTx(ctx, opts)
+	}
+	return c.Conn.Begin() //nolint:staticcheck // 非 ctx 驱动的防御回退;mysql 恒实现 ConnBeginTx
+}
+
+func (c instrumentedConn) Ping(ctx context.Context) error {
+	if p, ok := c.Conn.(driver.Pinger); ok {
+		return p.Ping(ctx)
+	}
+	return nil
+}
+
+func (c instrumentedConn) CheckNamedValue(nv *driver.NamedValue) error {
+	if nvc, ok := c.Conn.(driver.NamedValueChecker); ok {
+		return nvc.CheckNamedValue(nv)
+	}
+	return driver.ErrSkip // 退回 database/sql 默认参数转换
+}
+
+func (c instrumentedConn) ResetSession(ctx context.Context) error {
+	if sr, ok := c.Conn.(driver.SessionResetter); ok {
+		return sr.ResetSession(ctx)
+	}
+	return nil
+}
+
+func (c instrumentedConn) IsValid() bool {
+	if v, ok := c.Conn.(driver.Validator); ok {
+		return v.IsValid()
+	}
+	return true
 }
--- a/pkg/db/mysql.go
+++ b/pkg/db/mysql.go
@@ -19,7 +19,12 @@ import (
 // 不再走 dbr.Open（内部 sql.Open + 默认 driver），而是手动用
 // sql.OpenDB(instrumentedConnector{...}) 打开，并手搭 *dbr.Connection。这样：
 //   - 建连握手经 instrumentedConnector.Connect 计时（op="connect"）；
-//   - 每条查询经 metricEventReceiver 计时（op="query"）。
+//   - 拿到连接之后每条语句的纯执行经 instrumentedConn 计时（op="query"）。
+//
+// 计时落在驱动层(connect 在建连处、query 在连接已取得之后),所以「建连握手 / 取连接
+// 等待 / 执行」三段不重叠;取连接等待由连接池的 WaitDuration 体现。dbr 自身的
+// EventReceiver 不再用于打点(用 NullEventReceiver 兜底,dbr 的 Timing 含取连接等待,
+// 量的不是纯执行)。
 //
 // 池参数（MaxOpenConns/MaxIdleConns/ConnMaxLifetime）与迁移行为保持不变。
 // 注：ConnMaxLifetime 建议设为小于 MySQL 服务端 wait_timeout（默认 28800s），
@@ -38,7 +43,7 @@ func NewMySQL(addr string, maxOpenConns int, maxIdleConns int, connMaxLifetime t
 	conn := &dbr.Connection{
 		DB:            sqlDB,
 		Dialect:       dialect.MySQL,
-		EventReceiver: metricEventReceiver{&dbr.NullEventReceiver{}},
+		EventReceiver: &dbr.NullEventReceiver{},
 	}
 
 	session := conn.NewSession(nil)
--- a/pkg/db/mysql.go
+++ b/pkg/db/mysql.go
@@ -1,27 +1,45 @@
 package db
 
 import (
+	"database/sql"
 	"fmt"
 	"net/http"
 	"sort"
 	"strings"
 	"time"
 
-	_ "github.com/go-sql-driver/mysql" // mysql
+	"github.com/go-sql-driver/mysql"
 	"github.com/gocraft/dbr/v2"
+	"github.com/gocraft/dbr/v2/dialect"
 	migrate "github.com/rubenv/sql-migrate"
 )
 
 // NewMySQL 创建一个MySQL db，[path]db存储路径 [sqlDir]sql脚本目录
+//
+// 不再走 dbr.Open（内部 sql.Open + 默认 driver），而是手动用
+// sql.OpenDB(instrumentedConnector{...}) 打开，并手搭 *dbr.Connection。这样：
+//   - 建连握手经 instrumentedConnector.Connect 计时（op="connect"）；
+//   - 每条查询经 metricEventReceiver 计时（op="query"）。
+//
+// 池参数（MaxOpenConns/MaxIdleConns/ConnMaxLifetime）与迁移行为保持不变。
+// 注：ConnMaxLifetime 建议设为小于 MySQL 服务端 wait_timeout（默认 28800s），
+// 以免使用到被服务端单方面关闭的陈旧连接。
 func NewMySQL(addr string, maxOpenConns int, maxIdleConns int, connMaxLifetime time.Duration) *dbr.Session {
 
-	conn, err := dbr.Open("mysql", addr, nil)
+	base, err := mysql.MySQLDriver{}.OpenConnector(addr)
 	if err != nil {
 		panic(err)
 	}
-	conn.SetMaxOpenConns(maxOpenConns)
-	conn.SetMaxIdleConns(maxIdleConns)
-	conn.SetConnMaxLifetime(connMaxLifetime) //mysql 默认超时时间为 60*60*8=28800 SetConnMaxLifetime设置为小于数据库超时时间即可
+	sqlDB := sql.OpenDB(instrumentedConnector{Connector: base})
+	sqlDB.SetMaxOpenConns(maxOpenConns)
+	sqlDB.SetMaxIdleConns(maxIdleConns)
+	sqlDB.SetConnMaxLifetime(connMaxLifetime) //mysql 默认超时时间为 60*60*8=28800 SetConnMaxLifetime设置为小于数据库超时时间即可
+
+	conn := &dbr.Connection{
+		DB:            sqlDB,
+		Dialect:       dialect.MySQL,
+		EventReceiver: metricEventReceiver{&dbr.NullEventReceiver{}},
+	}
 
 	session := conn.NewSession(nil)
 
--- a/pkg/redis/instrument.go
+++ b/pkg/redis/instrument.go
@@ -1,0 +1,78 @@
+package redis
+
+import (
+	"sync/atomic"
+	"time"
+
+	rd "github.com/go-redis/redis"
+)
+
+// RedisObserver 接收一次 Redis 命令的耗时。cmd 是命令名（已小写，如 "get"/"set"/
+// "eval"），低基数；dur 为耗时；err 为命令结果（nil 表示成功）。
+//
+// 实现方应保证轻量且不 panic —— 它在每条命令的热路径上同步执行。
+//
+// octo-lib 不依赖 Prometheus —— 仅暴露这个回调，由 octo-server 在启动时用
+// SetRedisObserver 注入真实实现（接 #442 的 DependencyMetrics，label dependency=redis）。
+type RedisObserver func(cmd string, dur time.Duration, err error)
+
+// redisObserver 持有进程级 observer。用 atomic.Pointer：启动时 SetRedisObserver
+// 设置一次，业务热路径只读；未注入时 reportRedis 为安全 no-op，绝不 panic。
+var redisObserver atomic.Pointer[RedisObserver]
+
+// SetRedisObserver 注入进程级 Redis observer。传 nil 可清除（恢复 no-op）。
+// 通常在 main 启动早期调用一次。
+//
+// 注意：observer 是进程级单例。SetRedisObserver 影响之后经 instrumentClient 计时
+// 的所有 client 的上报目标；instrumentClient 自身在 New/NewWithOptions 构造时即已
+// 挂好 hook，与 observer 是否注入无关。
+func SetRedisObserver(o RedisObserver) {
+	if o == nil {
+		redisObserver.Store(nil)
+		return
+	}
+	redisObserver.Store(&o)
+}
+
+// reportRedis 把一次命令转发给已注入的 observer；未注入时为 no-op。
+func reportRedis(cmd string, dur time.Duration, err error) {
+	if p := redisObserver.Load(); p != nil {
+		(*p)(cmd, dur, err)
+	}
+}
+
+// instrumentClient 用 go-redis v6 的 WrapProcess/WrapProcessPipeline 给每条命令计时。
+// go-redis v6 没有 v8 的 AddHook，WrapProcess 是其等价物（包裹 process 函数）。
+//
+//   - 单条命令：op = cmd.Name()；
+//   - pipeline：作为一次往返整体上报 op = "pipeline"（管道内多命令共享一次 RTT，
+//     无法拆分单命令耗时）。
+//
+// redis.Nil（key 不存在 / 无匹配）是正常的「未命中」语义而非故障，归一为非错误，
+// 避免把命中率噪声混进 error 状态。WrapProcess 只观测、不改变返回给调用方的 err。
+func instrumentClient(client *rd.Client) {
+	client.WrapProcess(func(old func(rd.Cmder) error) func(rd.Cmder) error {
+		return func(cmd rd.Cmder) error {
+			start := time.Now()
+			err := old(cmd)
+			reportRedis(cmd.Name(), time.Since(start), normalizeRedisErr(err))
+			return err
+		}
+	})
+	client.WrapProcessPipeline(func(old func([]rd.Cmder) error) func([]rd.Cmder) error {
+		return func(cmds []rd.Cmder) error {
+			start := time.Now()
+			err := old(cmds)
+			reportRedis("pipeline", time.Since(start), normalizeRedisErr(err))
+			return err
+		}
+	})
+}
+
+// normalizeRedisErr 把 redis.Nil 归一为 nil（未命中不是故障），其余原样返回。
+func normalizeRedisErr(err error) error {
+	if err == rd.Nil {
+		return nil
+	}
+	return err
+}
--- a/pkg/redis/instrument.go
+++ b/pkg/redis/instrument.go
@@ -35,7 +35,11 @@ func SetRedisObserver(o RedisObserver) {
 }
 
 // reportRedis 把一次命令转发给已注入的 observer；未注入时为 no-op。
+//
+// 在接缝处兜住下游 observer 的 panic:observer 跑在每条命令的热路径上,一个
+// nil-deref 不该顺着 Redis 调用炸上来,最多丢掉这一条埋点样本。
 func reportRedis(cmd string, dur time.Duration, err error) {
+	defer func() { _ = recover() }()
 	if p := redisObserver.Load(); p != nil {
 		(*p)(cmd, dur, err)
 	}
@@ -63,7 +67,7 @@ func instrumentClient(client *rd.Client) {
 		return func(cmds []rd.Cmder) error {
 			start := time.Now()
 			err := old(cmds)
-			reportRedis("pipeline", time.Since(start), normalizeRedisErr(err))
+			reportRedis("pipeline", time.Since(start), pipelineErr(cmds, err))
 			return err
 		}
 	})
@@ -75,4 +79,18 @@ func normalizeRedisErr(err error) error {
 		return nil
 	}
 	return err
+}
+
+// pipelineErr 提炼一个 pipeline 批次里「真正的」错误用于打点。go-redis v6 的
+// pipeline 顶层错误是 cmdsFirstErr —— 按顺序取第一条命令的错误,因此一个打头的
+// redis.Nil（例如对不存在 key 的 GET）会把后面命令的真实失败盖住,导致整批被记成
+// 成功。这里扫描各命令,返回第一条非 redis.Nil 的命令错误;若只有 redis.Nil / 无
+// 命令错误,再回落到归一化后的顶层错误。
+func pipelineErr(cmds []rd.Cmder, topErr error) error {
+	for _, cmd := range cmds {
+		if e := cmd.Err(); e != nil && e != rd.Nil {
+			return e
+		}
+	}
+	return normalizeRedisErr(topErr)
 }
--- a/pkg/db/instrument_test.go
+++ b/pkg/db/instrument_test.go
@@ -6,8 +6,6 @@ import (
 	"errors"
 	"testing"
 	"time"
-
-	"github.com/gocraft/dbr/v2"
 )
 
 // capturedDB 记录 observer 收到的一次调用。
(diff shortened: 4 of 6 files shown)


pkg/db/instrument.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+	dbObserver.Store(&o)
+}
+
+// reportDB 把一次调用转发给已注入的 observer；未注入时为 no-op。
+func reportDB(op string, dur time.Duration, err error) {
+	if p := dbObserver.Load(); p != nil {
+		(*p)(op, dur, err)
+	}
+}
```
OctoBoooot, line 55: **[minor]** `reportDB` invokes the injected observer on every DB call with no `recover()`. The docstring tells the integrator the callback "应保证轻量且不 panic", but enforcing that in the seam is safer than trusting it — a nil-deref inside a downstream observer would propagate up through every MySQL query path instead of being lost as one stray instrumentation sample.

A `defer func() { _ = recover() }()` around the dispatch keeps the observer's blast radius to itself. Same comment applies to `pkg/redis/instrument.go:reportRedis`.

pkg/redis/instrument.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+	redisObserver.Store(&o)
+}
+
+// reportRedis 把一次命令转发给已注入的 observer；未注入时为 no-op。
+func reportRedis(cmd string, dur time.Duration, err error) {
+	if p := redisObserver.Load(); p != nil {
+		(*p)(cmd, dur, err)
+	}
+}
```
OctoBoooot, line 46: **[minor]** Same concern as in `pkg/db/instrument.go`: a panicking observer here will tear down whatever Redis command path is in flight. Single `defer recover()` around the dispatch in both `reportRedis` and `reportDB` is enough.

OctoBoooot (changes requested):
## Review: feat(db,redis): client-seam observers for MySQL connect/query and Redis command latency (#97)

**Verdict:** Request changes — the title-lint CI failure has to clear before merge; the code itself is in good shape and the only code-level note is one minor about panic-isolation in the observer hot path.

CI is red on two checks: `check-sprint` is the usual admin gate, but `pr-title-lint` is a real failure — annotation says `PR title does not match Conventional Commits: "feat(db,redis): …"`. The project's reusable lint job (`Mininglamp-OSS/.github/.../reusable-pr-title-lint.yml@v1`) rejects the comma-separated multi-scope. That has to be retitled before the verdict can flip.

### Blockers

- **`pr-title-lint` is failing because the scope `(db,redis)` is not a single Conventional-Commits token.** Annotation on check-run 83597961965 confirms it. The diff genuinely spans both packages, but the linter wants one scope — pick a single umbrella like `feat(observability):` / `feat(instrumentation):`, or split into two PRs. Once the title is updated, the synchronize event re-runs the check.

### Minor

- `pkg/db/instrument.go:43-45` / `pkg/redis/instrument.go:38-43` — `reportDB` and `reportRedis` invoke the injected observer synchronously with no `recover()`. The docstrings tell the integrator the callback "应保证轻量且不 panic", but a defensive seam in an OSS library is the right place to enforce that rather than trust it — a nil-deref inside a downstream observer would propagate up through every MySQL query / Redis command instead of being lost as one stray instrumentation sample. A trivial `defer func() { recover() }()` around the dispatch keeps the observer's blast radius to itself.

### Praise

- The "report only on `Timing`, never on `EventErr*`" decision is the kind of thinking I rarely see written down in instrumentation PRs. dbr's `defer TimingKv` semantics genuinely do produce a double-fire on failed queries if you naively hook both, and the doc comment + the negative test `TestMetricEventReceiverErrorPathDoesNotReport` together pin the invariant so a future contributor can't accidentally undo it.
- `instrumentedConnector` wrapping `driver.Connector` (rather than reaching into dbr or the mysql driver) is the cleanest place to split handshake from execute — `sql.OpenDB(connector)` is exactly the seam database/sql exposes, and the connector survives driver upgrades without code changes here. Followed the doc through and it lines up with what the test stub exercises.
- `redis.Nil → nil` normalization in `normalizeRedisErr` is the right call for a latency/error metric — conflating cache-miss with failure poisons the error rate. The dedicated `TestNormalizeRedisErr` makes the contract explicit.

### Out of scope (informational)

- `pkg/db/mysql_test.go` is flagged by the coverage-sniff heuristic, but `TestNewMySQLConstructsWithoutConnecting` in `pkg/db/instrument_test.go` does cover the `NewMySQL` happy path. Heuristic false positive; ignore.

Jerry-Xin (changes requested):
Project relevance gate: pass. This PR is in scope for `octo-lib` because it extends the existing MySQL and Redis client helpers.

🔴 Blocking

🔴 Critical: MySQL `op="query"` does not isolate execute latency as the PR claims. In `pkg/db/instrument.go` (`Timing`/`TimingKv` → `reportDB("query", ...)`), the reported duration is dbr's `dbr.exec`/`dbr.select` timing. In dbr v2.7.5 that timer is started and `defer TimingKv` is registered *before* calling `database/sql`'s `ExecContext`/`QueryContext` (see `dbr.go` `exec`/`queryRows`). Those `database/sql` calls perform pool acquisition and, on a cold or expired pool, invoke `instrumentedConnector.Connect` (wired in `pkg/db/mysql.go` via `sql.OpenDB`). Consequently a query on an empty pool emits both `op="connect"` (handshake) and `op="query"` that *includes* pool-wait + the same handshake + execution. The three segments therefore overlap rather than partition — handshake can be double-counted and pool-wait is folded into `query`, so this does not deliver the "connect handshake vs pool-wait vs execute" split the PR description promises. Either (a) document `op="query"` honestly as *total dbr operation latency* (and drop the segmentation claim), or (b) move query timing down to the driver/connection level (after a connection is acquired) if the metric must represent execute latency.

💬 Non-blocking

🟡 Warning: Redis pipeline error normalization can hide a real failure when an earlier command returns `redis.Nil`. In `pkg/redis/instrument.go`, the `WrapProcessPipeline` wrapper calls `normalizeRedisErr` on the single top-level pipeline error. go-redis v6's `defaultProcessPipeline` returns `cmdsFirstErr(cmds)` — the *first* command's error in order. For a pipeline like `[GET missing_key, <failing cmd>]`, the first error is `redis.Nil`, which normalizes to `nil`, masking the later command's real failure in the metric. If pipeline-level error fidelity matters, scan `cmds` and report the first non-`redis.Nil` command error when present. (Coarse "pipeline" op is acknowledged in the design notes, so keeping this non-blocking.)

✅ Highlights

The observer seams are relevant, low-dependency, and a safe no-op when unset (`atomic.Pointer`, nil-guarded). The MySQL connector wrapper preserves lazy connection behavior (`sql.OpenDB` stays lazy), the named mysql import keeps the `"mysql"` driver registered for sql-migrate, and pool params + migration path are unchanged. Focused unit tests cover the receiver/connector/observer plumbing without a live backend, and the `redis.Nil`-is-not-an-error normalization for single commands is the right call.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# octo-lib PR#97 Review Report
**PR:** https://github.com/Mininglamp-OSS/octo-lib/pull/97
**Head SHA:** `e53d3c79ba8a4f663d0efe41393ffba50f60b569`
**Title:** feat(db,redis): client-seam observers for MySQL connect/query and Redis command latency

---

## 1. 验证结论

| Item | Status | Evidence |
|------|--------|----------|
| `go build ./...` | ✅ Pass | Clean build, no warnings |
| `go vet ./...` | ✅ Pass | Clean |
| `go test ./pkg/db ./pkg/redis` | ✅ Pass | All tests pass |
| `dbr.Connection` manual construction | ✅ Correct | Verified struct fields (`*sql.DB`, `Dialect`, `EventReceiver`) match dbr v2.7.5 source at `dbr.go:45`. Fields `DB`, `Dialect: dialect.MySQL`, `EventReceiver: metricEventReceiver{...}` all correct types |
| `metricEventReceiver` implements `dbr.EventReceiver` | ✅ Correct | Embeds `*dbr.NullEventReceiver` (provides all 6 interface methods as no-op), overrides `Timing`/`TimingKv` |
| `instrumentedConnector` satisfies `driver.Connector` | ✅ Correct | Embeds `driver.Connector` (promotes `Driver()`), overrides `Connect(ctx)` |
| go-redis v6 `WrapProcess`/`WrapProcessPipeline` API | ✅ Verified | Signatures match `redis.go:170` and `redis.go:255` in `go-redis/redis@v6.15.9` |
| MySQL driver registration preserved | ✅ | Non-blank import `github.com/go-sql-driver/mysql` still triggers `init()` → `sql.Register("mysql", ...)` |
| `NewSession(nil)` inherits observer | ✅ | `conn.NewSession(nil)` in dbr source at `dbr.go:83` falls through to `log = conn.EventReceiver` → our `metricEventReceiver` |
| Pool params unchanged | ✅ | `SetMaxOpenConns`/`SetMaxIdleConns`/`SetConnMaxLifetime` called on `*sql.DB` from `sql.OpenDB` — identical to pre-PR behavior |
| `atomic.Pointer` thread safety | ✅ | `Load`/`Store` are atomic; loaded `*DBObserver` remains valid for the call's duration even if concurrently replaced |

## 2. 发现的问题

### P2-1: `op="query"` 不区分查询成功/失败（文档化 trade-off）

- **Severity:** P2 (informational / known limitation)
- **Diff-scope:** new — introduced by this PR
- **Evidence:** `pkg/db/instrument.go:67-74` — `Timing`/`TimingKv` always pass `nil` for `err`.
- **Root cause:** dbr's `exec()` and `queryRows()` both use `defer log.TimingKv(...)` (verified at `dbr.go:131-135` and `dbr.go:174-178`), so failed queries fire both `TimingKv` and `EventErrKv`. Reporting on both would double-count. The PR chose Timing-only, sacrificing error distinguishability.
- **Assessment:** This is a deliberate, well-documented trade-off (PR body + inline comment + follow-up #96). Not a defect. Suggest tracking resolution in #96 as planned.

### P2-2: Redis blocking commands (`BLPop`/`BRPoplpush`) time the block-wait as latency

- **Severity:** P2 (informational / known limitation)
- **Diff-scope:** new — introduced by this PR
- **Evidence:** `pkg/redis/instrument.go:59-65` — `WrapProcess` wraps `old(cmd)` and reports `time.Since(start)`. For `BLPop` with a long timeout, the duration includes the blocking wait.
- **Assessment:** Acknowledged in PR body as follow-up. The observer consumer (octo-server DependencyMetrics) should be aware that "latency" for blocking commands includes queue-wait. Not a defect in the library.

### P2-3: Pipeline reported as single `"pipeline"` sample — no per-command breakdown

- **Severity:** P2 (informational)
- **Diff-scope:** new
- **Evidence:** `pkg/redis/instrument.go:67-73` — `WrapProcessPipeline` reports `op="pipeline"` for the entire batch.
- **Assessment:** Correct design choice — pipeline shares one RTT, per-command timing isn't meaningful at this level. Consumer should treat `"pipeline"` as a separate metric.

### Nit-1: `NewSqlite` path not instrumented

- **Severity:** Nit
- **Diff-scope:** pre-existing — `sqlite.go:19` still uses `dbr.Open("sqlite3", filepath, nil)`.
- **Assessment:** Acknowledged in PR follow-ups. SQLite typically local/embedded, lower observability priority. No regression.

## 3. 建议

- For #96 follow-up: consider adding an `ok` boolean or error-status label to the observer callback when dbr's error path becomes distinguishable. This would improve metric utility without breaking the current callback signature (could add a new `DBObserverV2` with error status).
- Consider adding a brief godoc note on `SetDBObserver`/`SetRedisObserver` about concurrent access: "Safe for concurrent use. The observer runs synchronously on every DB/Redis call's hot path; implementations MUST be lightweight and MUST NOT panic."

## 4. 额外发现

- **Observer hot-path contract:** The observer callback is invoked synchronously on every DB/Redis call. A slow observer adds latency to every dependency call; a panicking observer crashes the application. This is documented in godoc comments but not enforced. Standard Go convention — caller's responsibility.
- **No resource lifecycle issues:** `atomic.Pointer[DBObserver]`/`atomic.Pointer[RedisObserver]` are process-level singletons with no goroutines, channels, or resources. `instrumentedConnector` and `metricEventReceiver` are pure wrappers with no lifecycle. `sql.OpenDB` doesn't create extra goroutines beyond what `sql.DB` normally manages.
- **Redis `WrapProcess` stacking:** If external code calls `client.WrapProcess` again after `instrumentClient`, wrappers stack. The external wrapper would be called first (outermost), then the instrument wrapper. This is standard go-redis behavior and not a concern for this PR.

## 5. 数据流回溯

### DB Query Path
```
User calls session.Select()/InsertInto()/etc.
  → dbr.exec() or dbr.queryRows() [dbr.go:116/157]
    → startTime = time.Now()
    → defer log.TimingKv("dbr.exec"/"dbr.select", time.Since(startTime).Nanoseconds(), kvs{...}) [dbr.go:131-135/174-178]
    → runner.ExecContext/QueryContext (actual DB call)
    → on completion, defer fires:
      → log = session.EventReceiver (inherited from Connection.EventReceiver via NewSession(nil))
      → Connection.EventReceiver = metricEventReceiver{&dbr.NullEventReceiver{}}
      → metricEventReceiver.TimingKv() [instrument.go:72-74]
        → reportDB("query", time.Duration(nanoseconds), nil)
        → dbObserver.Load() → if non-nil, (*p)("query", dur, nil)
```
**Verdict:** Data flows correctly. Duration is real query execution time. `err` is always nil (documented trade-off).

### DB Connect Path
```
sql.DB needs a new physical connection
  → connector.Connect(ctx) where connector = instrumentedConnector{base}
  → instrumentedConnector.Connect(ctx) [instrument.go:80-84]
    → start = time.Now()
    → conn, err = c.Connector.Connect(ctx)  // mysql.MySQLDriver connector
    → reportDB("connect", time.Since(start), err)
    → return conn, err  // err propagated correctly
```
**Verdict:** Data flows correctly. Both duration and error status are faithfully reported.

### Redis Command Path
```
User calls conn.Get()/Set()/etc.
  → rd.Client.Process(cmd)
  → c.process(cmd)  // wrapped by WrapProcess in instrumentClient
  → instrument wrapper [instrument.go:59-65]:
    → start = time.Now()
    → err = old(cmd)  // original process, includes retries
    → reportRedis(cmd.Name(), time.Since(start), normalizeRedisErr(err))
    → return err  // original err returned to caller (unmodified)
```
**Verdict:** Data flows correctly. Duration includes retries. `redis.Nil` normalized to nil for observer; caller gets original err.

## 6. 盲点 checklist (R5)

- **C1 双路径 parity:** N/A — No symmetric add/remove or create/delete paths. This PR adds instrumentation to two independent subsystems (MySQL, Redis); each is self-contained. SQLite gap is documented follow-up, not a parity defect.
- **C2 control-flow ordering / 嵌套复用:** ✅ Clear — `WrapProcess`/`WrapProcessPipeline` called once per client in `New`/`NewWithOptions`. No stacking. Observer `reportDB`/`reportRedis` are leaf functions with no re-entry.
- **C3 授权边界 ≠ 能力边界:** N/A — Library-level observability hooks. No authorization boundary changes. Observer is a process-level singleton set at startup by the application itself.
- **C4 授权生命周期 / 容器-成员状态级联:** N/A — No auth changes.
- **C5 build/note 通过 ≠ 运行期路径正确:** ✅ Verified — `go build`, `go vet`, `go test` all pass. `dbr.Connection` struct construction verified against dbr v2.7.5 source. go-redis v6 API signatures verified. This is a library (not a browser extension/CLI), so runtime path = library consumer usage, which is correctly handled.
- **C6 治理/策略/安全文档自洽性:** N/A — No governance/policy changes.

## 7. 跨轮 blocker 复检 (R6)

N/A — First review of this PR.

---

## Summary

This is a clean, well-designed observability instrumentation PR. The observer seam pattern (atomic.Pointer callback + no-op default) is sound. The manual `dbr.Connection` construction is verified correct against the dbr v2.7.5 source. All data flows (connect timing, query timing, Redis command timing) are faithfully wired with correct error propagation. Build, vet, and tests pass.

Known limitations (query error status, Redis blocking command timing, SQLite gap) are all explicitly documented in the PR body and tracked as follow-ups in #96.

No P0/P1 findings.

**[Octo-Q] verdict: APPROVE** — No blocking issues. Two P2 informational findings (documented trade-offs) and one nit (SQLite not instrumented, acknowledged as follow-up).

yujiawei (dismissed):
# Code Review — PR #97 (octo-lib)

**Head SHA reviewed:** `e53d3c79ba8a4f663d0efe41393ffba50f60b569`

Independent review (build/vet/test re-run locally; dbr, go-sql-driver, and go-redis v6 source verified). A second-opinion automated pass was also incorporated; a long-context cross-file pass timed out and did **not** contribute — see Coverage.

## 1. Spec / scope compliance — ✅

The diff matches its stated scope and does not over-reach:

- **Observer seams added at both client seams** — `DBObserver`/`SetDBObserver` (`pkg/db/instrument.go`) and `RedisObserver`/`SetRedisObserver` (`pkg/redis/instrument.go`), no-op until injected. ✓
- **No Prometheus dependency pulled in** — only callbacks are exposed; `go.mod` unchanged on that axis. ✓
- **Pool params + migration behavior unchanged** — `SetMaxOpenConns/SetMaxIdleConns/SetConnMaxLifetime` preserved; `Migration()` still runs against `session.DB` with the `"mysql"` sql-migrate dialect (`pkg/db/mysql.go:53`). ✓
- **Backward compatible** — `NewMySQL`, `redis.New`, `NewWithOptions` keep their signatures; the new API is purely additive. ✓
- No unrequested flags, fields, or behavior changes. Tests accompany the plumbing.

The hand-built `&dbr.Connection{DB, Dialect: dialect.MySQL, EventReceiver}` is behaviorally equivalent to the previous `dbr.Open("mysql", …)`: `dbr.Open` does `sql.Open("mysql", dsn)` + sets `dialect.MySQL` + the receiver, and because the MySQL driver implements `driver.DriverContext`, `sql.Open` already routes through `OpenConnector(dsn)`/`ParseDSN` — the same eager DSN parse + panic-on-bad-DSN path this PR keeps. The only added behavior is the connector/receiver instrumentation.

## 2. Code quality — Approved

No P0/P1 issues found. Correctness, thread-safety, and the error/no-op paths are sound:

- `atomic.Pointer` for both observers; `WrapProcess`/`WrapProcessPipeline` installed once at client construction before the client is handed out — no data race on the hot path, safe no-op when unset.
- `metricEventReceiver` embeds `*dbr.NullEventReceiver`, so every non-`Timing` event remains a true no-op; reporting only on `Timing`/`TimingKv` is the right choice — dbr fires `TimingKv` via `defer` on **every** query (success *and* failure, `dbr.go:130/173`), so this yields exactly one latency sample per query with no double-count against `EventErrKv`. The success/failure-blindness of `op="query"` is acknowledged and tracked in #96.
- `redis.Nil` correctly normalized to non-error (cache miss ≠ failure); `WrapProcess` only observes and passes the original `err` through untouched.

### P2 / advisory (non-blocking)

- **Connect latency is counted twice on the open path** — `pkg/db/instrument.go` (`metricEventReceiver.Timing*`). When `sql.DB` must open a connection (cold start, pool growth, `ConnMaxLifetime` expiry), dbr's `TimingKv` measures the whole `Exec/QueryContext` call, which includes `instrumentedConnector.Connect`. That handshake time therefore lands in **both** `op="connect"` and `op="query"` for those specific calls, so `query` is not pure execute on the open path. The `connect` bucket itself stays cleanly isolated, and steady-state pooled queries don't re-open, so impact is bounded — but consumers computing "execute time = query bucket" should be aware. Worth a one-line note alongside the existing #96 caveats; not a blocker.
- **Known limitations already tracked in #96** (no change requested): blocking commands (`BLPop`/`BRPoplpush`) record block-wait as latency; `op="query"` carries no ok/error status; `NewSqlite` not yet instrumented. Fine to defer.

### Security note (this PR was flagged security-sensitive — human verify downstream)

Within octo-lib the surface is clean: the callbacks receive only low-cardinality `op`/`cmd`, a duration, and an `err`. **No SQL text, no bound args, and no DSN/password ever reach the observer** — `TimingKv`'s `kvs` (which carries the SQL string) is deliberately dropped. Good.

One thing to verify in the **consumer** (octo-server #442) implementation, not here: the `connect` callback forwards the raw `err` from a failed handshake. The downstream `DependencyMetrics` must not place that raw error string into a Prometheus label — both a cardinality risk and, in edge cases, a potential connection-detail leak. octo-lib does the right thing by handing only the typed `error`; the responsibility to bucket it safely is on the injector.

## Verdict

Scope ✅ and no blocking quality issues → Approve. The P2 connect-double-count observation and the #96 follow-ups are advisory; none gate this merge.

Jerry-Xin (changes requested):
Re-review at new head `2f6dbaecea1b` (prior CR was at `e53d3c79ba8a`).

✅ Resolved since last review:
- 🟡 Redis pipeline error normalization — fixed. New `pipelineErr(cmds, topErr)` in `pkg/redis/instrument.go` scans all commands and returns the first non-`redis.Nil` error before falling back to the normalized top-level error. A leading `redis.Nil` (cache miss) no longer masks a later real failure. Good.
- 🟡 Observer panic isolation — fixed. Both seams (`reportDB`, `reportRedis`) now `defer func(){ _ = recover() }()`, so a downstream observer panic on the per-call hot path is contained to dropping one sample instead of propagating up through every query/command.

🔴 Blocking (still unaddressed) — MySQL `op="query"` does not isolate execute latency; PR body's "three segments: connect / pool-wait / execute" claim is inaccurate.
- The query-timing path is byte-unchanged at this head: `metricEventReceiver.Timing/TimingKv` → `reportDB("query", ...)` in `pkg/db/instrument.go`. The doc comments still describe `op="query"` as "执行耗时 / 查询的执行耗时".
- In dbr v2.7.5, the `Timing`/`TimingKv` duration comes from a timer started (`time.Now()`) before `runner.ExecContext`/`QueryContext`. Those `database/sql` calls acquire a connection from the pool (and, on a cold pool, perform the instrumented `Connect` handshake) before executing. So `op="query"` actually measures pool-wait + execute (and can overlap/double-count the handshake with the `connect` segment).
- This breaks the PR's stated goal of a clean connect / pool-wait / execute split — `op="query"` is not pure execute. Either (a) subtract pool `WaitDuration` to isolate execute, or (b) explicitly relabel/document `op="query"` as "pool-acquire + execute" in both the body and the code comments so consumers don't misread the metric. As-is, the headline metric semantics don't match the documentation.

The seam design, no-op-by-default safety, double-count avoidance (Timing-only), and test plumbing remain solid. Once the `op="query"` semantics are corrected or accurately documented, this is good to go.

yujiawei:
One additional advisory finding to fold into the #96 follow-ups (non-blocking, does not change my approval):

**Pipeline error normalization can mask a real failure (`pkg/redis/instrument.go`, `WrapProcessPipeline`).** go-redis v6's `generalProcessPipeline` returns `cmdsFirstErr(cmds)` — the *first* command's error in slice order (`command.go:35`). For a pipeline such as `[GET missing_key, <failing cmd>]`, the first error is `redis.Nil`, which `normalizeRedisErr` turns into `nil`, so that pipeline sample is recorded as success even though a later command failed. Impact is limited to the metric's error flag (the duration is still correct, and the original error is returned to the caller untouched), and the coarse `"pipeline"` op is already acknowledged in the design notes — so this is advisory, not a blocker. If pipeline error fidelity matters later, scan `cmds` for the first non-`redis.Nil` error instead of normalizing only the top-level result.

Also note `pr-title-lint` is now passing on the current head — the earlier title-lint failure appears to have cleared.

pkg/db/mysql.go
```diff
@@ -1,27 +1,45 @@
...
+	base, err := mysql.MySQLDriver{}.OpenConnector(addr)
 	if err != nil {
 		panic(err)
 	}
-	conn.SetMaxOpenConns(maxOpenConns)
-	conn.SetMaxIdleConns(maxIdleConns)
-	conn.SetConnMaxLifetime(connMaxLifetime) //mysql 默认超时时间为 60*60*8=28800 SetConnMaxLifetime设置为小于数据库超时时间即可
+	sqlDB := sql.OpenDB(instrumentedConnector{Connector: base})
+	sqlDB.SetMaxOpenConns(maxOpenConns)
+	sqlDB.SetMaxIdleConns(maxIdleConns)
```
OctoBoooot, line 40: **[blocker]** Co-signing Jerry-Xin's prior round-1 finding (which I missed): `op="query"` here does not measure pure execute time.

Verified in `gocraft/dbr/v2@v2.7.5/dbr.go:130-153` — `startTime := time.Now()` is set **before** `runner.ExecContext` / `runner.QueryContext`. Both of those `database/sql` calls go through `db.conn(ctx, …)`, which blocks on pool acquisition once `MaxOpenConns` is reached, and on a cold pool also calls into the instrumented `Connect`. So the value reported as `op="query"` is actually `pool-wait + (cold pool: connect) + execute` — overlapping the `pool-wait` segment the PR body claims is separately covered by `WaitDuration`, and double-counting the `connect` segment on a cold pool.

The fix is one of:

- (a) Pre-acquire the connection — `sqlConn, _ := sqlDB.Conn(ctx)` — and use a dbr Session whose `Runner` is the per-connection wrapper so dbr only times the exec on a connection already in hand. This genuinely isolates execute.
- (b) Keep the current mechanism but rename the label (e.g. `op="db_call"`) and update the docstring + PR body to say it measures "queue + execute" rather than pure execute.

Either closes the gap between the design notes and what the metric actually represents. I'll defer the choice to you; both are valid metric designs.

OctoBoooot (changes requested):
## Review: feat: add MySQL/Redis client-seam latency observers (#97) — delta vs e53d3c79

**Verdict:** Request changes — prior findings resolved, but Jerry-Xin's pool-wait conflation finding (which I missed in round 1 and have since verified myself in `gocraft/dbr/v2@v2.7.5/dbr.go:130-153`) is still unaddressed in this delta.

CI is now fully green — title is Conventional-Commits-clean and `check-sprint` is satisfied.

- My prior **blocker** on `pr-title-lint` is fixed: `feat: add MySQL/Redis client-seam latency observers` passes the lint job and the check is green at this SHA.
- My prior **minors** on panic isolation are fixed: both `reportDB` and `reportRedis` now wrap dispatch in `defer func() { _ = recover() }()`, with dedicated tests (`TestReportDBRecoversFromObserverPanic`, `TestReportRedisRecoversFromObserverPanic`) that install a panicking observer and assert the dispatch does not propagate.
- Adjacent fix from Jerry-Xin's secondary 🟡: the new `pipelineErr` walks `cmds` and returns the first non-`redis.Nil` command error, so a leading cache-miss in a pipeline no longer masks a real failure on a later command. Test `TestPipelineErrFallsBackToTopErr` covers the cmdsFirstErr fallback; the doc honestly notes that the "leading-Nil shadows real error" branch can't be unit-tested because `go-redis v6` doesn't export `setErr` and refers that to integration coverage.

### Still unaddressed (Jerry-Xin's finding, which I missed first round)

- **`pkg/db/mysql.go` is unchanged in this delta, so the `op="query"` semantic flaw stands.** Verified in `gocraft/dbr/v2@v2.7.5/dbr.go:130` (and the `queryRows` sibling at `:173`): the timer is `startTime := time.Now()` set **before** `runner.ExecContext` / `runner.QueryContext`. Those `database/sql` calls go through `db.conn(ctx, …)`, which blocks on pool acquisition when `MaxOpenConns` is saturated. So the `Timing` value reported as `op="query"` is `pool-wait + (cold-pool: instrumented Connect) + execute` — it is not a pure execute segment. The PR body's claim of a clean three-way split (`connect / pool-wait via WaitDuration / execute`) doesn't hold: the `query` segment overlaps the `pool-wait` segment, and on a cold pool also double-counts the `connect` segment. The fix is either (a) acquire the connection explicitly via `sql.DB.Conn(ctx)` and time only the subsequent exec to genuinely isolate execute, or (b) keep the current mechanism but rename the label and update the docstring/PR body to say `op="query"` measures end-to-end "queue + execute" rather than pure execute. Either resolves the description-vs-implementation gap; the choice is a metric-design call I'll defer to the author.

### Note on the verdict

Pure code quality on what shipped in this delta is good — the panic isolation and pipeline error scanning are solid, well-tested fixes. The blocker is purely the unresolved Round-1 finding from Jerry-Xin that I owe an honest co-sign on.

pkg/db/mysql.go
an9xyz: Fixed in `e3ee15d` — went with option (a), isolating execute via the driver seam.

Rather than pre-acquiring with `sqlDB.Conn(ctx)` + a per-connection runner (which would change how callers build sessions), `instrumentedConnector.Connect` now wraps the returned `driver.Conn` in an `instrumentedConn` that times `ExecContext`/`QueryContext`. Those fire only after `database/sql` has a connection in hand, so `op="query"` is now pure execute — no pool-wait, no cold-pool connect double-count. The dbr `metricEventReceiver` is dropped (the Connection uses a plain `NullEventReceiver`); pool-wait stays covered by the pool `WaitDuration`, connect by `op="connect"`.

Two side effects worth flagging:
- Driver-level returns the real execution error, so `op="query"` now carries ok/error status (also closes the #96 "no error status" item).
- The wrapper re-exposes every optional interface go-sql-driver/mysql implements (`ExecerContext`, `QueryerContext`, `ConnPrepareContext`, `ConnBeginTx`, `Pinger`, `NamedValueChecker`, `SessionResetter`, `Validator`) so wrapping the conn doesn't strip driver capabilities. Only Exec/Query are timed.

Coverage caveat (added to the PR body): execs issued through an explicit `driver.Stmt` bypass the conn-level hook; dbr interpolates args and calls Exec/QueryContext directly, so all dbr traffic is covered.

---
_Generated by [Claude Code](https://claude.ai/code)_

Jerry-Xin (dismissed):
Re-review at new head `e3ee15dec722` (my prior CR was at `2f6dbaecea1b`). The blocking 🔴 is now correctly resolved — APPROVE.

✅ 🔴 (pool-wait conflation) — fixed, and fixed the right way.
- Timing moved off dbr's `EventReceiver.Timing` (which started before `runner.ExecContext` → included pool-acquire) onto a new driver-layer `instrumentedConn` wrapper. `ExecContext`/`QueryContext` are invoked by `database/sql` only *after* a connection is checked out of the pool, so `op="query"` now measures pure execute — pool-wait is excluded and the connect / pool-wait / execute three-segment split is genuinely non-overlapping. dbr's receiver is now `NullEventReceiver` (no longer used for timing). PR body + code comments updated to match.
- Bonus: `op="query"` now carries real driver-level error status (previously the dbr `Timing` path reported `nil` only). Covered by `TestInstrumentedConnReportsQueryError`.
- `driver.ErrSkip` is correctly excluded from samples (it's the "fall back to Prepare" signal, not a real execution).

✅ driver.Conn wrapping done correctly — verified, no driver degradation.
- Wrapping a `driver.Conn` via interface embedding normally *hides* the optional interfaces `database/sql` probes by type assertion, which would silently degrade the driver (lost prepared-stmt context, session reset, named-value checking, conn liveness, ctx cancellation). The author re-exposes all 8 — `ExecerContext`, `QueryerContext`, `ConnPrepareContext`, `ConnBeginTx`, `Pinger`, `NamedValueChecker`, `SessionResetter`, `Validator` — with `var _ driver.X = instrumentedConn{}` compile-time assertions and delegation.
- Verified against go-sql-driver/mysql v1.7.1: `*mysqlConn` implements all 8, so every delegation takes the live branch; the fallbacks are pure defensive dead paths as documented. No capability is hidden.

✅ Prior 🟡s (resolved last commit, still good): Redis `pipelineErr` scans all cmds for the first non-`redis.Nil` error; observer panic isolation via `defer recover()` on both seams.

Verified locally at `e3ee15dec722`: `go build ./...` ok, `go vet ./pkg/db ./pkg/redis` ok, `go test ./pkg/db ./pkg/redis` ok. CI: pr-title-lint pass, check-sprint pass, Build/Vet/osv-scan/secret-scan pass.

Clean seam design, no-op-by-default safety, Timing-only double-count avoidance, and now an accurate execute-only query metric with full driver fidelity. Good to merge.

OctoBoooot (dismissed):
## Review: feat: add MySQL/Redis client-seam latency observers (#97) — delta vs 2f6dbaecea1b

**Verdict:** Approve. Pool-wait conflation is genuinely resolved by moving the timing seam down to the driver layer; CI is green.

- **Pool-wait conflation (my + Jerry-Xin's prior blocker) — fixed at the right level.** The `metricEventReceiver` (dbr-layer Timing hook) is gone; `instrumentedConnector.Connect` now wraps the returned `driver.Conn` in a new `instrumentedConn`, and `op="query"` is reported from `instrumentedConn.ExecContext` / `QueryContext`. By the time `database/sql` invokes those methods, the connection is already in hand — so the timer covers only the driver-level execute, not pool acquisition or the (cold-pool) connect handshake. The "three non-overlapping segments" claim in the PR body now matches the implementation.
- **Bonus: `op="query"` now carries real error status.** The prior dbr-Timing-only design couldn't distinguish ok/error; the driver-layer hook gets the actual `Exec/Query` return, and `TestInstrumentedConnReportsQueryError` pins that behavior. `driver.ErrSkip` is correctly excluded from samples (`TestInstrumentedConnExecErrSkipNotReported`).

### Praise

- The most important detail in this rewrite is the explicit re-exposure of every optional `driver.Conn` interface — `ExecerContext`, `QueryerContext`, `ConnPrepareContext`, `ConnBeginTx`, `Pinger`, `NamedValueChecker`, `SessionResetter`, `Validator` — backed by compile-time interface assertions (`var _ driver.ExecerContext = instrumentedConn{}`). `database/sql` uses type assertions on the concrete `driver.Conn` to decide between fast-path (ExecerContext / QueryerContext) and slow Prepare-then-Execute; embedding `driver.Conn` (interface) silently hides everything beyond `Begin`/`Prepare`/`Close`, which is a classic Go gotcha that silently regresses connection-reset / param-check / liveness behavior in production. The doc comment on `instrumentedConn` names this trap explicitly, the implementation handles it, and the test stub exercises the path. Exactly the right level of care for a low-level seam.
- `driver.ErrSkip` semantics are honored: the sentinel that asks `database/sql` to "fall back to Prepare" doesn't count as a real execute and isn't reported, so the metric stays clean. This is the kind of detail that's easy to miss when writing driver-layer instrumentation.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# octo-lib PR#97 Review Report — feat: add MySQL/Redis client-seam latency observers
**Head SHA:** `e3ee15dec722c8d93fa22c8cd63fa17c7ec997d0`
**Repo:** Mininglamp-OSS/octo-lib

---

## 1. Verification Summary

| Item | Result | Evidence |
|------|--------|----------|
| `go build ./pkg/db ./pkg/redis` | ✅ pass | Local build clean |
| `go vet ./pkg/db ./pkg/redis` | ✅ pass | No diagnostics |
| `go test ./pkg/db -v` | ✅ 8/8 pass | All instrument + NewMySQL tests green |
| `go test ./pkg/redis -v` | ✅ 14/14 pass (2 integration skipped — no REDIS_ADDR) | Unit tests green; integration gated on env |
| `go test -race ./pkg/db ./pkg/redis` | ✅ pass | No data races |
| dbr v2 Connection struct fields | ✅ exported | `dbr.go:45` — embedded `*sql.DB`, `Dialect`, `EventReceiver` (all promoted) |
| `dbr.NullEventReceiver` | ✅ implements EventReceiver | `event.go:29-51` — all 6 methods are no-ops |
| go-sql-driver/mysql v1.7.1 optional interfaces | ✅ all 8 implemented | `connection.go` + `driver_test.go:2194` compile-time assertions |
| go-redis v6 WrapProcess / WrapProcessPipeline | ✅ exist on `*Client` | `redis.go:170,255` via `baseClient` embedding |
| Coverage of construction sites | ✅ complete | `New` + `NewWithOptions` both call `instrumentClient`; `NewMySQL` uses `instrumentedConnector`; `NewRedis` (pkg/db) delegates to `redis.New` |

## 2. Findings

### F1 — P2 — Test isolation: global observer state mutated without restore in two tests

**Diff-scope:** new (tests introduced by this PR)

`TestReportDBNoObserverIsNoop` (`instrument_test.go:131-134`) calls `SetDBObserver(nil)` without `t.Cleanup`. `TestReportDBRecoversFromObserverPanic` (`instrument_test.go:137-141`) sets a panic observer with cleanup, but if the test before it (`TestReportDBNoObserverIsNoop`) runs first and sets nil, it's fine — but the reverse ordering could leave stale state.

Same pattern in `TestReportRedisNoObserverIsNoop` (`redis/instrument_test.go:21`).

Not a production defect (observer is set once at startup). Only a test-quality concern: if `t.Parallel()` is ever added, tests would interfere.

**Recommendation:** Add `t.Cleanup(func() { SetDBObserver(nil) })` / `SetRedisObserver(nil)` to each test that mutates the global, matching the pattern already used in `installDBObserver` and `TestSetRedisObserverRoundTrip`.

### F2 — P2 — driver.Stmt execution path not instrumented (known, documented)

**Diff-scope:** pre-existing (dbr v2 never uses prepared statements; this is inherent to the library choice)

`instrumentedConn` wraps `ExecContext`/`QueryContext` on the connection. If any caller uses `Prepare` + `Stmt.ExecContext`, that execution bypasses the instrumented conn. The PR explicitly documents this ("dbr 不预编译") and the `PrepareContext` delegation returns an unwrapped `driver.Stmt`.

Verified: dbr v2 interpolates parameters inline and calls `ExecContext`/`QueryContext` directly — no prepared-statement path in practice. Acceptable for this codebase.

### F3 — P2 — Blocking Redis commands time block-wait as latency (known, documented)

**Diff-scope:** pre-existing (WrapProcess wraps all commands uniformly; BLPop/BRPoplpush were always there)

`BLPop` and `BRPoplpush` in `redis.go` block until data arrives or timeout. WrapProcess times the entire block-wait as a single latency sample, inflating p99. PR explicitly lists this as a known follow-up.

### F4 — P2 — `NewSqlite` not instrumented (known, documented)

**Diff-scope:** pre-existing (sqlite.go unchanged)

`pkg/db/sqlite.go:19` still calls `dbr.Open("sqlite3", ...)`. PR explicitly notes this as a follow-up. Since SQLite is typically used for local/test and not production MySQL, this is acceptable scope deferral.

## 3. Suggestions

1. **Test isolation (F1):** Add `t.Cleanup` to the two bare tests. Low-effort fix.
2. **Future-proofing:** If `dbr.Connection` adds required fields in a future version, the manual struct literal in `NewMySQL` will break at compile time. Pin the dbr version or add a comment noting the coupling.

## 4. Additional Observations

- **Panic recovery is correctly implemented** at the seam (`reportDB`/`reportRedis` both `defer recover()`). A misbehaving observer cannot cascade into DB/Redis call failures. Good defense-in-depth.
- **atomic.Pointer usage is correct** — Store/Load semantics are clean, no ABA risk for a single-writer (startup) / multi-reader pattern.
- **`instrumentedConn` correctly re-exposes all 8 optional driver interfaces** (compile-time assertions verify this). Without this, `database/sql` would lose capabilities like `SessionResetter` (connection reuse hygiene), `NamedValueChecker` (parameter validation), and `Validator` (bad connection detection). This is the correct approach for a driver.Conn wrapper.
- **`pipelineErr` heuristic is sound** — scans all commands for the first non-`redis.Nil` error, falling back to normalized top-level error. The untested "head redis.Nil masking" branch requires go-redis internal `setErr` (unexported in v6), appropriately deferred to integration testing.
- **No double-counting:** The PR correctly moved timing from dbr's `EventReceiver` (which includes pool-wait) to the driver layer (pure execution). `NullEventReceiver` replaces the default, preventing duplicate metrics.

## 5. Data Flow Tracing

### MySQL path
```
app code → sess.Select/Insert/... (dbr)
  → dbr internally calls conn.DB.ExecContext/QueryContext
    → database/sql pool → picks/reuses instrumentedConn
      → [if new conn needed] instrumentedConnector.Connect(ctx)
        → base.Connect(ctx) [mysql.MySQLDriver connector]
        → reportDB("connect", dur, err) ← TIMING POINT 1
      → instrumentedConn.QueryContext(ctx, query, args)
        → type-assert c.Conn.(driver.QueryerContext) → underlying mysqlConn
        → qc.QueryContext(ctx, query, args)
        → reportDB("query", dur, err) ← TIMING POINT 2
```
**Verified:** `mysqlConn` implements `driver.ExecerContext` + `driver.QueryerContext` (compile-time asserted in go-sql-driver/mysql v1.7.1 `driver_test.go:2194`). Data flows correctly from driver through the instrumented wrapper to the observer.

### Redis path
```
app code → conn.GetString("key") → rc.client.Get("key").Result()
  → go-redis process pipeline → WrapProcess wrapper (installed by instrumentClient)
    → old(cmd) [actual Redis I/O]
    → reportRedis("get", dur, normalizeRedisErr(err)) ← TIMING POINT
      → redisObserver.Load() → if set, (*p)("get", dur, nil)
```
**Verified:** `WrapProcess` is called in both `New` (redis.go:30) and `NewWithOptions` (redis.go:49). `NewRedis` (pkg/db/redis.go:5) delegates to `redis.New`, so all Redis construction sites are covered.

### Observer lifecycle
```
octo-server main() → metrics.Setup() → SetDBObserver(fn) / SetRedisObserver(fn)
  → atomic.Pointer.Store(&fn)
  → all subsequent DB/Redis calls → reportDB/reportRedis → Load → call fn
```
**Verified:** No-op before injection (nil pointer check); panic recovery at seam; no double-counting.

## 6. Blind Spot Checklist (R5 — security_sensitive PR)

### C1 — Dual-path parity (symmetric paths)
**N/A.** This PR adds instrumentation seams, not create/delete/subscribe pairs. The observer Set/clear (`SetDBObserver(nil)`) path is trivially symmetric — Store(nil) vs Store(&fn).

### C2 — Control-flow ordering / nested reuse + security control non-canonical forms
**N/A.** No security controls (regex/escape/sanitize) are involved. The observer callback is a simple function type, not a pattern match. Ordering concern: observer fires after the DB/Redis call completes (not before), which is correct for latency measurement.

### C3 — Authorization boundary ≠ capability boundary
**N/A.** Observers report only `(op string, dur time.Duration, err error)` — operation name, duration, error status. No query content, no command arguments, no PII, no secrets. The observer cannot be used to exfiltrate data beyond what the injecting code (octo-server main) already has access to.

### C4 — Authorization lifecycle / container-member state cascade
**N/A.** No authorization changes in this PR.

### C5 — Build/note pass ≠ runtime path correct
**Clear.** Verified:
- Build passes (`go build ./pkg/db ./pkg/redis`)
- Tests pass with race detector
- `instrumentClient` is called in BOTH `New` and `NewWithOptions` (all Redis construction sites)
- `instrumentedConnector` is used in `NewMySQL` (only MySQL construction site; sqlite is out of scope)
- The instrumented wrapper is active in production runtime, not just tests
- `atomic.Pointer` is `sync/atomic` (Go 1.19+, compatible with go.mod `go 1.20`)

### C6 — Governance / policy / security document consistency
**N/A.** No governance or policy documents are modified.

## 7. Cross-round Blocker Re-check (R6)

**N/A.** First review of this PR.

---

## Severity Summary

| # | Severity | Finding |
|---|----------|---------|
| F1 | P2 | Test isolation: two tests mutate global observer without `t.Cleanup` |
| F2 | P2 | `driver.Stmt` path not instrumented (known, acceptable — dbr doesn't use prepared statements) |
| F3 | P2 | Blocking Redis commands time block-wait (known follow-up) |
| F4 | P2 | `NewSqlite` not instrumented (known follow-up) |

No P0 or P1 findings. The PR is well-structured, correctly handles concurrent access via `atomic.Pointer`, properly recovers from observer panics, re-exposes all necessary driver interfaces, and covers all MySQL and Redis construction sites.

[Octo-Q] verdict: APPROVE — Only P2/maintainability findings. Solid observability seam with correct panic recovery, proper driver interface delegation, and clean data flow. The test isolation nit (F1) is the only actionable item; the rest are documented follow-ups.

yujiawei (approved):
# Code Review — PR #97 (octo-lib)

**Head SHA reviewed:** `1dd2fe0df34598ef0646ff58f5d6dbf24689f97a`
(re-reviewed after the original `e3ee15d`; the only new commit is `test(db,redis): restore global observer via t.Cleanup` — test-only, no production change)

Independent review: full diff read; `go build ./...`, `go vet`, and `go test ./pkg/db ./pkg/redis` re-run locally (all green); and every load-bearing claim re-verified against the actual dependency source (`database/sql` stdlib, `go-sql-driver/mysql` v1.7.1, `gocraft/dbr/v2` v2.7.5, `go-redis/redis` v6.15.9). This PR is `security-sensitive` (DB/Redis client seams), so the verification was done at source level rather than trusting the doc comments.

## 1. Spec / scope compliance — ✅

The diff matches its stated scope and does not over-reach:

- **Observer seams added at both client seams, no-op until injected** — `DBObserver`/`SetDBObserver` (`pkg/db/instrument.go`) and `RedisObserver`/`SetRedisObserver` (`pkg/redis/instrument.go`), backed by `atomic.Pointer` with a safe no-op default. ✓
- **No Prometheus / metrics dependency pulled in** — only callbacks are exposed; the consumer wires the real implementation. `go.mod`/`go.sum` are untouched. ✓
- **Pool params + migration behavior unchanged** — `SetMaxOpenConns/SetMaxIdleConns/SetConnMaxLifetime` preserved on the `*sql.DB`; `Migration()` still runs against `session.DB` with the `mysql` sql-migrate dialect. ✓
- **Backward compatible / purely additive** — `NewMySQL`, `redis.New`, `NewWithOptions` keep their signatures; no unrequested flags, fields, or behavior changes. ✓

The hand-built `&dbr.Connection{DB, Dialect: dialect.MySQL, EventReceiver: &dbr.NullEventReceiver{}}` is behaviorally equivalent to the previous `dbr.Open("mysql", addr, nil)` — `dbr.Open` does exactly `sql.Open` + set `dialect.MySQL` + the receiver (verified `dbr.go:15-36`). The only added behavior is the connector/conn instrumentation.

**Spec: ✅** — no missing, no extra, no divergence.

## 2. Code Quality — Approved

The query-timing seam was correctly moved off dbr's `EventReceiver.Timing` (which started before pool acquisition, conflating pool-wait + handshake) down to the `driver.Conn` layer. This is the right fix and it holds up under inspection.

Verified correct:

- **`op="query"` is now genuinely pure execute.** `database/sql` acquires a connection (pool-wait + cold-pool handshake) inside `db.conn()` *before* `execDC`/`queryDC` invoke `ExecContext`/`QueryContext` on the driver conn. So the timer placed in `instrumentedConn.ExecContext`/`QueryContext` structurally excludes pool-wait and handshake; the `connect` segment wraps exactly `db.connector.Connect`. The three-segment split (connect / pool-wait via pool `WaitDuration` / execute) is now non-overlapping as the PR body claims.
- **`driver.Conn` wrapping does not degrade the driver.** Embedding a `driver.Conn` by interface normally *hides* the optional interfaces `database/sql` probes by type assertion. The author re-exposes all 8 — `ExecerContext`, `QueryerContext`, `ConnPrepareContext`, `ConnBeginTx`, `Pinger`, `NamedValueChecker`, `SessionResetter`, `Validator` — with compile-time `var _ driver.X = instrumentedConn{}` assertions and delegation. Enumeration is complete: the deprecated `driver.Execer`/`Queryer` are never reached because `database/sql` checks the `*Context` variant first (`sql.go:1700-1703`, `1772-1775`), and Stmt-/Rows-level optional interfaces survive because `PrepareContext`/`QueryContext` return the native `*mysqlConn` results unwrapped. Verified against `go-sql-driver/mysql` v1.7.1: `*mysqlConn` implements all 8, so every delegation takes the live branch and the fallbacks are pure defensive dead paths. The fallback returns (`ErrSkip` for `CheckNamedValue`, `nil` for `ResetSession`/`Ping`, `true` for `IsValid`) match `database/sql`'s "interface absent" semantics.
- **`driver.ErrSkip` is correctly excluded** from samples in both `ExecContext` and `QueryContext` — it's the "fall back to Prepare" signal, not a real execute. `op="query"` now also carries real driver-level error status (covered by `TestInstrumentedConnReportsQueryError`).
- **Redis pipeline error fidelity is fixed.** `pipelineErr` scans all commands for the first non-`redis.Nil` error before falling back to the normalized top-level error, so a leading cache-miss (`GET missing_key`) no longer masks a real failure later in the batch — this correctly improves on go-redis v6's `cmdsFirstErr`, which returns the first error including `redis.Nil`. `WrapProcessPipeline` wraps both `processPipeline` and `processTxPipeline`. `redis.Nil → nil` normalization for single commands keeps cache-miss out of the error rate.
- **Concurrency / panic isolation.** `atomic.Pointer` load/store is race-free; `instrumentClient` runs inside `New`/`NewWithOptions` before the client is shared. `reportDB`/`reportRedis` wrap dispatch in `defer recover()`, so a downstream observer panic is contained to dropping one sample without altering the DB/Redis return path. `NewWithOptions` shallow-copies the caller's `*rd.Options` before applying defaults — the caller's struct is not mutated.

**Quality: Approved** — no P0/P1. The items below are non-blocking suggestions (P2); none must change before merge.

### Non-blocking suggestions (P2)

- **Harden the observer-contract godoc for the metrics consumer.** The `err` forwarded to the callback is the raw driver error. A MySQL `*MySQLError` can embed row data or SQL fragments (e.g. `Error 1062 (23000): Duplicate entry 'john@example.com' for key 'users.email'`), and a Redis `proto.RedisError` echoes the raw server reply. The `op`/`cmd` labels themselves are safe (static literals / lowercased verb only — the SQL string and Redis args are never forwarded), so nothing leaks inside this library. But since the stated consumer turns this into dependency metrics, the `DBObserver`/`RedisObserver` doc comments would benefit from an explicit note: use `err` only as an ok/error boolean — never as a metric label or high-cardinality log field — because it may contain PII and unbounded cardinality.
- **Blocking Redis commands fold server-side wait into latency.** `BLPop`/`BRPoplpush` report `wait + execute` as the command duration. This is real client-observed latency (not a bug), but mixing a `BLPOP timeout=30s` into the same `dependency=redis` latency series as sub-ms `GET`/`SET` will distort p99. Worth a one-line note for the consumer to bucket by command.
- **Doc precision on `op="query"` coverage.** dbr builds its interpolator with `IgnoreBinary:true`, so a `[]byte` argument is emitted as a real placeholder rather than inlined; on a DSN without `interpolateParams=true`, the mysql driver returns `ErrSkip` and `database/sql` falls back to the Prepare→Stmt path, which is not instrumented — those queries produce no `query` sample. The instrumented code is correct (it rightly does not count an `ErrSkip`), but the coverage comment ("dbr doesn't prepare") could mention this `[]byte`-arg edge so the metric's coverage is not misread.

These are doc/consumer-contract refinements; the shipped code is correct as-is.

## 3. Overall verdict

**APPROVE** — Spec ✅ and Quality Approved. Clean seam design, no-op-by-default safety, pure-execute query timing with full `driver.Conn` fidelity, and correct pipeline error handling, all verified against the dependency source. The remaining notes are non-blocking suggestions for the doc/consumer contract.

Jerry-Xin (approved):
Re-affirm APPROVE at new head `1dd2fe0df345` (my prior APPROVE was at `e3ee15dec722`).

Delta since my last approve is **test-only** and is a clean improvement:
- `pkg/db/instrument_test.go` + `pkg/redis/instrument_test.go`: add `t.Cleanup(func(){ SetDBObserver(nil) })` / `SetRedisObserver(nil)` to the two no-observer tests, restoring global observer state after the test — addresses the cross-test-interference note. No production code changed.

All prior approvals stand: the `op="query"` execute-only timing fix (driver-layer `instrumentedConn`), full re-export of all 8 optional `driver.Conn` interfaces (verified against go-sql-driver/mysql v1.7.1), Redis `pipelineErr`, and observer panic isolation are unchanged.

Verified locally at `1dd2fe0df345`: `go test ./pkg/db ./pkg/redis` pass. Good to merge.

OctoBoooot (approved):
## Review: feat: add MySQL/Redis client-seam latency observers (#97) — delta vs e3ee15dec722

**Verdict:** Approve. Delta is two lines of test hygiene only; production code unchanged from my prior APPROVE.

CI fully green.

### Delta scope

- `pkg/db/instrument_test.go:131` and `pkg/redis/instrument_test.go:25` add a single `t.Cleanup(func() { SetDBObserver(nil) })` / `SetRedisObserver(nil)` to the respective `TestReport…NoObserverIsNoop` cases. Both tests already set the observer to nil at entry, so the Cleanup is belt-and-suspenders — but it makes the "I leave the global observer in nil state when I'm done" contract explicit, which is the right way to write tests that touch process-global state. Matches the pattern other tests in this file already follow. Addresses mochashanyao's prior 🟡 on test-isolation.

### Carry-forward

- The substantive resolution from my prior APPROVE (driver-layer `instrumentedConn` for genuine execute-only timing, full re-export of all 8 optional `driver.Conn` interfaces, panic isolation, Redis pipeline error scanning) is untouched at this SHA. Nothing else moved.

PR #104, opened 2026-06-27
feat(redis): export Instrument for raw rd.Client observability

an9xyz:
## What

Export the former unexported `instrumentClient` as **`redis.Instrument(*rd.Client)`** so a consumer can opt a **raw** go-redis client into the same per-command latency hook that `New`/`NewWithOptions` already install.

## Why

`#97` instrumented clients built via `pkg/redis` (`New`/`NewWithOptions`), which covers the shared cache (`ctx.GetRedisConn()`) — the data-plane bulk. But octo-server constructs **~15 raw `rd.NewClient(...)` instances** directly (rate-limit token bucket, OIDC state/bind/logout/locks, bot registry, user/group/space auth, health) because they need `Eval`/`Script`/`SetNX` primitives the `Conn` wrapper doesn't expose. Those bypass instrumentation entirely, so `dependency="redis"` is blind to the control plane — a real alerting hazard flagged across the #474 review (yujiawei / OctoBoooot / Octo-Q). This is the primary follow-up in #96.

With `Instrument` exported, the consumer can do:

```go
rlRedis := rd.NewClient(octoredis.MustBuildOptions(...))
redis.Instrument(rlRedis) // now its commands feed dependency="redis"
```

## Changes

- `pkg/redis/instrument.go` — rename unexported `instrumentClient` → exported `Instrument`; godoc documents the raw-client use case and warns the hook **stacks** on repeated calls (call once per client; don't re-instrument a `Conn`-built client).
- `pkg/redis/redis.go` — `New`/`NewWithOptions` now call `Instrument`.
- tests — add `TestInstrumentRawClientReportsCommands` (REDIS_ADDR-gated) driving a bare `rd.NewClient` + `Instrument` to confirm the exported path reports commands.

Purely additive: no signature changes to existing API; `New`/`NewWithOptions` behavior unchanged.

## Validation

`go build ./pkg/redis/...`, `go vet`, `go test ./pkg/redis` green; `golangci-lint` clean on the changed files. No lingering `instrumentClient` references.

Refs #96 — addresses the primary raw-client coverage follow-up; the remaining #96 items (blocking-command timing, doc nits, NewSqlite) stay open.

---
_Generated by [Claude Code](https://claude.ai/code/session_01HHTDHbMPAVkNvYdQkh2ts3)_

### DIFF
--- a/pkg/redis/instrument.go
+++ b/pkg/redis/instrument.go
@@ -23,9 +23,9 @@ var redisObserver atomic.Pointer[RedisObserver]
 // SetRedisObserver 注入进程级 Redis observer。传 nil 可清除（恢复 no-op）。
 // 通常在 main 启动早期调用一次。
 //
-// 注意：observer 是进程级单例。SetRedisObserver 影响之后经 instrumentClient 计时
-// 的所有 client 的上报目标；instrumentClient 自身在 New/NewWithOptions 构造时即已
-// 挂好 hook，与 observer 是否注入无关。
+// 注意：observer 是进程级单例。SetRedisObserver 影响之后经 Instrument 计时的所有
+// client 的上报目标；Instrument 自身在 New/NewWithOptions 构造时（或裸 client 手动
+// 调用时）即已挂好 hook，与 observer 是否注入无关。
 func SetRedisObserver(o RedisObserver) {
 	if o == nil {
 		redisObserver.Store(nil)
@@ -45,16 +45,25 @@ func reportRedis(cmd string, dur time.Duration, err error) {
 	}
 }
 
-// instrumentClient 用 go-redis v6 的 WrapProcess/WrapProcessPipeline 给每条命令计时。
-// go-redis v6 没有 v8 的 AddHook，WrapProcess 是其等价物（包裹 process 函数）。
+// Instrument 给一个 go-redis v6 的 *rd.Client 挂上每条命令的计时 hook,使其命令也灌入
+// 已注册的 RedisObserver —— 与经 New/NewWithOptions 构造的客户端完全一致。
 //
+// 用途:为**直接用 rd.NewClient 构造的裸客户端**补插桩 —— 它们往往需要 Eval/Script/
+// SetNX 等 pkg/redis 的 Conn 包装未暴露的原语(如限流令牌桶、OIDC 锁、health 探针),
+// 因而绕过了 New/NewWithOptions 的自动插桩。调用方在构造后调一次本函数即可纳入
+// dependency=redis 指标。
+//
+// 注意:New/NewWithOptions 已经调过 Instrument,**不要**对 Conn 构造出来的 client 再调;
+// hook 会在重复调用时层叠(导致重复计数)。故每个 client 只在构造后、共享前调用一次。
+//
+// 计时细节:
 //   - 单条命令：op = cmd.Name()；
 //   - pipeline：作为一次往返整体上报 op = "pipeline"（管道内多命令共享一次 RTT，
 //     无法拆分单命令耗时）。
 //
 // redis.Nil（key 不存在 / 无匹配）是正常的「未命中」语义而非故障，归一为非错误，
 // 避免把命中率噪声混进 error 状态。WrapProcess 只观测、不改变返回给调用方的 err。
-func instrumentClient(client *rd.Client) {
+func Instrument(client *rd.Client) {
 	client.WrapProcess(func(old func(rd.Cmder) error) func(rd.Cmder) error {
 		return func(cmd rd.Cmder) error {
 			start := time.Now()
--- a/pkg/redis/instrument.go
+++ b/pkg/redis/instrument.go
@@ -107,14 +107,16 @@ func Instrument(client *rd.Client) {
 	if client == nil {
 		return
 	}
+	// 全程持锁:把 WrapProcess 的安装也纳入临界区。否则并发 Instrument(同一 client)时,
+	// 第二个 goroutine 可能在第一个尚未装完 hook 时就看到「已登记」而返回,其调用方随即
+	// 发命令 —— 与 go-redis v6 未加锁的 WrapProcess 赋值并发即数据竞争。WrapProcess 只是
+	// 字段赋值、且仅启动期调用,持锁开销可忽略。
 	instrumentedMu.Lock()
+	defer instrumentedMu.Unlock()
 	if _, ok := instrumented[client]; ok {
-		instrumentedMu.Unlock()
 		return
 	}
 	instrumented[client] = struct{}{}
-	instrumentedMu.Unlock()
-
 	wrapClient(client)
 }
 
--- a/pkg/redis/instrument.go
+++ b/pkg/redis/instrument.go
@@ -1,6 +1,7 @@
 package redis
 
 import (
+	"sync"
 	"sync/atomic"
 	"time"
 
@@ -45,6 +46,18 @@ func reportRedis(cmd string, dur time.Duration, err error) {
 	}
 }
 
+// instrumented 标记已挂过 hook 的 client,使 Instrument 幂等:重复调用是 no-op。
+// go-redis v6 的 WrapProcess 每调一次就在上一层外再包一层,没有这道 guard,对同一
+// client 调两次(或对 New/NewWithOptions 构造的 client 再手动调一次)会让每条命令
+// 重复计数 2×,且无运行期信号 —— ~15 个裸 client 调用方场景下极易误触。
+//
+// 代价:map 持有 client 指针,会钉住已插桩的 client 不被 GC。本函数面向「长生命周期」
+// 客户端(启动期单例),不要在热循环里对临时 client 反复调用。
+var (
+	instrumentedMu sync.Mutex
+	instrumented   = map[*rd.Client]struct{}{}
+)
+
 // Instrument 给一个 go-redis v6 的 *rd.Client 挂上每条命令的计时 hook,使其命令也灌入
 // 已注册的 RedisObserver —— 与经 New/NewWithOptions 构造的客户端完全一致。
 //
@@ -53,8 +66,8 @@ func reportRedis(cmd string, dur time.Duration, err error) {
 // 因而绕过了 New/NewWithOptions 的自动插桩。调用方在构造后调一次本函数即可纳入
 // dependency=redis 指标。
 //
-// 注意:New/NewWithOptions 已经调过 Instrument,**不要**对 Conn 构造出来的 client 再调;
-// hook 会在重复调用时层叠(导致重复计数)。故每个 client 只在构造后、共享前调用一次。
+// 幂等且 nil-safe:client 为 nil 时直接返回;对同一 client 重复调用(或对
+// New/NewWithOptions 已自动插桩的 client 再调)是 no-op,不会层叠 hook、不会重复计数。
 //
 // 计时细节:
 //   - 单条命令：op = cmd.Name()；
@@ -64,6 +77,17 @@ func reportRedis(cmd string, dur time.Duration, err error) {
 // redis.Nil（key 不存在 / 无匹配）是正常的「未命中」语义而非故障，归一为非错误，
 // 避免把命中率噪声混进 error 状态。WrapProcess 只观测、不改变返回给调用方的 err。
 func Instrument(client *rd.Client) {
+	if client == nil {
+		return
+	}
+	instrumentedMu.Lock()
+	if _, ok := instrumented[client]; ok {
+		instrumentedMu.Unlock()
+		return
+	}
+	instrumented[client] = struct{}{}
+	instrumentedMu.Unlock()
+
 	client.WrapProcess(func(old func(rd.Cmder) error) func(rd.Cmder) error {
 		return func(cmd rd.Cmder) error {
 			start := time.Now()
--- a/pkg/redis/instrument.go
+++ b/pkg/redis/instrument.go
@@ -89,8 +89,10 @@ func wrapClient(client *rd.Client) {
 // 因而绕过了 New/NewWithOptions 的自动插桩。调用方在构造后调一次本函数即可纳入
 // dependency=redis 指标。
 //
-// 时机:必须在 client 被共享 / 发起命令**之前**调用。go-redis v6 的 WrapProcess 赋值
-// 未加锁,与在途命令并发会 race。
+// 时机:必须在 client 被共享 / 发起命令 / 被 WithContext 克隆**之前**调用。go-redis v6
+// 的 WrapProcess 赋值未加锁,与在途命令并发会 race;且 client.WithContext(ctx) 返回的是
+// 克隆,克隆时即固化当时的 process —— 先克隆、后对原 client 调 Instrument,克隆不会继承
+// 本 hook。需要被观测的克隆,应对其本身调用 Instrument,或在克隆前就插桩好原 client。
 //
 // 幂等且 nil-safe:client 为 nil 时直接返回;对同一 client 重复调用是 no-op,不会层叠
 // hook、不会重复计数。注意它会在内部表里持有该 client 的引用(见 instrumented),故面向
--- a/pkg/redis/instrument.go
+++ b/pkg/redis/instrument.go
@@ -46,17 +46,40 @@ func reportRedis(cmd string, dur time.Duration, err error) {
 	}
 }
 
-// instrumented 标记已挂过 hook 的 client,使 Instrument 幂等:重复调用是 no-op。
-// go-redis v6 的 WrapProcess 每调一次就在上一层外再包一层,没有这道 guard,对同一
-// client 调两次(或对 New/NewWithOptions 构造的 client 再手动调一次)会让每条命令
-// 重复计数 2×,且无运行期信号 —— ~15 个裸 client 调用方场景下极易误触。
+// instrumented 标记已经过 Instrument(公开入口)挂过 hook 的 client,使其幂等:
+// 重复调用是 no-op。go-redis v6 的 WrapProcess 每调一次就在上一层外再包一层,没有
+// 这道 guard,对同一 client 调两次会让每条命令重复计数 2×,且无运行期信号 —— ~15 个
+// 裸 client 调用方场景下极易误触。
 //
-// 代价:map 持有 client 指针,会钉住已插桩的 client 不被 GC。本函数面向「长生命周期」
-// 客户端(启动期单例),不要在热循环里对临时 client 反复调用。
+// 仅公开的 Instrument 登记此表;New/NewWithOptions 走内部 wrapClient,不登记 —— 它们
+// 对刚构造、必然只插一次的 client 操作,无重复风险,也就不必把(可能短生命周期的)
+// New 构造 client 钉在表里不被 GC。所以本表只持有调用方显式传入的长生命周期裸 client。
 var (
 	instrumentedMu sync.Mutex
 	instrumented   = map[*rd.Client]struct{}{}
 )
+
+// wrapClient 把每条命令的计时 hook 挂到 client 上(WrapProcess/WrapProcessPipeline)。
+// 这是内部低层实现,不做 nil / 幂等 guard —— 供 New/NewWithOptions 对自家刚构造的
+// client 调用。公开的 Instrument 在其之上叠加 nil + 幂等 guard。
+func wrapClient(client *rd.Client) {
+	client.WrapProcess(func(old func(rd.Cmder) error) func(rd.Cmder) error {
+		return func(cmd rd.Cmder) error {
+			start := time.Now()
+			err := old(cmd)
+			reportRedis(cmd.Name(), time.Since(start), normalizeRedisErr(err))
+			return err
+		}
+	})
+	client.WrapProcessPipeline(func(old func([]rd.Cmder) error) func([]rd.Cmder) error {
+		return func(cmds []rd.Cmder) error {
+			start := time.Now()
+			err := old(cmds)
+			reportRedis("pipeline", time.Since(start), pipelineErr(cmds, err))
+			return err
+		}
+	})
+}
 
 // Instrument 给一个 go-redis v6 的 *rd.Client 挂上每条命令的计时 hook,使其命令也灌入
 // 已注册的 RedisObserver —— 与经 New/NewWithOptions 构造的客户端完全一致。
@@ -66,8 +89,12 @@ var (
 // 因而绕过了 New/NewWithOptions 的自动插桩。调用方在构造后调一次本函数即可纳入
 // dependency=redis 指标。
 //
-// 幂等且 nil-safe:client 为 nil 时直接返回;对同一 client 重复调用(或对
-// New/NewWithOptions 已自动插桩的 client 再调)是 no-op,不会层叠 hook、不会重复计数。
+// 时机:必须在 client 被共享 / 发起命令**之前**调用。go-redis v6 的 WrapProcess 赋值
+// 未加锁,与在途命令并发会 race。
+//
+// 幂等且 nil-safe:client 为 nil 时直接返回;对同一 client 重复调用是 no-op,不会层叠
+// hook、不会重复计数。注意它会在内部表里持有该 client 的引用(见 instrumented),故面向
+// 长生命周期客户端(启动期单例),不要在热循环里对临时 client 反复调用。
 //
 // 计时细节:
 //   - 单条命令：op = cmd.Name()；
@@ -88,22 +115,7 @@ func Instrument(client *rd.Client) {
 	instrumented[client] = struct{}{}
 	instrumentedMu.Unlock()
 
-	client.WrapProcess(func(old func(rd.Cmder) error) func(rd.Cmder) error {
-		return func(cmd rd.Cmder) error {
-			start := time.Now()
-			err := old(cmd)
-			reportRedis(cmd.Name(), time.Since(start), normalizeRedisErr(err))
-			return err
-		}
-	})
-	client.WrapProcessPipeline(func(old func([]rd.Cmder) error) func([]rd.Cmder) error {
-		return func(cmds []rd.Cmder) error {
-			start := time.Now()
-			err := old(cmds)
-			reportRedis("pipeline", time.Since(start), pipelineErr(cmds, err))
-			return err
-		}
-	})
+	wrapClient(client)
 }
 
 // normalizeRedisErr 把 redis.Nil 归一为 nil（未命中不是故障），其余原样返回。
--- a/pkg/redis/instrument_test.go
+++ b/pkg/redis/instrument_test.go
@@ -167,6 +167,22 @@ func TestNewClientInstrumentedReportsCommands(t *testing.T) {
 	}
 }
 
+// New/NewWithOptions 走内部 wrapClient,不应把自家 client 登记进全局幂等表 —— 否则
+// 短生命周期的 New() client 会被钉住不回收(Jerry-Xin review)。非集成,CI 可跑。
+func TestNewDoesNotRetainInGuardMap(t *testing.T) {
+	count := func() int {
+		instrumentedMu.Lock()
+		defer instrumentedMu.Unlock()
+		return len(instrumented)
+	}
+	before := count()
+	conn := NewWithOptions(&rd.Options{Addr: "127.0.0.1:1"}) // 惰性,不会真正连接
+	defer func() { _ = conn.Close() }()
+	if after := count(); after != before {
+		t.Fatalf("New/NewWithOptions must not register clients in the idempotence map: len %d -> %d", before, after)
+	}
+}
+
 // 集成路径：验证导出的 Instrument 能给「裸」*rd.Client 补插桩 —— 即不经
 // New/NewWithOptions、需要 Eval/SetNX 等原语而直接构造的客户端(限流/锁/health)。
 // 同时验证幂等：连调两次 Instrument,一条 GET 只应产生一条样本（无 hook 层叠/重复计数）。
--- a/pkg/redis/instrument_test.go
+++ b/pkg/redis/instrument_test.go
@@ -122,3 +122,41 @@ func TestInstrumentClientReportsCommands(t *testing.T) {
 		t.Fatalf("expected a 'get' sample, got %+v", samples)
 	}
 }
+
+// 集成路径：验证导出的 Instrument 能给「裸」*rd.Client 补插桩 —— 即不经
+// New/NewWithOptions、需要 Eval/SetNX 等原语而直接构造的客户端(限流/锁/health)。
+// 需要真实 Redis。
+func TestInstrumentRawClientReportsCommands(t *testing.T) {
+	addr := os.Getenv("REDIS_ADDR")
+	if addr == "" {
+		t.Skip("REDIS_ADDR not set, skipping integration test")
+	}
+
+	var (
+		mu     sync.Mutex
+		sawGet bool
+	)
+	SetRedisObserver(func(cmd string, _ time.Duration, _ error) {
+		mu.Lock()
+		if cmd == "get" {
+			sawGet = true
+		}
+		mu.Unlock()
+	})
+	t.Cleanup(func() { SetRedisObserver(nil) })
+
+	// 裸客户端,不经 New/NewWithOptions,手动 Instrument。
+	client := rd.NewClient(&rd.Options{Addr: addr, Password: os.Getenv("REDIS_PASSWORD")})
+	defer func() { _ = client.Close() }()
+	Instrument(client)
+
+	if err := client.Get("test:instrument:raw:missing").Err(); err != nil && err != rd.Nil {
+		t.Fatalf("Get: %v", err)
+	}
+
+	mu.Lock()
+	defer mu.Unlock()
+	if !sawGet {
+		t.Fatal("expected a 'get' sample from the manually-instrumented raw client")
+	}
+}
--- a/pkg/redis/instrument_test.go
+++ b/pkg/redis/instrument_test.go
@@ -77,8 +77,52 @@ func TestSetRedisObserverRoundTrip(t *testing.T) {
 	}
 }
 
-// 集成路径：需要真实 Redis。验证 WrapProcess 把命令名/未命中正确灌给 observer。
-func TestInstrumentClientReportsCommands(t *testing.T) {
+// 这些集成测试共享进程级 observer 单例,勿加 t.Parallel()。
+
+// Instrument(nil) 必须是安全 no-op,不得 panic。
+func TestInstrumentNilIsNoop(t *testing.T) {
+	Instrument(nil)
+}
+
+// 幂等性（非集成,CI 可跑）：对同一 client 连调两次 Instrument,一条命令只应产生一条
+// 样本。命令打向无人监听的地址会立即失败,但 WrapProcess 仍按「包了几层」触发对应次数
+// 的上报 —— 借此在没有真实 Redis 的情况下探测 hook 是否层叠。
+func TestInstrumentIdempotent(t *testing.T) {
+	var (
+		mu       sync.Mutex
+		getCount int
+	)
+	SetRedisObserver(func(cmd string, _ time.Duration, _ error) {
+		mu.Lock()
+		if cmd == "get" {
+			getCount++
+		}
+		mu.Unlock()
+	})
+	t.Cleanup(func() { SetRedisObserver(nil) })
+
+	c := rd.NewClient(&rd.Options{Addr: "127.0.0.1:1", MaxRetries: 0, DialTimeout: 200 * time.Millisecond})
+	defer func() { _ = c.Close() }()
+	t.Cleanup(func() {
+		instrumentedMu.Lock()
+		delete(instrumented, c)
+		instrumentedMu.Unlock()
+	})
+
+	Instrument(c)
+	Instrument(c) // 幂等：第二次应为 no-op,不再叠一层 hook
+
+	_ = c.Get("k").Err() // 连接失败会返回 error,但 hook 仍会触发上报
+
+	mu.Lock()
+	defer mu.Unlock()
+	if getCount != 1 {
+		t.Fatalf("Instrument should be idempotent: want 1 'get' sample, got %d (hook stacked?)", getCount)
+	}
+}
+
+// 集成路径：需要真实 Redis。验证经 New（→Instrument）的 client 把命令名/未命中正确灌给 observer。
+func TestNewClientInstrumentedReportsCommands(t *testing.T) {
 	addr := os.Getenv("REDIS_ADDR")
 	if addr == "" {
 		t.Skip("REDIS_ADDR not set, skipping integration test")
@@ -125,6 +169,7 @@ func TestInstrumentClientReportsCommands(t *testing.T) {
 
 // 集成路径：验证导出的 Instrument 能给「裸」*rd.Client 补插桩 —— 即不经
 // New/NewWithOptions、需要 Eval/SetNX 等原语而直接构造的客户端(限流/锁/health)。
+// 同时验证幂等：连调两次 Instrument,一条 GET 只应产生一条样本（无 hook 层叠/重复计数）。
 // 需要真实 Redis。
 func TestInstrumentRawClientReportsCommands(t *testing.T) {
 	addr := os.Getenv("REDIS_ADDR")
@@ -133,21 +178,27 @@ func TestInstrumentRawClientReportsCommands(t *testing.T) {
 	}
 
 	var (
-		mu     sync.Mutex
-		sawGet bool
+		mu       sync.Mutex
+		getCount int
 	)
 	SetRedisObserver(func(cmd string, _ time.Duration, _ error) {
 		mu.Lock()
 		if cmd == "get" {
-			sawGet = true
+			getCount++
 		}
 		mu.Unlock()
 	})
 	t.Cleanup(func() { SetRedisObserver(nil) })
 
-	// 裸客户端,不经 New/NewWithOptions,手动 Instrument。
+	// 裸客户端,不经 New/NewWithOptions,手动 Instrument 两次（幂等：第二次应为 no-op）。
 	client := rd.NewClient(&rd.Options{Addr: addr, Password: os.Getenv("REDIS_PASSWORD")})
 	defer func() { _ = client.Close() }()
+	t.Cleanup(func() { // 别让测试用 client 滞留在全局幂等表里
+		instrumentedMu.Lock()
+		delete(instrumented, client)
+		instrumentedMu.Unlock()
+	})
+	Instrument(client)
 	Instrument(client)
 
 	if err := client.Get("test:instrument:raw:missing").Err(); err != nil && err != rd.Nil {
@@ -156,7 +207,7 @@ func TestInstrumentRawClientReportsCommands(t *testing.T) {
 
 	mu.Lock()
 	defer mu.Unlock()
-	if !sawGet {
-		t.Fatal("expected a 'get' sample from the manually-instrumented raw client")
+	if getCount != 1 {
+		t.Fatalf("expected exactly 1 'get' sample (idempotent instrument), got %d", getCount)
 	}
 }
--- a/pkg/redis/redis.go
+++ b/pkg/redis/redis.go
@@ -27,7 +27,7 @@ func New(addr string, password string) *Conn {
 		MaxRetries: 3, // 失败重试次数
 		Password:   password,
 	})
-	Instrument(c.client)
+	wrapClient(c.client)
 	return c
 }
 
@@ -46,7 +46,7 @@ func NewWithOptions(opts *rd.Options) *Conn {
 		local.MaxRetries = 3
 	}
 	client := rd.NewClient(&local)
-	Instrument(client)
+	wrapClient(client)
 	return &Conn{client: client}
 }
 
--- a/pkg/redis/redis.go
+++ b/pkg/redis/redis.go
@@ -27,7 +27,7 @@ func New(addr string, password string) *Conn {
 		MaxRetries: 3, // 失败重试次数
 		Password:   password,
 	})
-	instrumentClient(c.client)
+	Instrument(c.client)
 	return c
 }
 
@@ -46,7 +46,7 @@ func NewWithOptions(opts *rd.Options) *Conn {
 		local.MaxRetries = 3
 	}
 	client := rd.NewClient(&local)
-	instrumentClient(client)
+	Instrument(client)
 	return &Conn{client: client}
 }
 


pkg/redis/instrument.go
```diff
@@ -45,16 +45,25 @@ func reportRedis(cmd string, dur time.Duration, err error) {
...
+//
+// 计时细节:
 //   - 单条命令：op = cmd.Name()；
 //   - pipeline：作为一次往返整体上报 op = "pipeline"（管道内多命令共享一次 RTT，
 //     无法拆分单命令耗时）。
 //
 // redis.Nil（key 不存在 / 无匹配）是正常的「未命中」语义而非故障，归一为非错误，
 // 避免把命中率噪声混进 error 状态。WrapProcess 只观测、不改变返回给调用方的 err。
-func instrumentClient(client *rd.Client) {
+func Instrument(client *rd.Client) {
```
OctoBoooot, line 108: **[minor]** Now that `Instrument` is exported and the WrapProcess hook stacks on repeated calls, the no-double-instrument contract is enforced only by the godoc. With ~15 raw-client adopters incoming, a second accidental call (or instrumenting a `New`-built client's underlying handle) silently doubles every latency sample with no runtime signal.

Worth considering a cheap idempotence guard — e.g. a package-level `sync.Map[*rd.Client]struct{}` marker so a repeat `Instrument(c)` is a no-op. If you'd rather keep it doc-only, that's defensible given the data-plane clients are out of reach of this export — just flagging it's an easy footgun to trip and hard to notice post-merge.

OctoBoooot (commented):
## Review: feat(redis): export Instrument for raw rd.Client observability (#104)

**Verdict:** Comment — clean additive change, no code defects. One design question on the stacking footgun; otherwise approvable once CI is green.

CI is red only on `check-sprint` (org admin gate — author can't fix; not a code failure). Build/Test/Vet are green locally and pending in CI.

### Questions

- `pkg/redis/instrument.go:46-70` — `Instrument` is now exported and its hook **stacks** on repeated calls (correctly documented in the godoc). But the contract is enforced only by the doc comment: calling `Instrument` twice on the same client, or calling it on a `New`/`NewWithOptions`-built `Conn`'s underlying client, silently double-counts every command with no runtime guard. Given ~15 raw-client call sites are about to adopt this, is a guard worth it — e.g. a `sync.Map`/marker tracking already-instrumented clients so a second call is a no-op? Not blocking (the doc is clear and the data-plane clients are out of reach of this export), but with this many adopters a footgun that manifests as 2× latency-sample counts is easy to trip and hard to notice. If the answer is "doc is enough," that's a defensible call — just confirming it's deliberate.

### Praise

- The rename keeps the seam honest: `New`/`NewWithOptions` route through the same exported `Instrument`, so the auto-instrumented and manually-instrumented paths can't drift apart. The grep confirms zero lingering `instrumentClient` references — clean rename, no dead aliases. The PR body is exemplary: it names *why* (~15 raw `rd.NewClient` sites needing `Eval`/`Script`/`SetNX` that the `Conn` wrapper doesn't expose, blind to `dependency="redis"`), ties it to the #474 review and #96 follow-up, and is explicit that it's purely additive with no signature changes.

Jerry-Xin (dismissed):
The PR is in scope for `octo-lib`: it extends existing `pkg/redis` observability behavior to raw `go-redis` clients without changing existing `New` / `NewWithOptions` behavior.

💬 Non-blocking
- 🔵 Suggestion: Consider whether `Instrument(nil)` should be a documented panic or a no-op guard in `pkg/redis/instrument.go:66`. Current behavior is acceptable for an instrumentation helper, especially since callers are expected to pass a freshly constructed client.

✅ Highlights
- `Instrument` preserves the existing hook implementation and error normalization behavior, so exported raw-client instrumentation matches wrapped clients: `pkg/redis/instrument.go:66`.
- Existing constructors still install instrumentation automatically: `pkg/redis/redis.go:30`, `pkg/redis/redis.go:49`.
- The new integration test covers the intended raw `rd.NewClient` path: `pkg/redis/instrument_test.go:129`.

Validation performed:
- `go test ./pkg/redis`
- `go test -race ./pkg/redis`
- `go test ./...`

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# octo-lib PR#104 Review Report — feat(redis): export Instrument for raw rd.Client observability
**Head SHA:** d1681772b1fd4dab4d84103ed72b42d0ba102f06
**Files:** pkg/redis/instrument.go (+15/-6), pkg/redis/instrument_test.go (+38/-0), pkg/redis/redis.go (+2/-2)

---

## 1. Verification Summary

| Check | Status | Evidence |
|---|---|---|
| Rename complete (no lingering `instrumentClient` refs) | ✅ | grep: only match is test function *name* `TestInstrumentClientReportsCommands` — cosmetic, not a code reference |
| `go build ./pkg/redis/...` | ✅ | Clean compile |
| `go vet ./pkg/redis/...` | ✅ | Clean |
| Unit tests (`TestNormalize*`, `TestReportRedis*`, `TestPipelineErr*`, `TestSetRedisObserver*`) | ✅ | All PASS |
| Godoc updated with stacking warning | ✅ | instrument.go:53-65 — explicit "do not re-instrument" warning |
| Data-flow: Instrument → WrapProcess → reportRedis → observer | ✅ | Traced end-to-end, see §5 |

---

## 2. Findings

### No P0 or P1 findings.

### P2 — Test singleton fragility (pre-existing, amplified by addition)

**Diff-scope three questions:**
1. **Pre-existing** — the `redisObserver` process-level singleton is a pre-existing design constraint, not introduced by this PR.
2. The PR adds a second integration test (`TestInstrumentRawClientReportsCommands`) that calls `SetRedisObserver(...)` on the same global singleton as `TestInstrumentClientReportsCommands`.
3. Both tests use `t.Cleanup(func() { SetRedisObserver(nil) })`, so sequential execution is safe. But `go test -parallel > 1` would cause observer cross-talk between these two tests.

**Severity rationale:** Not a regression — follows the established test pattern. Flagging for awareness. A future refactor to per-test observer injection (via `testing.TB` helper or context) would eliminate this fragility for all instrument tests.

### Nit — Stale test function name

`TestInstrumentClientReportsCommands` (instrument_test.go:81) still references the old unexported function name in its identifier. The test validates the `New()` → `Instrument` path, so a name like `TestNewClientInstrumentedReportsCommands` would be clearer. Purely cosmetic — not blocking.

---

## 3. Suggestions

- Consider adding a `// These tests share a process-level observer singleton; do not add t.Parallel().` comment above the integration tests to protect future contributors from parallelism footguns.
- (Future) If the number of integration tests grows, consider refactoring the observer to support per-test injection.

---

## 4. Additional Observations

- **Hook stacking is documented but not guarded at runtime.** The godoc warns "don't call Instrument twice on the same client," but the function doesn't detect or prevent it (e.g., with a `sync.Once` or a marker field). Acceptable for now — the API contract is clear in the docs, and adding runtime guards would complicate the zero-cost abstraction.
- **octo-server is the primary consumer** of `rd.NewClient` raw instances (~15 sites per PR body). This PR correctly leaves octo-server integration to a follow-up; the library change is self-contained.

---

## 5. Data-Flow Trace

```
rd.NewClient(opts)
  → *rd.Client (bare, no hooks)

Instrument(client)  [instrument.go:66]
  → client.WrapProcess(wrapFn)        // wraps c.process = wrapFn(c.process)
  → client.WrapProcessPipeline(wrapFn) // wraps c.processPipeline

client.Get("key")
  → c.process(cmd)                    // now the wrapped version
    → start := time.Now()
    → err := old(cmd)                 // original go-redis process
    → reportRedis(cmd.Name(), time.Since(start), normalizeRedisErr(err))
      → reportRedis [instrument.go:43]
        → defer recover() (panic-safe)
        → p := redisObserver.Load()
        → if p != nil: (*p)(cmd, dur, err)  // observer callback

normalizeRedisErr(err) [instrument.go:88]
  → redis.Nil → nil (cache miss ≠ failure)
  → other errors pass through

pipelineErr(cmds, topErr) [instrument.go:97]
  → scan cmds for first non-Nil error
  → fallback: normalizeRedisErr(topErr)
```

**Verdict on data flow:** Every consumed value traces to a valid upstream source. `cmd.Name()` comes from go-redis's `Cmder` interface (lowercase, low-cardinality). `time.Since(start)` is locally scoped. `normalizeRedisErr` correctly maps `redis.Nil → nil`. The `redisObserver` atomic pointer is safe for concurrent reads; writes are expected only at startup. ✅

---

## 6. Blind-Spot Checklist (R5 — security_sensitive PR, all items answered)

- **C1 — Dual-path parity:** N/A. No symmetric create/delete or subscribe/unsubscribe paths. This is a pure rename + export.
- **C2 — Control-flow ordering / nested reuse:** Clear. `Instrument` calls `WrapProcess` and `WrapProcessPipeline` exactly once each, in a fixed order. No nesting or re-entry risk. The stacking warning in godoc is accurate for go-redis v6's `WrapProcess` semantics (`c.process = fn(c.process)` — each call wraps the previous).
- **C3 — Authorization boundary ≠ capability boundary:** N/A. `Instrument` is a library function in a public package. Any importer can call it, which is the intended design. The observer singleton is already accessible via `SetRedisObserver`. No new authorization surface.
- **C4 — Authorization lifecycle / container-member state cascade:** N/A. No auth mechanism.
- **C5 — build/note ≠ runtime correctness:** N/A. Pure Go library change — no build artifacts, browser extensions, CLI packaging, or relative-URL concerns. `go build` and `go vet` pass. Runtime data flow verified in §5.
- **C6 — Governance / security doc self-consistency:** N/A. No governance, SECURITY.md, or policy documents are modified.

---

## 7. Cross-Round Blocker Recheck (R6)

---

[Octo-Q] verdict: APPROVE — No P0/P1 findings. Clean rename of an unexported helper to an exported API with accurate godoc, a well-structured integration test, and verified data-flow integrity. The P2 test-singleon fragility is pre-existing and follows established patterns. Recommend landing as-is with the optional suggestions from §3.

yujiawei (dismissed):
# Code Review — PR #104 (octo-lib)

**Verdict: APPROVED** — clean, purely-additive visibility change with sound docs and green checks. No blocking findings.

## Summary

This PR renames the unexported `instrumentClient(*rd.Client)` to the exported `Instrument(*rd.Client)`, rewires the two internal callers (`New`, `NewWithOptions`), refreshes the godoc, and adds an integration test. The goal — letting consumers opt a raw `rd.NewClient(...)` instance into the same per-command latency hook so the control-plane clients (rate-limit token bucket, OIDC locks, health probes, etc.) stop being blind to `dependency="redis"` — is well-motivated and the implementation is minimal.

## Spec compliance ✅

- **No missing work:** rename done; both internal call sites (`redis.go` `New`, `NewWithOptions`) updated; zero lingering `instrumentClient` references remain in the repo.
- **No over-build:** the `Instrument` body is byte-for-byte identical to the former unexported function — this is a visibility-only change. No new parameters, flags, fields, or behavioral changes were introduced. `New`/`NewWithOptions` semantics are unchanged.
- **No deviation:** matches the stated "purely additive, no signature changes to existing API" intent.

## Code quality — Approved

Verified against the checked-out tree at head `d1681772`:

- **Security (PR is `security_sensitive`):** The observer callback receives only `cmd.Name()` (the lowercased command verb, e.g. `get`/`eval`), the duration, and a normalized error — never command arguments, keys, values, or passwords. go-redis `baseCmd.Name()` returns only `stringArg(0)`, so secrets cannot reach the metric label via this path. `Instrument` touches no connection options, TLS, or auth; it only attaches read-only timing closures to a client the caller already owns. No new exfiltration or privilege path.
- **Hot-path safety:** `reportRedis` already wraps the observer in `defer recover()` (pre-existing, unchanged); a misbehaving observer drops at most one sample rather than propagating a panic up the Redis call.
- **Idempotency / double-counting:** the hook stacks if `Instrument` is called more than once on the same client. This is correctly documented in the new godoc as a "call once per client; don't re-instrument a `Conn`-built client" warning. The two internal constructors each call it exactly once, before the client is shared.
- **Concurrency:** `WrapProcess`/`WrapProcessPipeline` are not thread-safe setters, but both constructors call `Instrument` before returning the client, so setup completes before the client is observable to other goroutines. The godoc steers external callers to the same "instrument before share" discipline.
- **Test:** `TestInstrumentRawClientReportsCommands` exercises the exported path (bare `rd.NewClient` → `Instrument` → `Get` → asserts the observer saw `get`). Its `REDIS_ADDR` skip-gate mirrors the existing sibling integration test verbatim — established convention, not a new gap; the pure-logic unit tests remain ungated and run in CI.
- `go build`, `go vet`, and `go test ./pkg/redis/...` all pass locally.

## Non-blocking notes (no action required to merge)

- The raw error object reaching the observer could, in principle, carry a key name inside an error string if a downstream observer stringified it into a label. This is pre-existing behavior, unchanged and not widened by this diff, and is the observer implementation's responsibility — not a defect of this PR.
- Out of this repo's scope, but worth a downstream eye: the ~15 raw-client construction sites in the consumer must each call `Instrument` exactly once and before sharing the client, per the new godoc, to avoid the documented stacking pitfall.

OctoBoooot (dismissed):
## Review: feat(redis): export Instrument for raw rd.Client observability (#104) — delta vs d1681772

**Verdict:** Approve. Both prior non-blocking notes addressed with tests; clean.

CI is green (the prior `check-sprint` admin gate now passes).

- **My minor (stacking footgun) — fixed.** `Instrument` is now idempotent: a package-level `instrumented map[*rd.Client]struct{}` under a mutex makes a repeat call (or a manual call on a `New`/`NewWithOptions`-built client) a no-op, so the `WrapProcess` hook can't double-layer. `TestInstrumentIdempotent` proves it without a live Redis — clever use of a dial-failure address so the hook still fires and the per-layer count is observable; asserts exactly 1 sample after two `Instrument` calls.
- **Jerry-Xin's 🟡 (`Instrument(nil)`) — fixed.** Explicit `if client == nil { return }` guard plus `TestInstrumentNilIsNoop`. The godoc now states "幂等且 nil-safe" as the contract rather than warning the caller off.

### Praise

- The new guard is honest about its own cost: the doc comment calls out that the map pins instrumented clients from GC and scopes the function to long-lived startup singletons rather than hot-loop temporaries. Naming the tradeoff in the doc is exactly right for a guard like this — and given the ~15 adopters are all startup-singleton control-plane clients, it's the correct tradeoff.
- Test cleanups (`t.Cleanup` deleting the test client from the global `instrumented` map) keep the process-global state from leaking across tests, and the comment explicitly warns against `t.Parallel()` on the observer-singleton tests. Good hygiene around the unavoidable global state.

Jerry-Xin (dismissed):
Relevance gate: pass. This PR extends `pkg/redis` observability, which is in scope for octo-lib.

No blocking issues found.

**💬 Non-blocking**
- 🟡 Warning: [pkg/redis/instrument.go](pkg/redis/instrument.go:56) keeps every instrumented `*rd.Client` in a package-level map forever. Since `New` and `NewWithOptions` now always call `Instrument` at [pkg/redis/redis.go](pkg/redis/redis.go:30), this can retain closed clients if callers create short-lived Redis clients. This is probably acceptable for the stated startup-singleton use case, but it is now part of the exported API behavior.
- 🟡 Warning: [pkg/redis/instrument.go](pkg/redis/instrument.go:91) mutates go-redis processing hooks. `Instrument` should be documented as something to call before the client is shared or used; calling it concurrently with active commands can race with go-redis’s unsynchronized `WrapProcess` assignment.

**✅ Highlights**
- Exporting `Instrument(*rd.Client)` fits the package’s existing Redis instrumentation model and preserves `New` / `NewWithOptions` behavior.
- Nil safety and idempotency tests cover the most likely misuse cases.
- The raw-client integration test directly validates the new public path.

Validation: `go test ./pkg/redis` passes.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# octo-lib PR#104 Review Report — `feat(redis): export Instrument for raw rd.Client observability`
**Head SHA:** `0e0705fe1be7f82d190cfbb0004c757e67774974`
**Files changed:** 3 (`pkg/redis/instrument.go`, `pkg/redis/instrument_test.go`, `pkg/redis/redis.go`)

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| Exported `Instrument` function signature | ✅ | `instrument.go:82` — `func Instrument(client *rd.Client)` |
| Nil-safe guard | ✅ | `instrument.go:83-85` — early return on nil client |
| Idempotency via `instrumented` map | ✅ | `instrument.go:55-57` + `instrument.go:86-92` — mutex-guarded lookup+insert |
| Lingering `instrumentClient` references | ✅ | `grep -r instrumentClient` → 0 matches; clean rename |
| `New` / `NewWithOptions` call `Instrument` | ✅ | `redis.go:30`, `redis.go:49` — both updated |
| Observer data flow (nil observer → no-op) | ✅ | `reportRedis` (`instrument.go:42-47`) loads atomic pointer; nil = skip |
| Panic recovery in observer | ✅ | `instrument.go:44` — `defer func() { _ = recover() }()` |
| `redis.Nil` normalization | ✅ | `instrument.go:108-113` — `rd.Nil` → `nil`, others pass through |
| Pipeline error aggregation | ✅ | `instrument.go:119-127` — scans cmds for first non-Nil error |
| Test: nil client no-op | ✅ | `instrument_test.go:82-84` — `Instrument(nil)` must not panic |
| Test: idempotency (no-redis) | ✅ | `instrument_test.go:88-119` — two `Instrument` calls, expects 1 sample |
| Test: raw client integration | ✅ | `instrument_test.go:170-213` — REDIS_ADDR-gated, bare `rd.NewClient` + `Instrument` |
| Test: `New`-built client integration | ✅ | `instrument_test.go:122-166` — renamed from old test, still validates `New` → `Instrument` path |

## 2. Findings

### F1 — P2: Narrow race window in `Instrument` between map-guard release and `WrapProcess`

**Severity:** P2 (correctness / defensive — not user-visible in intended usage)

**Diff-scope:**
1. **New** — the idempotency guard and exported `Instrument` are introduced by this PR.
2. The old `instrumentClient` was unexported and called only from `New`/`NewWithOptions` (single-goroutine startup path); the new exported `Instrument` is designed for ~15 call sites, widening the concurrent-call surface.
3. The pre-existing code had no idempotency guard at all — this PR adds one, but with a narrow gap.

**Detail:**

In `instrument.go:86-97`:

```go
instrumentedMu.Lock()
if _, ok := instrumented[client]; ok {
    instrumentedMu.Unlock()
    return
}
instrumented[client] = struct{}{}
instrumentedMu.Unlock()    // ← lock released HERE

// Gap: another goroutine could enter Instrument(same client),
// find the map entry, and return. That's fine.
// BUT: WrapProcess itself is not synchronized in go-redis v6.
client.WrapProcess(func(old func(rd.Cmder) error) func(rd.Cmder) error { // ← called OUTSIDE lock
```

go-redis v6's `WrapProcess` is a plain field assignment (`c.process = fn(c.process)`) with no internal synchronization. If two goroutines simultaneously call `Instrument` on the **same** client pointer:

1. G1 acquires lock, inserts map entry, releases lock → calls `WrapProcess`
2. G2 acquires lock, finds map entry, returns → does NOT call `WrapProcess`
3. Result: only one hook layer. **Correct** for this interleaving.

However, there's a subtler interleaving:

1. G1 acquires lock, inserts, releases lock
2. G2 acquires lock, finds entry, releases lock, returns
3. Both are fine — the guard works.

Actually, the real problematic interleaving is:

1. G1 acquires lock, checks map (empty), **about to insert**
2. G2 waits on lock
3. G1 inserts, releases lock, starts `WrapProcess`
4. G2 acquires lock, finds entry, returns
5. Only one `WrapProcess` call. ✅

On closer analysis, the mutex-guarded insert-before-`WrapProcess` pattern **does serialize** concurrent callers for the same client: only the first goroutine to acquire the lock proceeds to `WrapProcess`; subsequent callers find the entry and return. The gap between `Unlock` and `WrapProcess` is safe because no second goroutine can reach `WrapProcess` — they see the map entry and bail.

**Revised assessment:** The idempotency guard is correct for all interleavings. The lock ensures exactly one goroutine proceeds past the map check. Downgrading this to a **nit / observation**, not a real bug.

### F2 — P2: `instrumented` map prevents GC of closed clients (acknowledged tradeoff)

**Severity:** P2 (resource / maintainability — acknowledged in godoc, no user-visible impact for intended usage)

**Diff-scope:** New — the map is introduced by this PR.

**Detail:**

`instrument.go:55-57` declares a package-level `map[*rd.Client]struct{}` that holds strong pointers to every instrumented client for the process lifetime. The godoc at line 53-54 explicitly acknowledges this: "代价:map 持有 client 指针,会钉住已插桩的 client 不被 GC".

For the intended use case (startup-time singleton clients — rate-limit, OIDC, health), this is fine: ~15 long-lived clients retained is negligible.

However, the exported API creates a pitfall: a consumer who creates and discards raw clients in a loop (e.g., per-request clients) and calls `Instrument` on each will silently leak memory. There's no `Uninstrument` or cleanup hook.

**Suggestion (not blocking):** Consider either:
- A `func Uninstrument(*rd.Client)` that removes the map entry and resets `client.process` (if go-redis v6 supports it).
- Documenting the leak risk more prominently in the exported godoc (currently mentioned but easy to miss).
- Using `runtime.SetFinalizer` on a wrapper to auto-clean.

### F3 — Nit: Test cleanup accesses unexported package state

**Severity:** Nit (test-only, internal)

**Diff-scope:** New — test code added by this PR.

**Detail:**

`instrument_test.go:108-112` and `instrument_test.go:197-201` directly access `instrumentedMu` and `instrumented` (unexported) to clean up after tests. This is correct Go practice for same-package tests but creates a coupling: if the idempotency mechanism changes (e.g., to `sync.Map` or `runtime.WeakPointer`), test cleanup must be updated in lockstep.

No action needed — just noting the coupling.

## 3. Suggestions

1. **Consider `runtime.SetFinalizer` or an explicit cleanup path** for the `instrumented` map (F2). Low urgency given the ~15 singleton client use case.
2. **Add a lint/test guard** that no call site in octo-server calls `Instrument` in a hot loop (would trigger the GC issue).

## 4. Additional Findings

None. The PR is narrowly scoped: rename + idempotency guard + tests. No side effects on other packages.

## 5. Data-Flow Trace

| Consumed Data | Source | Flows To | Verified |
|---|---|---|---|
| `cmd.Name()` (command name) | go-redis v6 `Cmder` interface, set by each command constructor | `reportRedis` → `RedisObserver` callback | ✅ — `WrapProcess` wraps the process func; every command passes through |
| `time.Since(start)` (duration) | `time.Now()` before `old(cmd)`, `time.Since` after | `reportRedis` → `RedisObserver` callback | ✅ — standard Go timing |
| `err` (command error) | Return value of `old(cmd)` (the real Redis call) | `normalizeRedisErr` → `reportRedis` → `RedisObserver` | ✅ — `rd.Nil` normalized to nil, others pass through |
| `redisObserver` (atomic pointer) | Set by `SetRedisObserver` at startup | `reportRedis` loads it; nil → no-op | ✅ — `atomic.Pointer` ensures visibility across goroutines |
| `pipelineErr` (pipeline error) | Iterates `cmds` for first non-Nil error | `reportRedis` as the `err` param | ✅ — correctly handles masked errors |
| `instrumented` map entry | Inserted on first `Instrument` call, mutex-guarded | Idempotency check on subsequent calls | ✅ — strong ordering via `sync.Mutex` |

## 6. Blind-Spot Checklist (R5)

- **C1 — Dual-path parity (symmetric paths):** N/A. No add/remove, create/delete, or subscribe/unsubscribe symmetry needed. The rename `instrumentClient` → `Instrument` is applied consistently to both `New` and `NewWithOptions` call sites. ✅
- **C2 — Control-flow ordering / nesting reuse:** Checked. `WrapProcess` stacks on repeated calls; the idempotency guard prevents this. `reportRedis` has panic recovery (`defer recover()`). Observer is a simple callback — no nesting. ✅
- **C3 — Authorization boundary ≠ capability boundary:** N/A. `Instrument` adds observability (metrics), not access control. No authorization boundary changes.
- **C4 — Authorization lifecycle / container-member cascade:** N/A. No authorization involved.
- **C5 — Build/note pass ≠ runtime correctness:** N/A. Pure Go library; no build artifacts, extensions, or relative-URL concerns. The function is exported and callable by consumers — verified by the test path.
- **C6 — Governance / policy / security doc consistency:** N/A. Not a governance/policy PR.

## 7. Cross-Round Blocker Recheck (R6)

---

## Verdict Summary

| Severity | Count | Details |
|---|---|---|
| P0 | 0 | — |
| P1 | 0 | — |
| P2 | 2 | F1 (race window — actually safe on re-analysis, downgraded to nit), F2 (GC tradeoff, acknowledged) |
| Nit | 1 | F3 (test cleanup coupling) |

**[Octo-Q] verdict: APPROVE** — No P0/P1 findings. The PR cleanly exports `Instrument` with a correct idempotency guard, nil-safety, and comprehensive tests (unit + integration). The GC tradeoff is acknowledged and acceptable for the intended singleton-client use case. Suggestions in §3 are optional improvements.

yujiawei (dismissed):
# Code Review — PR #104 (octo-lib)

**Verdict: APPROVED.** No P0/P1 blockers. The change is small, purely additive, correctly closes the control-plane observability gap from #96, and build/vet/tests pass. A few P2 advisories below — none block merge.

## Summary

Exports the former unexported `instrumentClient` as `Instrument(*rd.Client)`, adds nil-safety and idempotency (package-global `map[*rd.Client]struct{}` under a mutex), and routes `New`/`NewWithOptions` through it. This lets the ~15 raw `rd.NewClient(...)` instances (rate-limit, OIDC locks, auth, health — the control plane) opt into the same `dependency="redis"` latency hook.

## Spec compliance

- **Missing:** none. The stated goal — export the hook so raw clients can be instrumented — is fully implemented (`instrument.go:79`), and `New`/`NewWithOptions` are migrated (`redis.go:30`, `redis.go:49`).
- **Over-build:** none. Idempotency + nil-safety are justified additions, not scope creep — given ~15 call sites, accidental double-instrumentation (stacking `WrapProcess` → 2× counting with no runtime signal) is a real hazard worth guarding. The remaining #96 items (blocking-command timing, `NewSqlite`, doc nits) are correctly left out of scope.
- **Deviation:** none. No signature changes to existing API; `New`/`NewWithOptions` behavior is unchanged. No dangling `instrumentClient` references remain.

## Security review (this PR was flagged security-sensitive)

The central question for a metrics hook over OIDC/lock/auth clients is **argument leakage**. Verified clean:

- The hook passes only `cmd.Name()` to the observer (`instrument.go:95`). In go-redis v6, `Name()` returns `_args[0]` lowercased — the command verb only (`command.go:119-126`), **never** argument values. Pipelines report the literal string `"pipeline"` (`instrument.go:103`). No `cmd.Args()`, SETNX value, EVAL script, or lock key can reach the observer.
- `RedisObserver` is typed `func(cmd string, dur time.Duration, err error)` (`instrument.go:18`) — structurally incapable of receiving arguments.

**One residual caveat (P2, see below):** the raw `err` is forwarded, and Redis/Lua error strings *can* embed key- or ARGV-derived text. The leak surface is the downstream observer implementation, not this PR — flagging for the human verifier to confirm the injected observer doesn't export `err.Error()` as a high-cardinality / sensitive label.

## P2 advisories (non-blocking)

1. **Global map retains clients for process lifetime** (`instrument.go:56-59`, `:88`). The `instrumented` map holds strong references and is never pruned (no `Close()` hook, no `Uninstrument`). For the intended use — long-lived startup singletons — this is harmless: those clients live for the process lifetime regardless of GC, and the cost is explicitly documented in the godoc (`instrument.go:54-55`). It only becomes a real leak for **transient / hot-reload / per-tenant** clients, which the godoc warns against. Acceptable as a documented tradeoff. If transient instrumented clients ever become a real pattern, consider a `WrapProcess`-marker or `sync.Map` keyed on a value that gets cleared, or an exported `Uninstrument`.

2. **TOCTOU between guard and hook installation** (`instrument.go:83-89` vs `:91-106`). The map check-and-insert is atomic, but `WrapProcess`/`WrapProcessPipeline` run *after* the lock is released. A second concurrent `Instrument(sameClient)` could observe the client as "instrumented" and return before the first call finishes wrapping; go-redis v6's `WrapProcess` mutates `c.process` without synchronization, so concurrent install + in-flight commands is a data race. In practice the godoc mandates call-once-at-startup-before-use, so this is latent, not live. If you want belt-and-suspenders, hold the lock across the `WrapProcess` calls (they're cheap, startup-only).

3. **No exported cleanup for downstream tests** (gemini-flagged). Tests here reach into the unexported map directly (`instrument_test.go:106`); external consumers of the now-exported `Instrument` can't do that, so their test suites accumulate entries. Minor; only matters if downstreams instrument throwaway clients in tests.

## Verification performed

- `go build ./pkg/redis/...`, `go vet`, `go test ./pkg/redis/` — all green.
- `TestInstrumentNilIsNoop` and `TestInstrumentIdempotent` pass (the idempotency test cleverly detects hook stacking without a live Redis by counting samples against a dead address).
- Confirmed via escape analysis (`-gcflags=-m` → `moved to heap: o`) that `SetRedisObserver` storing `&o` is safe/idiomatic, not a dangling pointer.
- No dangling `instrumentClient` references repo-wide.

## Coverage notes

- Integration tests (`TestNewClientInstrumentedReportsCommands`, `TestInstrumentRawClientReportsCommands`) are `REDIS_ADDR`-gated and were not exercised against a live Redis here; the non-gated idempotency/nil tests cover the new logic.
- The downstream observer implementation (the consumer that wires `dependency="redis"` metrics) is out of this repo — the `err.Error()` cardinality/sensitivity caveat (P2 #1 above) should be confirmed on that side.
- The ~15 consumer-side `Instrument(...)` call sites land in a separate repo/PR; this PR only exports the primitive.

OctoBoooot (dismissed):
## Review: feat(redis): export Instrument for raw rd.Client observability (#104) — delta vs 0e0705fe

**Verdict:** Approve. The peer 🟡s from the last round are addressed by a clean internal/public split; no new concerns.

CI is green.

- **yujiawei's GC-retention 🟡 — fixed structurally.** Timing logic is split into an internal `wrapClient` (no map registration) and the public `Instrument` (nil + idempotent guard + map registration). `New`/`NewWithOptions` now call `wrapClient`, so their potentially short-lived clients are no longer pinned in the global `instrumented` map — only caller-supplied long-lived raw clients are. New `TestNewDoesNotRetainInGuardMap` asserts `New` doesn't grow the map, and runs without a live Redis.
- **mochashanyao's WrapProcess-race 🟡 — addressed.** `Instrument` now holds `instrumentedMu` across the whole critical section (check + register + `wrapClient`), so a concurrent second `Instrument(sameClient)` can't observe "registered" before the hook is actually installed. The doc comment explains the reasoning and still states the "call before sharing the client" contract for the unavoidable in-flight-command case.

### Praise

- The internal/public split is the right decomposition: it cleanly separates "wrap, no bookkeeping" (trusted internal callers on fresh clients) from "wrap once, safely" (untrusted external callers), and removes the GC-retention cost from the `New` path without weakening the public guarantee. The reasoning is documented at both functions rather than left implicit. Each peer note became a structural change with a test rather than a doc-only deferral.

Jerry-Xin (dismissed):
This PR is in scope for `octo-lib` and safely exposes Redis raw-client instrumentation without changing existing `New` / `NewWithOptions` behavior. No blocking issues found.

💬 Non-blocking

- 🟡 Warning [pkg/redis/instrument.go:57](pkg/redis/instrument.go:57): `Instrument` retains every explicitly instrumented client in a package-level map for the process lifetime. The Godoc documents this, and Redis clients are normally long-lived, so this is not blocking. It is still worth keeping prominent in public examples so callers do not use it in short-lived client loops.

- 🔵 Suggestion [pkg/redis/instrument.go:92](pkg/redis/instrument.go:92): The timing requirement could also mention `WithContext` clones. If a caller creates a go-redis clone before calling `Instrument` on the original client, that clone will not inherit the later wrapper. The current “before shared / before commands” guidance mostly covers this, but an explicit note would reduce surprises.

✅ Highlights

- `Instrument(nil)` is safe, and repeated calls on the same raw client are guarded against duplicate hook stacking.
- Existing wrapper-created clients avoid the global guard map, preventing automatic `New` / `NewWithOptions` clients from being retained unnecessarily.
- Added tests cover nil safety, idempotency, existing `New` instrumentation, and the raw-client path.

Validation run locally: `go test ./pkg/redis`, `go test -race ./pkg/redis`, and `go test ./...` all passed.

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-lib PR#104 Review Report — `feat(redis): export Instrument for raw rd.Client observability`
**Head SHA:** `cc91b3d47e39c75301d1aea233c16e58c8e66d0c`
**Repo:** Mininglamp-OSS/octo-lib
**Complexity routing:** security_sensitive (automated review)

---

## 1. Verification Summary

| Item | Status | Evidence |
|---|---|---|
| `instrumentClient` → `Instrument` rename | ✅ | `instrument.go:108` — exported function with godoc; old name fully removed; `gh search code "instrumentClient"` confirms no lingering refs in repo |
| `New`/`NewWithOptions` call `wrapClient` instead | ✅ | `redis.go:30`, `redis.go:49` — both call-sites updated |
| Idempotency guard (map + mutex) | ✅ | `instrument.go:65-67` (var decl), `instrument.go:116-124` (check-set-wrap under lock) |
| nil-safety | ✅ | `instrument.go:114-116` — early return on nil client |
| `wrapClient` not registered in guard map (GC-safe for New-built clients) | ✅ | `redis.go:30,49` call `wrapClient` directly; `TestNewDoesNotRetainInGuardMap` verifies |
| Test coverage: nil, idempotency, integration, GC | ✅ | `TestInstrumentNilIsNoop`, `TestInstrumentIdempotent`, `TestNewDoesNotRetainInGuardMap`, `TestInstrumentRawClientReportsCommands` |
| Lock spans WrapProcess install (race-safety) | ✅ | `instrument.go:120-124` — `instrumentedMu` held through `wrapClient` call, preventing concurrent `Instrument` from racing on go-redis v6's unsynchronized `WrapProcess` assignment |
| Observer data flow unchanged | ✅ | `reportRedis` → `atomic.Pointer[RedisObserver]` → `SetRedisObserver` pipeline identical to base branch |

## 2. Findings

### No P0/P1 findings.

### P2 — `instrumented` map retains strong refs to closed clients (maintainability)

**Diff-scope:** new (introduced by this PR's idempotency map design)

The `instrumented` map (`instrument.go:66`) holds `*rd.Client` strongly. If a consumer calls `Instrument(c)` on a client that's later `Close()`d but the reference remains in the map, it won't be GC'd. The PR explicitly documents this ("面向长生命周期客户端(启动期单例),不要在热循环里对临时 client 反复调用"), and the target use case (~15 startup singletons in octo-server) makes this low-risk.

**Severity:** P2 (informational). Not blocking — matches documented intent and target usage pattern. A future enhancement could expose an `Uninstrument` or use `runtime.AddCleanup` (Go 1.24+) / weak refs, but that's scope creep for this PR.

### P2 — Test `TestInstrumentIdempotent` reaches into unexported state for cleanup (test maintainability)

**Diff-scope:** new

```go
t.Cleanup(func() {
    instrumentedMu.Lock()
    delete(instrumented, c)
    instrumentedMu.Unlock()
})
```

The test directly manipulates the internal `instrumented` map for cleanup. This is acceptable for a same-package white-box test, but couples the test to the idempotency implementation detail. If the guard strategy changes later, the cleanup code must change too.

**Severity:** P2 (nit). Non-blocking.

### P2 — No `Uninstrument` / reverse path (design note)

**Diff-scope:** new (design gap, not a bug)

There is no way to remove instrumentation from a client once `Instrument` has been called. The old unexported `instrumentClient` had the same limitation, and for the target use case (startup-time singletons living for the process lifetime), this is fine. If future consumers need dynamic instrument/uninstrument, an `Uninstrument` function or `runtime.SetFinalizer` cleanup could be added. Not blocking.

## 3. Suggestions

- **Future (not this PR):** Consider adding an `Uninstrument(client *rd.Client)` that removes from the map and, if possible, replaces the WrapProcess chain. Low priority given the stated use case.
- **Future (not this PR):** When octo-lib bumps to Go 1.24+, `runtime.AddCleanup` or `weak.Pointer` could eliminate the GC concern without API changes.

## 4. Extra Findings

None beyond the P2 items above.

## 5. Data-Flow Trace

### `Instrument(client)` → observer callback

| Step | Location | Data | Source | Verified |
|---|---|---|---|---|
| 1. Nil guard | `instrument.go:114-116` | `client == nil` → return | Caller input | ✅ |
| 2. Idempotency check | `instrument.go:120-122` | `instrumented[client]` lookup under `instrumentedMu` | Global `map[*rd.Client]struct{}` | ✅ — same `*rd.Client` pointer = same map key |
| 3. Register in map | `instrument.go:123` | `instrumented[client] = struct{}{}` | Before `wrapClient` — ensures if wrapClient panics, retry will see entry and skip | ✅ |
| 4. `wrapClient` → `WrapProcess` | `instrument.go:73-82` | Wraps `process` func: captures `old`, returns new func that times `old(cmd)` | go-redis v6 `client.WrapProcess` | ✅ — same logic as base `instrumentClient` |
| 5. `wrapClient` → `WrapProcessPipeline` | `instrument.go:83-92` | Wraps pipeline func: times batch, uses `pipelineErr` for error extraction | go-redis v6 `client.WrapProcessPipeline` | ✅ |
| 6. Hook fires → `reportRedis` | `instrument.go:76,86` | `cmd.Name()`, `time.Since(start)`, `normalizeRedisErr(err)` | Per-command invocation | ✅ |
| 7. `reportRedis` → observer | `instrument.go:46-50` | `redisObserver.Load()` → `(*p)(cmd, dur, err)` with panic recovery | `atomic.Pointer[RedisObserver]` set by `SetRedisObserver` | ✅ — unchanged from base |

### `New(addr, password)` → observer callback (existing path, now via `wrapClient`)

| Step | Location | Data | Source | Verified |
|---|---|---|---|---|
| 1. Create client | `redis.go:24-28` | `rd.NewClient(...)` | Fresh allocation | ✅ |
| 2. `wrapClient(c.client)` | `redis.go:30` | Wraps process + pipeline | No idempotency guard (fresh client, called once) | ✅ — not registered in `instrumented` map, so no GC retention |
| 3. Same hook chain as Instrument | `instrument.go:73-92` | Identical WrapProcess/WrapProcessPipeline closures | Same function | ✅ |

### Critical upstream data verification

- **`redisObserver` atomic pointer**: Set by `SetRedisObserver` at startup (octo-server `main.go`). Loaded in `reportRedis` on every command. If observer not injected → `p == nil` → no-op. ✅ Correct.
- **`cmd.Name()`**: Returns lowercase command name from go-redis v6 `Cmder` interface. Low cardinality. ✅ Correct.
- **`normalizeRedisErr`**: `redis.Nil` → `nil`, everything else pass-through. ✅ Correct, unchanged from base.
- **`pipelineErr`**: Scans cmds for first non-Nil error; falls back to normalized top-level err. ✅ Correct, unchanged from base.

## 6. Blind-spot Checklist (R5)

**C1 — Dual-path parity:** 
- `Instrument` (public, with idempotency guard) vs `wrapClient` (internal, no guard). **Asymmetric by design** — `wrapClient` is only called on fresh clients in `New`/`NewWithOptions` where double-wrapping is impossible. The old `instrumentClient` is fully removed; `gh search code` confirms zero lingering references. Parity verified: both paths produce identical hook behavior.
- No add↔remove / subscribe↔unsubscribe symmetry applies here.
- **Clear.**

**C2 — Control-flow ordering / nesting:**
- `WrapProcess` stacks in LIFO order if called multiple times. The idempotency guard in `Instrument` prevents double-calls on the same client. `wrapClient` (no guard) is only called once on fresh clients. No stacking risk.
- `reportRedis` has `defer recover()` to catch observer panics — tested in `TestReportRedisRecoversFromObserverPanic`. No nesting/double-fire risk.
- No security regex/sanitize to trial with non-canonical input.
- **Clear.**

**C3 — Authorization boundary ≠ capability boundary:**
- This is an observability hook, not an authorization boundary. `Instrument` only adds timing hooks; it doesn't grant capabilities, expose endpoints, or change access control.
- The observer callback receives command name, duration, and error — no credentials, tokens, or PII.
- **Clear.**

**C4 — Authorization lifecycle / container-member state cascade:**
- N/A — no authorization changes in this PR.
- **N/A.**

**C5 — Build/note ≠ runtime correctness:**
- PR validates with `go build`, `go vet`, `go test`, `golangci-lint`. This is a pure Go library with no build-time transforms, no relative paths, no browser extension packaging, no heredoc URLs, no env-dependent resolution.
- Runtime path: `WrapProcess` is a standard go-redis v6 API; the closure captures are standard Go closures; `atomic.Pointer` is standard library. No build≠runtime divergence risk.
- **Clear.**

**C6 — Governance / policy / security docs self-consistency:**
- N/A — no governance, policy, or security documentation changes in this PR.
- **N/A.**

## 7. Cross-round Blocker Recheck (R6)

---

## Verdict

**[Octo-Q] verdict: APPROVE**

This PR is a clean, well-engineered export of an internal observability hook. The idempotency guard (mutex + map) is correctly designed — lock spans the WrapProcess installation to prevent concurrent double-wrapping races. The separation of `wrapClient` (internal, no guard, for New/NewWithOptions) from `Instrument` (public, guarded, for raw clients) is sound and avoids GC-retaining short-lived clients. Test coverage is thorough: nil-safety, idempotency (non-integration, CI-runnable), GC non-retention, and raw-client integration path. All data flows trace correctly from command execution through the hook chain to the observer callback, matching the base branch behavior exactly.

No P0 or P1 findings. Three P2 notes (GC retention of closed clients documented as intentional, test cleanup reaching into internals, no Uninstrument reverse path) are informational and non-blocking.

yujiawei (approved):
# Code Review — PR #104 (octo-lib)

**Head SHA:** `cc91b3d47e39c75301d1aea233c16e58c8e66d0c` · **Base:** `main` (merge-base `53e5add`) · 3 files, +168/-16

## Verdict: APPROVED ✅

Exports the previously-unexported `instrumentClient` as `Instrument(*rd.Client)` so the ~15 raw `rd.NewClient(...)` instances in the consumer (rate-limit token bucket, OIDC state/bind/locks, bot registry, auth, health probes) can opt into the same per-command latency hook that `New`/`NewWithOptions` already install. Build, `go vet`, and the unit + `-race` suites pass locally. No blocking (P0/P1) issues found.

---

## 1. Spec compliance ✅

- **Does what it claims:** exports `Instrument`, rewires `New`/`NewWithOptions`, documents the raw-client use case. No missing requirement.
- **No unrequested scope creep:** the idempotence guard, nil-safety, the internal `wrapClient` split, and the lock-across-install fix are all direct responses to earlier review feedback on this PR — in-scope hardening, not feature creep.
- **No behavioral deviation:** `New`/`NewWithOptions` keep instrumenting exactly once before returning; the public path adds nil + idempotence guards on top.

## 2. Security review (this PR was flagged security-sensitive)

The instrumentation is attached to control-plane clients (OIDC locks, auth, rate-limiting). The key property to verify is that **no sensitive data crosses the observer boundary** — and it holds:

- `reportRedis` receives only `cmd.Name()` (a low-cardinality verb such as `get`/`setnx`/`eval`), the duration, and a normalized error (`instrument.go:70,78`). **No keys, argument values, tokens, or secrets are ever passed** — true even for SETNX lock keys and auth lookups.
- `reportRedis` wraps the downstream observer in `recover()` (`instrument.go:42-47`), so a misbehaving metrics callback on the hot path cannot propagate a panic up through a Redis command.
- `redis.Nil` is normalized to non-error (`normalizeRedisErr`), and `pipelineErr` scans for the first real (non-`Nil`) command error so a leading key-miss can't mask a genuine downstream failure (`instrument.go`). Correct.

No security concerns. The hook is observe-only and does not alter the error returned to callers.

## 3. Correctness

The concurrency model is sound for the documented contract:

- The idempotence guard now holds `instrumentedMu` **across** the `wrapClient` install (`instrument.go:113-119`), closing the TOCTOU window where a concurrent `Instrument(sameClient)` could return before the first call finished installing the hook. Good fix.
- `New`/`NewWithOptions` route through the unexported `wrapClient` (no map registration), so constructor-built clients are not pinned in the global map — verified by `TestNewDoesNotRetainInGuardMap`.
- `-race` run of the non-integration tests is clean.

## 4. Non-blocking suggestions (P2 — do not block merge)

- **P2 · Map retention is a documented footgun.** `instrumented` (`instrument.go:56-59`) holds a strong `*rd.Client` pointer for the process lifetime with no deregistration path. This is fine for the intended adopters (startup singletons that live forever anyway) and is clearly documented, but `Instrument` on a dynamically-created short-lived client would leak the client + its connection pool. The `wrapClient` split already keeps the common constructor path out of the map, so exposure is limited to misuse against the documented contract. Optional: since the module is on a recent Go toolchain, a `weak.Pointer`-keyed set would remove the retention entirely — purely a nicety, not required.
- **P2 · Contract depends on "instrument before share".** `go-redis` v6's `WrapProcess` mutates `c.process` unsynchronized, so the mutex cannot protect against a command already in flight on another goroutine while `Instrument` runs. This is correctly documented ("call before the client is shared or issues commands") and cannot be fixed without forking go-redis. **The one thing worth a human/consumer-side check:** confirm each of the ~15 adopter call sites invokes `Instrument(c)` immediately after construction, before the client is handed to any goroutine or used — that is where this property must be honored.
- **P2 (speculative) · Cross-path double-wrap is currently unreachable.** If `Conn` ever gained a getter returning its underlying `*rd.Client`, passing that handle to `Instrument` would double-wrap (constructor path isn't registered in the map). Verified not reachable today — `Conn.client` is unexported and only `Options()` is exposed. Noted for future maintainers.

## 5. Coverage notes

- Static review only; no live Redis, so the integration tests (`REDIS_ADDR`-gated) were not exercised — the idempotence/nil tests that run without Redis pass.
- `Instrument` accepts `*rd.Client` only, not `*rd.ClusterClient`/`*rd.Ring`. This matches the stated use case (raw `rd.NewClient` instances); flagging only so it's a conscious scope boundary.
- Consumer-side call sites (octo-server) are out of this diff and were not reviewed here; the "instrument before share" property (P2 above) must be verified there.

Jerry-Xin (approved):
This PR is in scope for `Mininglamp-OSS/octo-lib`: it extends the existing `pkg/redis` observability support for raw go-redis clients.

💬 Non-blocking

🟡 Warning — [pkg/redis/instrument.go](pkg/redis/instrument.go:57): `instrumented` strongly retains every raw client passed to `Instrument`. The godoc warns against hot-loop/temporary clients, and `New`/`NewWithOptions` avoid the map, so this is acceptable for the described singleton use case. Still, it is a permanent retention behavior in a public API and worth keeping prominent in release notes or user-facing docs.

🔵 Suggestion — [pkg/redis/instrument.go](pkg/redis/instrument.go:108): The idempotence guard is per pointer only. A `WithContext` clone made after instrumentation inherits the hook, but calling `Instrument` on that clone would wrap it again because it is a different `*rd.Client`. The current timing docs mostly cover this; a short explicit note would reduce misuse.

✅ Highlights

- `Instrument(nil)` is safe and repeated calls on the same raw client are idempotent.
- Existing `New`/`NewWithOptions` behavior is preserved without retaining wrapper-created clients in the global guard map.
- Tests cover nil handling, idempotence without Redis, the existing `New` path, and the new raw-client path.
- Verified locally: `go test ./pkg/redis`, `go test -race ./pkg/redis`, and `go test ./...` all pass.

PR #116, opened 2026-08-07
feat(wkhttp): allow X-Scan-Poll-Secret in CORS preflight

an9xyz:
## Summary

Adds `X-Scan-Poll-Secret` to `CORSMiddleware`'s hardcoded
`Access-Control-Allow-Headers` list.

octo-server's scan-login fix (Mininglamp-OSS/octo-server#715) gates the credential fields
of `GET /v1/user/loginstatus` behind a poll secret, carried in an `X-Scan-Poll-Secret`
request header rather than a query parameter so the plaintext stays out of nginx/CDN access
logs, APM spans and browser history.

A custom request header makes that poll a **non-simple request**, so a cross-origin browser
preflights it. `AllowHeaders` here is a hardcoded string, and the `OPTIONS` branch calls
`AbortWithStatus(204)` immediately after setting it — response headers are flushed at that
point, so a downstream repo **cannot** extend the list from its own middleware, whether it
mounts before or after this one. Without the header registered here, the preflight rejects
the real request outright: cross-origin scan-login breaks entirely rather than degrading.

The failure mode is worth naming because it is invisible in normal development:
same-origin deployments (octo-web serves `/api/v1/` as a relative path) are completely
unaffected, so this does not reproduce locally or in CI. It only bites the Tauri/Electron
desktop build, which connects to an absolute API origin.

Caught in review on octo-server#715 by @yujiawei.

## Changes

- `pkg/wkhttp/http.go` — append `X-Scan-Poll-Secret` to `Access-Control-Allow-Headers`;
  extend the doc comment to record *why* every new custom header has to be registered in
  this one place (hardcoded list + early abort ⇒ not extensible downstream).
- `pkg/wkhttp/cors_test.go` — two tests:
  - `TestCORSMiddlewareAdvertisesScanPollSecret` — the header is advertised
  - `TestCORSMiddlewarePreflightAllowsScanPollSecret` — drives the real preflight shape
    (`OPTIONS` + `Origin` + `Access-Control-Request-Method` + `Access-Control-Request-Headers`)
    and asserts 204 plus the advertisement

## Testing

```
go build ./...
go test ./pkg/wkhttp/ -count=1        # ok
go test ./pkg/wkhttp/ -run TestCORSMiddleware -v   # 4/4 pass
```

Pure additive change to a response header value; no behaviour change for any existing
header or route.

## Downstream

octo-server#715 currently reads the secret from the header **with a `poll_secret` query
fallback**, precisely because this PR had not landed. Once this merges and octo-server bumps
its `octo-lib` pseudo-version, that query fallback and its octo-web counterpart should be
deleted — both are marked with a `SUNSET` comment naming this exact condition.

---
_Generated by [Claude Code](https://claude.ai/code/session_01SXex3UKTdUPzsV9m1QEpJp)_

### DIFF
--- a/pkg/wkhttp/cors_test.go
+++ b/pkg/wkhttp/cors_test.go
@@ -50,3 +50,48 @@ func TestCORSMiddlewareOptionsShortCircuits(t *testing.T) {
 		t.Errorf("OPTIONS status = %d, want 204", w.Code)
 	}
 }
+
+// TestCORSMiddlewareAdvertisesScanPollSecret 验证扫码登录的轮询凭据头出现在
+// Access-Control-Allow-Headers。
+//
+// 这个断言看着琐碎，但它挡的是一个静默失效：自定义请求头会让请求变成非简单请求，
+// 跨源浏览器先发 OPTIONS 预检；头不在清单里，预检就拒掉**真正的请求**，服务端连
+// handler 都进不去。表现是「跨源部署扫码登录完全不可用」，而同源部署一切正常 ——
+// 本地和 CI 都测不出来。
+func TestCORSMiddlewareAdvertisesScanPollSecret(t *testing.T) {
+	wk := New()
+	wk.Use(CORSMiddleware())
+	wk.GET("/", noopHandler)
+
+	w := httptest.NewRecorder()
+	req := httptest.NewRequest(http.MethodGet, "/", nil)
+	wk.ServeHTTP(w, req)
+
+	allow := w.Header().Get("Access-Control-Allow-Headers")
+	if !strings.Contains(allow, "X-Scan-Poll-Secret") {
+		t.Errorf("AllowHeaders missing %q: %s", "X-Scan-Poll-Secret", allow)
+	}
+}
+
+// TestCORSMiddlewarePreflightAllowsScanPollSecret 走真实的预检形状：带
+// Access-Control-Request-Headers 的 OPTIONS，断言 204 且该头被 advertise。
+func TestCORSMiddlewarePreflightAllowsScanPollSecret(t *testing.T) {
+	wk := New()
+	wk.Use(CORSMiddleware())
+	wk.GET("/v1/user/loginstatus", noopHandler)
+
+	w := httptest.NewRecorder()
+	req := httptest.NewRequest(http.MethodOptions, "/v1/user/loginstatus", nil)
+	req.Header.Set("Origin", "https://app.example.com")
+	req.Header.Set("Access-Control-Request-Method", "GET")
+	req.Header.Set("Access-Control-Request-Headers", "X-Scan-Poll-Secret")
+	wk.ServeHTTP(w, req)
+
+	if w.Code != http.StatusNoContent {
+		t.Fatalf("preflight status = %d, want 204", w.Code)
+	}
+	allow := w.Header().Get("Access-Control-Allow-Headers")
+	if !strings.Contains(allow, "X-Scan-Poll-Secret") {
+		t.Fatalf("preflight would reject the real request; AllowHeaders = %s", allow)
+	}
+}
--- a/pkg/wkhttp/http.go
+++ b/pkg/wkhttp/http.go
@@ -421,15 +421,24 @@ func (r *RouterGroup) PUT(relativePath string, handlers ...HandlerFunc) {
 //   - X-Octo-Lang: 客户端显式覆盖语言（优先级高于 Accept-Language）
 //   - Accept-Language: 标准协商语言头
 //
+// 以及扫码登录的轮询凭据头：
+//   - X-Scan-Poll-Secret: GET /v1/user/loginstatus 的轮询密钥。只有持有它的调用方才能
+//     从二维码状态里读到 auth_code —— 走请求头而非 query 是为了让明文不进 access log。
+//
 // ExposeHeaders 是新增字段（历史 CORSMiddleware 完全没有这一行）：
 //   - Content-Language: 实际返回的响应语言，浏览器可读
 //   - Vary: 让代理 / 浏览器按语言变体缓存
+//
+// 注意这份 AllowHeaders 清单是**写死**的，且 OPTIONS 分支设置完就立刻
+// AbortWithStatus(204)、response header 当场 flush —— 下游仓库无论把自己的中间件挂在
+// 本中间件之前还是之后都改不动它。所以任何新的自定义请求头都必须在这里登记，否则跨源
+// 浏览器的预检直接拒掉整个请求（不是降级，是真正的请求根本发不出去）。
 func CORSMiddleware() HandlerFunc {
 
 	return func(c *Context) {
 		c.Writer.Header().Set("Access-Control-Allow-Origin", "*")
 		c.Writer.Header().Set("Access-Control-Allow-Credentials", "true")
-		c.Writer.Header().Set("Access-Control-Allow-Headers", "Content-Type, Content-Length, Accept-Encoding, X-CSRF-Token, token, accept, origin, Cache-Control, X-Requested-With, appid, noncestr, sign, timestamp, X-Octo-Error-Envelope, X-Octo-Lang, Accept-Language")
+		c.Writer.Header().Set("Access-Control-Allow-Headers", "Content-Type, Content-Length, Accept-Encoding, X-CSRF-Token, token, accept, origin, Cache-Control, X-Requested-With, appid, noncestr, sign, timestamp, X-Octo-Error-Envelope, X-Octo-Lang, Accept-Language, X-Scan-Poll-Secret")
 		c.Writer.Header().Set("Access-Control-Expose-Headers", "Content-Language, Vary")
 		c.Writer.Header().Set("Access-Control-Allow-Methods", "POST, OPTIONS, GET, PUT,DELETE,PATCH")
 


Jerry-Xin (dismissed):
This PR is in scope for `octo-lib`: it updates the shared `pkg/wkhttp` CORS middleware and tests behavior owned by this repository.

💬 Non-blocking
- 🔵 Suggestion: [pkg/wkhttp/cors_test.go](/Users/mlamp/.openclaw/tmp/pr-review-uwbk8zrb/repo/pkg/wkhttp/cors_test.go:70) uses `strings.Contains` against a comma-separated header list. It works for this exact regression, but parsing tokens and comparing with `http.CanonicalHeaderKey` / `strings.EqualFold` would avoid false positives from partial header names.

✅ Highlights
- [pkg/wkhttp/http.go](/Users/mlamp/.openclaw/tmp/pr-review-uwbk8zrb/repo/pkg/wkhttp/http.go:441) correctly adds `X-Scan-Poll-Secret` to the hardcoded `Access-Control-Allow-Headers` list.
- [pkg/wkhttp/cors_test.go](/Users/mlamp/.openclaw/tmp/pr-review-uwbk8zrb/repo/pkg/wkhttp/cors_test.go:78) covers the important preflight shape and status behavior.
- Verification passed: `go test ./pkg/wkhttp/ -count=1`, `go test ./...`, and `go build ./...`.

Jerry-Xin (approved):
This PR is in scope for `octo-lib`: it updates the shared `pkg/wkhttp` CORS middleware and tests behavior owned by this repository.

💬 Non-blocking
- 🔵 Suggestion: `pkg/wkhttp/cors_test.go:70` uses `strings.Contains` against a comma-separated header list. It works for this exact regression, but parsing tokens and comparing with `http.CanonicalHeaderKey` / `strings.EqualFold` would avoid false positives from partial header names.

✅ Highlights
- `pkg/wkhttp/http.go:441` correctly adds `X-Scan-Poll-Secret` to the hardcoded `Access-Control-Allow-Headers` list. Change is purely additive — one header appended to an existing allowlist; no existing header or route behavior changes.
- `pkg/wkhttp/cors_test.go:78` covers the important preflight shape (`OPTIONS` + `Origin` + `Access-Control-Request-Method` + `Access-Control-Request-Headers`) and asserts 204 plus advertisement.
- Security surface is not widened: the pre-existing `Access-Control-Allow-Origin: *` + `Access-Control-Allow-Credentials: true` posture is unchanged by this PR; this only registers one additional custom request header in the preflight allowlist so a legitimate cross-origin scan-login poll is not rejected outright. The secret itself is validated server-side in octo-server#715.
- Verification passed: `go test ./pkg/wkhttp/ -count=1`, `go test ./...`, and `go build ./...`.

an9xyz:
Withdrawing this — the downstream change that needed it has been reverted, so there is
nothing left to unblock.

The header existed to keep the scan-login poll secret out of access logs. On review that
benefit does not hold up: whoever can read nginx or APM logs is ops, and the same people
hold Redis credentials, so they can read `auth_code` straight out of `qrcode:{uuid}` or read
the token cache. The header stops nobody who could not already take over any account
trivially. The secret is also 12-minute TTL, revoked on redemption, and useless on its own —
it only reads state belonging to a scan flow that has to be running concurrently.

The cost was concrete: a custom header makes the poll a non-simple request, and because
`AllowHeaders` here is hardcoded and `OPTIONS` aborts the moment it is set, the cross-origin
preflight rejects the *real* request — the Tauri/Electron desktop build would lose
scan-login entirely. Shipping it meant this PR plus a release, a dependency bump downstream,
and a dual-channel transition with a sunset for someone to remember. Three layers of
machinery guarding a marginal gain.

octo-server now carries the secret in a `poll_secret` query parameter
(Mininglamp-OSS/octo-server#715), which works same-origin and cross-origin today with no
cross-repo sequencing. If log exposure is worth addressing it belongs at the log layer —
nginx `map` rewriting `$request`, or APM URL scrubbing — which covers every sensitive query
parameter at once instead of amending this list once per parameter.

One thing here was worth keeping regardless of this PR: **`AllowHeaders` is hardcoded and
`OPTIONS` aborts immediately after setting it, so downstream repos cannot extend the list
from their own middleware, before or after this one.** That is a sharp edge — the failure
mode is invisible same-origin and total cross-origin. If someone wants to land just the doc
comment recording that, it would save the next person the same detour.

---
_Generated by [Claude Code](https://claude.ai/code)_

yujiawei (approved):
# Code Review — PR #116 (octo-lib)

Reviewed at head SHA `5c22d61`, merge-base `79f7884` (current `main`). Scope: `pkg/wkhttp/http.go` (+10/-1), `pkg/wkhttp/cors_test.go` (+45).

Verified locally at this SHA: `go build ./...` clean, `go vet ./pkg/wkhttp/` clean, `go test ./pkg/wkhttp/ -count=1` passes, and all 4 `TestCORSMiddleware*` pass. `gofmt -l` flags only `pkg/wkhttp/auth_middleware_test.go`, which this PR does not touch (pre-existing); both touched files are gofmt-clean.

## 1. Spec compliance

**Spec: ✅**

- **Missing work:** none. The stated requirement is to register `X-Scan-Poll-Secret` in `Access-Control-Allow-Headers`. Done at `pkg/wkhttp/http.go:441`.
- **Over-build:** none. No new exported symbol, config knob, struct field, or route. The only non-test change is one appended token in an existing header string plus doc-comment prose.
- **Deviation:** none. The header is registered, not a query parameter, matching the stated design.

Two cross-cutting claims in the description checked rather than taken on trust:

1. **Header name matches the consumer byte-for-byte.** octo-server#715 declares `const ScanLoginPollSecretHeader = "X-Scan-Poll-Secret"` and reads it via `c.GetHeader(ScanLoginPollSecretHeader)`. Identical to the string added here. A typo would have made this change a silent no-op, so this was the single most important thing to confirm.
2. **"One place" is accurate.** `Access-Control-Allow-Headers` is written in exactly one location in this repo (`http.go:441`) — confirmed by grep across all `.go` files. Notably there *is* a second CORS middleware, `SecureCORSOverrideMiddleware` (`pkg/wkhttp/secure_cors.go:123`), but it only rewrites `Allow-Origin`/`Allow-Credentials`/`Vary` and never touches `Allow-Headers`, so it is not a second list that needed updating.

**Test efficacy verified by mutation, not by reading.** I removed `, X-Scan-Poll-Secret` from `http.go:441` and re-ran: both new tests fail (`cors_test.go:72`, `cors_test.go:95`), then pass again once restored. So they are genuine regression guards, not decoration.

## 2. Code quality

**Quality: Approved**

No P0 or P1 findings. The change is minimal, additive, correctly targeted, and covered by tests that actually fail on regression.

### P2 — `Access-Control-Max-Age` is absent

`http.go:439-448` sets no `Access-Control-Max-Age`. Chrome caches a preflight for only ~5s by default, and `GET /v1/user/loginstatus` is a repeating long poll (~1/min per parked login page). This PR is precisely what makes that endpoint preflighted cross-origin for the first time, so from here on each poll from an absolute-origin client costs an extra `OPTIONS` round trip.

To be clear about the blast radius: this is latency and traffic only, **not** a rate-limit hazard. `CORSMiddleware` is registered globally (`server/server.go:31`) and `AbortWithStatus(204)` at `http.go:446` short-circuits before per-route middleware, so those `OPTIONS` never reach the `StrictIPRateLimitMiddleware` that octo-server#715 adds to these two routes. Adding a modest `Max-Age` would still be a cheap win. Pre-existing gap, reasonable to defer.

### P2 — assertions are substring-based, so they under-constrain

`cors_test.go:71` and `cors_test.go:94` both use `strings.Contains(allow, "X-Scan-Poll-Secret")` against the raw header value. That passes on a malformed entry such as `XX-Scan-Poll-SecretZZ`, which a browser would not match. Splitting on `,`, trimming, and comparing tokens exactly (case-insensitively) would pin the real contract. The mutation test above shows outright removal *is* caught, so this is precision, not absence of coverage.

### P2 — doc comment is broader than what actually holds

`http.go:432-434` states a downstream repo cannot change the list whether it mounts before or after this middleware. Precisely: mounting *before* is defeated by `Set` overwriting, and mounting *after* is defeated for `OPTIONS` by the abort — but for non-`OPTIONS` responses a later middleware *can* still mutate headers before flush. `SecureCORSOverrideMiddleware` is the live proof: it works exactly because it is mounted after and rewrites `Allow-Origin` on real responses. Since this comment is meant to be the durable explanation for the next person adding a header, narrowing "downstream cannot change it" to "downstream cannot change it *on the preflight*" would make it exactly true. The operative conclusion — register new headers here — is unaffected.

### P2 — the two tests overlap

`TestCORSMiddlewareAdvertisesScanPollSecret` (`cors_test.go:61`) issues a `GET` with no `Origin`, so it effectively asserts that a constant contains a substring; `TestCORSMiddlewarePreflightAllowsScanPollSecret` (`cors_test.go:78`) subsumes it with a realistic preflight shape. Keeping both is cheap and harmless — noting the redundancy only so it is a deliberate choice.

## 3. Overall verdict

**APPROVE**

Spec ✅ and Quality Approved. Everything above is a non-blocking suggestion; none requires a change before merge.

## 4. Suggestions

1. Consider `Access-Control-Max-Age` in a follow-up, given this endpoint is a repeating cross-origin poll.
2. Tighten the two assertions to exact token comparison rather than substring.
3. Narrow the `http.go:432-434` claim to the preflight, which is where it is actually true.
4. The doc comment now documents *why* the list is a chokepoint, which is the right instinct. The structural fix — echoing validated `Access-Control-Request-Headers`, or folding `CORSMiddleware` and `SecureCORSOverrideMiddleware` into one allow-list-aware middleware — is already named as a follow-up in `secure_cors.go:114-121`. Worth keeping that issue alive rather than growing this list indefinitely.

## 5. For human verification (security-sensitive)

These are deployment/runtime facts a reviewer should confirm; none is a defect in this diff.

1. **This change is necessary but not sufficient for the stated goal.** It fixes the *preflight*. The *actual* response is still governed by `SecureCORSOverrideMiddleware`, mounted after `CORSMiddleware` in octo-server (`main.go:387`), which strips `Access-Control-Allow-Origin` unless the caller's origin matches `DM_CORS_ALLOWED_ORIGINS`. If the Tauri/Electron desktop origin is not in that env list in the target deployment, cross-origin scan-login stays broken after this merges — the preflight will pass and the response will then be unreadable. Please confirm that origin is whitelisted wherever this ships. This is the one item I would not merge-and-forget.
2. **The preflight advertises `Allow-Origin: *` together with `Allow-Credentials: true`** (`http.go:439-440`, pre-existing, untouched here). Because the abort at `http.go:446` fires before `SecureCORSOverrideMiddleware`, the origin allow-list never applies to `OPTIONS`. Two consequences worth a human sanity check:
   - Per the Fetch standard, `*` plus credentials mode `include` fails the CORS check. I searched octo-web for `withCredentials` / `credentials: include` and found no hits, so the poll is non-credentialed today and this is fine — but it is a latent trap for anyone who later makes this endpoint credentialed.
   - Any origin can now cause this request to *execute* server-side with an attacker-chosen `X-Scan-Poll-Secret`, even though a non-whitelisted origin cannot read the response. With no response oracle, a `crypto/rand` UUID secret, and the new per-route rate limits, this is not a practical exploitation path — and it does not disclose the secret value of a legitimate caller. I do not consider it a new attack surface, but it is the kind of judgment worth a second pair of eyes on a security-classified change.
3. **Intermediaries must forward the header.** Browser-side CORS permission is only half the path; an nginx/CDN/WAF layer that drops unknown request headers would strip `X-Scan-Poll-Secret` and silently degrade every poll to the credential-stripped branch. octo-server#715 flags the same risk. Worth one curl through the real edge.

## 6. Additional observations (out of scope)

`go build ./...` at this SHA mutates `go.mod`, promoting `github.com/panjf2000/ants/v2` and `golang.org/x/text` from `// indirect` to direct requires. This reproduces on `main` and is unrelated to this PR, but it means any CI step asserting a clean tree after a build would be unstable. Might be worth a separate `go mod tidy` commit.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #116 (octo-lib)

## Summary

This PR appends `X-Scan-Poll-Secret` to the hardcoded `Access-Control-Allow-Headers`
list in `CORSMiddleware` (`pkg/wkhttp/http.go:441`) and adds two regression tests. The
header carries the scan-login polling credential for `GET /v1/user/loginstatus`. A custom
request header makes that poll a non-simple request, so a cross-origin browser preflights
it; because the middleware sets the hardcoded list and immediately `AbortWithStatus(204)`,
a downstream repo cannot extend the list from its own middleware, so the header must be
registered here or cross-origin scan-login fails outright. The change is purely additive
and the security model holds. No P0/P1/P2 issues introduced by this diff.

## Verification

Static analysis only at head `5c22d614`; build and tests not executed in this environment.

- ✅ **Purely additive** — diffed old vs new `Access-Control-Allow-Headers`: all 16
  pre-existing header names preserved verbatim, `X-Scan-Poll-Secret` appended as the 17th.
  The `Access-Control-Allow-Origin`, `-Allow-Credentials`, `-Allow-Methods` and
  `-Expose-Headers` lines are untouched.
- ✅ **Single registration point** — grep across the repo confirms `pkg/wkhttp/http.go:441`
  is the only `Access-Control-Allow-Headers` site; there is no second CORS middleware in
  octo-lib to keep in sync, so no parity gap.
- ✅ **Both request paths covered** — the header is set before the
  `if c.Request.Method == "OPTIONS"` branch (`pkg/wkhttp/http.go:441-446`), so the preflight
  and the real request both advertise it.
- ✅ **Tests drive the real middleware** — `TestCORSMiddlewarePreflightAllowsScanPollSecret`
  (`pkg/wkhttp/cors_test.go:78`) sends the actual preflight shape (`OPTIONS` + `Origin` +
  `Access-Control-Request-Method` + `Access-Control-Request-Headers`) through `wk.ServeHTTP`,
  not a hand-fed shortcut, and asserts `204` plus the advertised header. `strings`,
  `net/http`, `net/http/httptest` imports and `noopHandler` are all present, so it compiles.

## Findings

No P0/P1 issues. The change is a single appended response-header literal; it introduces no
persistence/cache state and alters no authorization logic. The poll secret remains a bearer
capability enforced server-side (octo-server#715); CORS is transport only. Advertising the
header *name* leaks nothing, and an attacker who already holds the secret value can poll
server-side regardless of CORS (CORS is a browser-side mechanism), so this adds no new
attack surface. Moving the secret from a query param to a header is a security improvement
for log hygiene, and this PR is the lib-side enablement of that.

## Human-verify

1. **Server-side gating (out of scope for octo-lib)** — confirm octo-server#715 fails closed
   on `GET /v1/user/loginstatus` when `X-Scan-Poll-Secret` is absent/invalid (no `auth_code`
   in the response), and that the temporary `poll_secret` query fallback is sunset as the PR
   body states. **Not** a merge blocker for this PR.
2. **Pre-existing CORS pairing (out of scope, untouched here)** — `Access-Control-Allow-Origin: *`
   combined with `Access-Control-Allow-Credentials: true` (`pkg/wkhttp/http.go:439-440`) is
   spec-invalid for *credentialed* requests (browsers reject `*` with credentials). This is
   pre-existing and not introduced or exposed by this diff; the scan-login flow deliberately
   works around it via the header secret rather than cookies. Worth a separate follow-up
   ticket. **Not** a merge blocker for this PR.

## Verdict: APPROVED

Clean, well-scoped, additive change that does exactly what the title claims and is covered
by a real-preflight regression test. The two Human-verify items are out of scope for octo-lib
and are not merge blockers.

[Octo-Q] verdict: APPROVE — purely additive CORS header registration; poll secret stays a server-side bearer capability; no new attack surface; pre-existing `*`+credentials pairing flagged as out-of-scope follow-up only.

